PR #17, opened 2015-06-19
Imgdetect demo

alx:
Static html interface to visualize results from image classifier predict api.


alx:
replace vendor dependencies with bower

beniz:
If I understand correctly, this requires bower installation, which is not packaged with the reference distro Ubuntu 14.04. Is there any other way to let the browser get the required bootstrap file online automatically ?

alx:
Replaced bower by cdnjs links.

beniz:
Many thanks. However, this does not work as is: the DeepDetect server complains with 
```
ERROR - Unknown Service=api
```
I guess this is due to passing the /api to the DeepDetect server. Any idea ?

beniz:
OK, it's `/api` in the rewrite rule instead of `locationapi`.

beniz:
Now, there's a bug in the demo:
- classify one image
- classify another image
-> the tags from both images are shown instead of only the predicted tags of the last image

alx:
You need to use the nginx rules as specified in the readme to have
deepdetect server behind a proxy on the same host/port as the webpage.

PR #41, opened 2016-01-03
Character-level 1D convolutional deep neural models for text

beniz:
This is the support for models that read characters in text and thus do not require any word parsing, counting, TFIDF, as is usual in otherwise BoW models.

The relevant publication is "Character-level Convolutional Networks for Text Classification",  http://arxiv.org/abs/1509.01626

These new models are of high interest to a range of tasks, from user generated content classification to the processing of URLs or raw logs. Since there's no parsing involved, the models are especially suited to multi-lingual data and / or data in which new or mis-spelled words appear outside of the training set.

This PR adds the following features:

- Convolutional Networks templating for 1D-convolutions with application to textual data
- API control at `txt` connector level: `characters` boolean keyword, along with `sequence` as an `int` defining the length of every sequence

A more thorough description through examples should be available on the website in the near future.


beniz:
Ready for merge, results from the original paper have been reproduced with sometimes even better accuracy. Tested on a range of other applications. Documentation is coming up.

Another discussion thread FTR: https://github.com/shogun-toolbox/shogun/issues/2701

PR #73, opened 2016-02-27
Example of image similarity search with both indexing and searching, based on Deep Convolutional Net

beniz:
See https://github.com/beniz/deepdetect/tree/imgsearch/demo/imgsearch and the instructions in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,13 @@ By default DeepDetect automatically relies on a modified version of Caffe, https
 
 The code makes use of C++ policy design for modularity, performance and putting the maximum burden on the checks at compile time. The implementation uses many features from C++11.
 
-##### Visual Demo
+##### Demo
 
+- Image classification Web interface:
 HTML and javascript classification image demo in [demo/imgdetect](https://github.com/beniz/deepdetect/tree/master/demo/imgdetect)
+
+- Image similarity search:
+Python script for indexing and searching images is in [demo/imgsearch](https://github.com/beniz/deepdetect/tree/master/demo/imgsearch)
 
 ##### Examples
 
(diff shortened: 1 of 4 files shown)


masakinakada:
Hi Em, Thank you very much for this awesome update. 

I tried to run with following command, but failed with the error message below.
python imgsearch.py --index /path/to/your/images --index-batch-size 64

///////////////////////////////////////////////// 
0/77Traceback (most recent call last):
  File "imgsearch.py", line 78, in <module>
    for p in classif['body']['predictions']:
KeyError: 'body'
/////////////////////////////////////////////////

Do you know the reason I am getting this error? Do we have the key 'body'? 
I am sorry for taking you extra time, but I really appreciate if you could help me with this.

Thank you so much, Em.
Masaki

beniz:
Ah you certainly missing the model template, I've added it just in case. This includes an update to master, you'll need to clear everything in your `build` directory and recompile from scratch due to a change in Caffe.

neo01124:
In the readme for this demo you allude to 

> 
 more complicated, ways to index and perform the search and scale.

Could you give some hints about these techniques ?

beniz:
@neo01124 see this conversation, https://discuss.elastic.co/t/what-is-the-best-scheme-for-similarity-search-based-on-binary-codes-with-elasticsearch-context-is-image-search-with-deep-nets/42915

Maybe you can share your current state of performances (e.g. number of images indexed, size of the feature vectors, etc...) as it will be useful to others who are using this similarity search demo as an embryo for their applications. 

Depending on how far you currently are from your 10 million images objective, the best response may wary.

PR #80, opened 2016-03-06
making client API compatible with python3

kyrs:
Hi beniz,
I have created the client Api code compatible with python3 and python2 both. Have a look at it.

### DIFF
--- a/clients/python/dd_client.py
+++ b/clients/python/dd_client.py
@@ -12,13 +12,22 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 """
 
-import ConfigParser
-import urllib2
-import httplib
+try:
+    import urllib.request as urllib2
+except ImportError:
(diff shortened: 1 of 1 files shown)


beniz:
Thanks for the PR @kyrs this is well done!

I've played with it a bit and here are my suggested changes:

- replace `hack_decode(js)` with `js.decode('utf-8')` since this is what is recommended with Python3, see http://stackoverflow.com/questions/6862770/python-3-let-json-object-accept-bytes-or-let-urlopen-output-strings
- remove the whole `hack_decode` function
- remove the `from builtins import str as text`

Could you fix the points above and commit the changes please ? (Btw, this helped me finding a big bug in the dd_client file, which I'll fix after your changes are being merged).

Thanks again!

kyrs:
@beniz  I have made the necessary changes you have asked me to do. Is there any other changes to be done ?

beniz:
This is perfect thank you, and it completes #63.

beniz:
Good job and thanks for your contribution :)

PR #97, opened 2016-04-04
Add cmake dependency, add missing 'cd build' command

alx:


beniz:
Good catch, thanks ^^

Issue #95, opened 2016-04-01
Add service selection to image classification frontend


beniz:
Javacript / HTML frontend from https://github.com/beniz/deepdetect/tree/master/demo/imgdetect would benefit from having a `service` selection menu.

beniz:
Done. Thanks @alx, cc @revilokeb.

PR #100, opened 2016-04-07
Add service selector on image classification frontend

alx:
Frontend is listing the services available on dd server.

User won't be able to upload an image if not service is available.

User are able to select the service to use during the image classification process.

Solves issue #95

### DIFF
--- a/demo/imgdetect/js/main.js
+++ b/demo/imgdetect/js/main.js
@@ -1,15 +1,47 @@
 $(document).ready(function() {
 
+  $('#service_select').select2();
+
+  var listServices = function() {
+
+    var service_url = "/api/info";
+
+    $.ajax({
+      type: "GET",
+      url: service_url,
+      dataType: 'json',
+      success: function(data) {
+
+        if(data.head.services && data.head.services.length > 0) {
+          $('#serviceForm').removeClass('bg-info')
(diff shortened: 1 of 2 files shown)


beniz:
Thanks for this @alx, it works just fine. There's one glitch, which is that every category now links to a wrong URL on imagenet, whereas I believe there should be no link at all. Can you fix this before merge ?

BTW, screenshot below, using two different models:

imgdetect_service_tab

demo/imgdetect/js/main.js
```diff
@@ -1,15 +1,47 @@
...
   $('a#urlSubmit').click(function() {
 
     $('.loading').removeClass('hidden');
 
     var post_url = "/api/predict";
     var post_data = {
-      service: "imageserv",
+      service: $('#service_select').val(),
       parameters: {
         input: {width: 224, height: 224},
```
beniz: This line should be removed or commented out as 224x224 is model specific. Any call with this on a model with inputs different than 224x224 yields a bad param error, which is normal (tested btw).

beniz:
Thanks, works fine. 

Beware: now image input size is mandatory in the `PUT /service` creation call.

Example of multi-service classification:

imgdetect_service_example

PR #127, opened 2016-05-17
Recovery from Caffe GPU memory overload on net prediction

beniz:
GPU memory overload with Caffe appears to be eratically recoverable. 

This PR makes the recovery deterministically possible. However it is obversed that multiple predict calls within memory bounds are often necessary for the service to recover after a GPU memory overflow.

For instance, on a 4GB using `resnet_152`, passing more than 7 or 8 images leads to 

```
{"status":{"code":500,"msg":"InternalError","dd_code":1007,"dd_msg":"src/caffe/util/im2col.cu:61 / Check failed (custom): (error) == (cudaSuccess)"}}
```

In our tests, trying a few more `predict` calls on the same service, with less images per batch, unlocks the memory and allows for the service to proceed correctly.

Note that deleting the service, re-creating it, and passing less images on the first `predict` call does not appear to prevent the need for a few calls to recover the GPU memory.

The same behavior is observed with and without CUDNN. While this PR is a step forward toward flawless recovery, the root causes of the behavior described above remain unknown at this stage, and more investigation is required.

### DIFF
--- a/src/caffelib.cc
+++ b/src/caffelib.cc
@@ -1224,7 +1224,10 @@ namespace dd
     if (ad.has("gpu"))
       _gpu = ad.get("gpu").get<bool>();
     if (ad.has("gpuid"))
-      _gpuid = ad.get("gpuid").get<int>();
+      {
+	_gpuid = ad.get("gpuid").get<int>();
+	_gpu = true;
+      }
     if (ad.has("nclasses"))
       _nclasses = ad.get("nclasses").get<int>();
     if (ad.has("regression") && ad.get("regression").get<bool>())
--- a/src/caffelib.cc
+++ b/src/caffelib.cc
@@ -1197,6 +1197,7 @@ namespace dd
 	  }
 	catch (std::exception &e)
 	  {
+	    LOG(ERROR) << "Error creating network";
 	    throw;
 	  }
 	LOG(INFO) << "Using pre-trained weights from " << this->_mlmodel._weights << std::endl;
@@ -1206,6 +1207,7 @@ namespace dd
 	  }
 	catch (std::exception &e)
 	  {
+	    LOG(ERROR) << "Error copying pre-trained weights";
 	    delete _net;
 	    _net = nullptr;
 	    throw;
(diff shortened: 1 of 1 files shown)


tmeroniown:
beniz - what about simply limiting the number if images that are fed into the network to be < batch_size (that the network was created with in the deploy file). This wont solve the root cause but its also a way of making sure it doesnt surface.

beniz:
@tmeroniown the proper `batch_size` is machine dependent as you know, so we can't provide a fixed value. Letting users set it up would lead to the service failing at creation when not enough memory is available. 

I do see your point for working around the recovery issue though. But I feel it would put too strong of a limit where there shouldn't be one, though I may be still missing some elements.

One simple thing we could do though is build a table of max `batch_size` for most common networks, GPU and associated image sizes (often fixed for a given network). This table could then make it to the Python client for instance. 

Discussion of such matters is highly appreciated, as it improves the user experience, and lowers support :) Let me know your thoughts.

tmeroniown:
@beniz  - The fact that batch size if machine dependent is orthogonal to this issue, I believe. Basically the service would not have been created without a proper batch_size number (because a too large number would create a memory problem) - this is regardless of the machine. so if the service WAS created successfully, then we know the batch_size is good. therefore, I think we can use it as a way of preventing the memory overload during prediction. Regardless I agree with you that we should have a table with recommended batch sizes per network per HW config. at the minimum we should have it for the AWS instances which many people are probably using.

beniz:
OK, I better see your point now, thanks. 

What we could do then is add a `batch_size` (or other name, keeping 'batch_size' for training?) parameter to the `parameters:mllib` object for Caffe, and that would instantiate the corresponding array/tensor across the model at service creation.

The `image` connector could then enforce the limitation somehow. I like the idea since it would make things more robust in production. This should probably go into an issue of its own though. Want to create it ?

Table should go into an issue of its own too.

tmeroniown:
ok I created the issue. btw - If the user passes in more images in the image data array during prediction than allowed by the batch_size, it would be great if dede would automatically queue these and feed into the network in batches that are smaller than batch size. I included that in the issue description

beniz:
> If the user passes in more images in the image data array during prediction than allowed by the batch_size, it would be great if dede would automatically queue these and feed into the network in batches that are smaller than batch size

Yes, agreed.

Issue #130, opened 2016-05-17
batch_size config table


tmeroniown:
We should have a table that calls our recommended batch_size per network per HW config (for the most common HW configs). 
at the minimum we should call out the right configs for the AWS instances.

PR #132, opened 2016-05-18
Image client

tmeron:
added the deepdetect image client to simplify image classification and feature extraction tasks.


beniz:
Thanks, this should prove very useful. Two things to start with, can you:
- change the class name to `DDImage` to use the same scheme as the current generic client 
- rename the file to `dd_image_client.py`

tmeroniown:
made the changes you have requested.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        self.mltype = ''
+        self.extract_layer = ''
+        self.nclasses = 0
+        self.width = self.height = 224
+        self.binarized = False
+        self.GPU = True
+        self.model_repo = ''
+        self.dd = None
+        self.model_name = ''
+        self.batch_size = None
```
beniz: Let's have a default `batch_size` instead of a non value ?

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+
+    mllib = 'caffe'
+
+    def __init__(self, host):
+
+        self.host = host
+        self.sname = ''
+        self.description = ''
+        self.mltype = ''
+        self.extract_layer = ''
```
beniz: Let's make `extract_layer` configurable in the constructor since there are many layers of interest in a given net.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        self.nclasses = 0
+        self.width = self.height = 224
+        self.binarized = False
```
beniz: `binarized` in constructor ?

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        self.dd.set_return_format(self.dd.RETURN_PYTHON)
+
+    def init_for_extraction(self, model_name, model_repo):
+        self.sname = model_name + "_extraction"
+        self.description = self.sname + ' for feature extraction'
+        self.mltype = 'unsupervised'
+        self.model_name = model_name
+        self.model_repo = model_repo
+
+        if model_name == 'resnet152':
```
beniz: We may want to differentiate the model name from the template name, 'googlenet' is a template name in dd

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+            self.extract_layer = 'fc1000'
+            self.nclasses = 1000
+            self.batch_size = self.RESNET152_BATCH_SIZE
+        elif model_name == 'googlenet':
+            self.extract_layer = 'pool5/7x7_s1'
+            self.nclasses = 1000
+            self.batch_size = self.GOOGLENET_BATCH_SIZE
+        else:
```
beniz: We can't fail just because we're not using `googlenet` or `resnet152`.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        self.nclasses = 0
+        self.width = self.height = 224
+        self.binarized = False
+        self.GPU = True
+        self.model_repo = ''
+        self.dd = None
+        self.model_name = ''
+        self.batch_size = None
+        self.GOOGLENET_BATCH_SIZE = 40
+        self.RESNET152_BATCH_SIZE = 4
```
beniz: We may want to start the configuration table suggested in #130 right here, and leave free values otherwise.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        urls = []
+        url_file = open(data_file, 'r')
+        for url in url_file:
+            urls.append(url.rstrip('\n'))
+
+        print("total number of image URL's to process " + str(len(urls)))
+
+        uri_to_features = {}
+        image_processed = 0
+        num_of_erros = 0
```
beniz: typo? `num_errors`

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+            t0 = time.time()
+            responses = self._make_feature_extraction_call(urls_to_process)
+
+            if responses is not None:
+                for response in responses:
+                    uri_to_features[response['uri']] = response['vals']
+                else:
+                    num_of_erros += 1
+
+            t1 = time.time()
```
beniz: FTR, processing time should already be returned by the server. Keeping track of processing time outside the server accounts for network latency. This is something useful, but I suggest the breakdown is returned, with total time, network (and image conversion in the base64 case) latency, and server-side processing time. Very useful for debug and setting up the proper low latency architecture.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+            image_processed = image_processed + self.batch_size
+            print("total time it took to process " + str(i) + " images was " + str(t1 - t0) + ' seconds')
+
+        full_file = open(target_file, 'w')
```
beniz: Very often we may want to extract the features into memory (we do this very often in image similarity search applications). I suggest the extracted features are optionnally dumped to disk.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+            urls.append(url.rstrip('\n'))
+
+        print("total number of image URL's to process " + str(len(urls)))
+
+        uri_to_features = {}
+        image_processed = 0
+        num_of_erros = 0
+
+        start_time = time.time()
+        while len(urls) > 0:
```
beniz: Any reason why we don't have this loop in `extract_features_from_url` as well ?

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        image_vectors = {}
+
+        while len(image_files) > 0:
+            img_path = image_files.pop()
+            img_data = self._pre_process_image(img_path)
+            response = self._make_feature_extraction_call(img_data)
+            image_vectors[img_path] = response['vals']
+            print("files left to process " + str(len(image_files)))
+
+        target_file = open(target_file_path, 'w')
```
beniz: Could be useful to be a version of this that returns extracted features into memory.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        return None
+
+    def _convert_img_to_base64(self, img):
+        tmp_buffer = cStringIO.StringIO()
+        img.save(tmp_buffer, format="JPEG")
+        base64_img_str = base64.b64encode(tmp_buffer.getvalue())
+        tmp_buffer.close()
+        return base64_img_str
+
+    def _rotate_image_if_needed(self, img, path):
```
beniz: DD supports cropping and mirroring, we should have the same option as we have here for rotations.

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        try:
+            exif = dict((ExifTags.TAGS[k], v) for k, v in img._getexif().items() if k in ExifTags.TAGS)
+            if exif['Orientation'] == 6:
+                print("rotating image " + path)
+                img = img.rotate(270, expand=True)
```
beniz: I believe we should have this operation into DD, same for mirroring and (random) cropping

clients/python/dd_image_client.py
```diff
@@ -0,0 +1,316 @@
...
+        except Exception:
+            print(path + " probably does not need rotation")
+        return img
+
+    def _pre_process_image(self, image_path):
```
beniz: There are many other useful image processing steps we could think of. I guess the discussion is to whether have them here or into DD... No real opinion here, though maybe those already supported at training time (rotate, crop, mirror) should be available at prediction time.

beniz:
Finally, training is missing in the script and I believe this is something that is needed. I can help with this last part as we have some dedicated code for image training.

clients/python/dd_image_client.py
tmeroniown: The implementation is tied to the models that DD currently comes with. I agree it may be a good idea to add a more generalized one, but that wasn't the intent of this implementation. I wanted people to be able to start using the client with 3 lines of code which is why the init methods basically take care of everything and fail otherwise.

clients/python/dd_image_client.py
tmeroniown: The batch size is defined in the init_for methods depending on the model you select. I dont think this belongs in the constructor.

clients/python/dd_image_client.py
tmeroniown: I can extract that in the init_for method which controls this and all other inits for extraction. I dont think this belongs in the constructor.

clients/python/dd_image_client.py
tmeroniown: do you see a situation where template is not the same as model? if the template today defines the deply.prototxt files than I dont see how that can happen.

clients/python/dd_image_client.py
tmeroniown: agreed

clients/python/dd_image_client.py
tmeroniown: agreed

clients/python/dd_image_client.py
tmeroniown: need to give you some context here. I implemented this mainly because I am working with mobile images which for the most part come rotated horizontally. I agree this doesnt belong in the client.

clients/python/dd_image_client.py
tmeroniown: with regards to training - I havent used DD for training thus far, so I cant really help here in terms of implementation

clients/python/dd_image_client.py
tmeroniown: again - this was done in specific prediction context while using images taken by mobile devices. I agree it can be implemented server side

clients/python/dd_image_client.py
beniz: Yes, if I get it correctly, `model_name` is actually part of the final service name. You can have many services running in parallel using the same template (but different weights).

clients/python/dd_image_client.py
beniz: Templates are just the tip of the iceberg. Basically most of the Caffe model zoo can be used. Also, the number of classes cannot be fixed, most applications finetune these pre-trained models for custom applications. My hunch is that having a generic image client, and adding the table suggested in #130 is the right way to go. Open to suggestions to move faster though, if this is too much work. Note that just an embryo of table is needed, we can help filling it up.

clients/python/dd_image_client.py
beniz: OK, are you running the Python client code on mobile then ?

clients/python/dd_image_client.py
beniz: OK I can add it later.

clients/python/dd_image_client.py
beniz: But `batch_size` is common to both classification and extraction, it only depends on the machine capabilities and template name I think.

clients/python/dd_image_client.py
tmeroniown: ok I see your point. I think what I will do is that I will extract all model based config out and put it in some dictionary/config file structure. agreed?

beniz: Thanks for taking the time for this. Yes I agree that this would be a useful generic addition, with the ability to capture future additions to the API without any code change to the client. To achieve this you may want to use four dictionaries, one for the generic parameters, the others for `input`, `mllib` and `output`. Not saying you need this, just that you may.

Issue #70, opened 2016-02-23
Does not install eigen on os x 10.11.3 el capitan


deeperx:
**When try to build as:**

```
mkdir build && cd build
cmake ..
```

**cmake gives the following error:**

```
FindEigen3.cmake
CMake Error at CMakeLists.txt:47 (find_package):
  By not providing "FindEigen3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Eigen3", but
  CMake did not find one.

  Could not find a package configuration file provided by "Eigen3" with any
  of the following names:

    Eigen3Config.cmake
    eigen3-config.cmake

  Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
  "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "/Users/inanc/dev/deepdetect/build/CMakeFiles/CMakeOutput.log".
```

----

It completely does not respect the path being set as described in the message. Such as: While ```CMAKE_MODULE_PATH``` is set to homebrew`s eigen cmake file directory, it does not respect that.

Can I send a bug fix for this?

beniz:
hi @deeperx OSX is not a supported platform, only Ubuntu 14.04LTS is at the moment. However, any patch and fixes that you can submit and that do not break the reference platform builds are highly welcome and are to be brought into the codebase, yes.

beniz:
Also, have you tried `cmake -DCMAKE_MODULE_PATH=/path/to/where` ?

deeperx:
Hey @beniz Yes, I've did tried that but still had no chance. I can send a patch but I can't validate whether it would work on ubuntu or no.

beniz:
Great please do open a PR with description and usage example or send a patch if you cannot do otherwise, I can do the checking against the reference platform.

beniz:
Not sure where you are at with this one @deeperx but a workaround to move a step further on OSX could be edit `deepdetect/CMakeLists.txt`, comment out `find_package(Eigen3 REQUIRED)` and replace `include_directories("${EIGEN3_INCLUDE_DIR}")` with `include_directories("/path/to/eigen3")`. If this works, then something else will probably fail, but this could pave the way to an OSX (unsupported) build.

beniz:
[...]

PR #153, opened 2016-06-28
Fix cmake issue for OSX El Capitan

neo01124:
The functions deepdetect_detect_darwin_version() and deepdetect_option() can't be found while executing 
`cmake ..`

I have used @beniz suggestions pertaining to https://github.com/beniz/deepdetect/issues/70#issuecomment-227422068 to fix that.


beniz:
Thanks for contributing this. Let's wait until #70 is possibly resolved and build up in here the required changes you may found required, if you don't mind.

PR #174, opened 2016-09-11
Neural text embeddings via Caffe

beniz:
This PR adds embeddings via the `Embed` layer in Caffe. This is useful for large text vocabularies, recurrent networks, and more.

Example on the news20 dataset:
- create the service, as an MLP with input embedding size of 100, and one dense layer of 50 neurons + softmax:

```
curl -X PUT "http://localhost:8080/services/n20" -d "{\"mllib\":\"caffe\",\"description\":\"newsgroup classification service\",\"type\":\"supervised\",\"parameters\":{\"input\":{\"connector\":\"txt\"},\"mllib\":{\"template\":\"mlp\",\"nclasses\":20,\"layers\":[100,50],\"activation\":\"relu\",\"embedding\":true}},\"model\":{\"templates\":\"../templates/caffe/\",\"repository\":\"/path/to/model/\"}}"
```
- train a model with embedding:

```
curl -X POST "http://localhost:8080/train" -d "{\"service\":\"n20\",\"async\":true,\"parameters\":{\"mllib\":{\"gpu\":true,\"solver\":{\"iterations\":50000,\"test_interval\":1000,\"base_lr\":0.01,\"solver_type\":\"ADAM\"},\"net\":{\"batch_size\":300}},\"input\":{\"sequence\":1024,\"embedding\":true,\"shuffle\":true,\"test_split\":0.2,\"min_count\":10,\"min_word_length\":5,\"count\":false},\"output\":{\"measure\":[\"mcll\",\"f1\"]}},\"data\":[\"/path/to/data/news20/\"]}"
```

Embeddings have been tested on both CPU and GPU.

This should be the last update to the network creation facilities before a complete refactoring that will allow to go much further.

Forthcoming:
- [x] Test with `db:true`
- [ ] Documentation
- [ ] Unit tests


kyrs:
Hi, I am trying to run this `PR` for `news 20` dataset. Although, I ran it perfectly with bag of word format. but when I tried to run it with flag `characters = True` it failed with error `{"code":500,"msg":"InternalError","dd_code":1007,"dd_msg":"src/caffe/layers/embed_layer.cpp:15 / Check failed (custom): (K_) > (0)"}` Does present PR support character based encoding ?? or am I doing something wrong??

beniz:
Good point, it doesn't support character level encoding. It could be added...

PR #208, opened 2016-10-27
lstm support for deepdetect

kyrs:
I have made some changes in `to_datum` function of `caffeinputconns.h`. I guess after slicing along `width` we can extract the `delta` necessary for lstm test.


kyrs:
still untested... :)

PR #220, opened 2016-12-02
Fix _best parameter overwritted by bbox

### LINKED ISSUE #219: Best argument overwritten when predicting
This command will return predictions for every classes (1000 instead of 5) even if "best":5 has been set.
```
       "service":"test_caffe_classification",
       "parameters":{
         "output":{
           "best":5
         }
       },
       "data":["http://www.dogbreedinfo.com/images26/LabradorRetrieverBlackPurebredDogSonny8YearsOldTim1.JPG"]
     }'
```
I think this is because of:
*  https://github.com/beniz/deepdetect/blob/master/src/caffelib.cc#L2272
*  https://github.com/beniz/deepdetect/blob/master/src/supervisedoutputconnector.h#L197
*  https://github.com/beniz/deepdetect/blob/master/src/supervisedoutputconnector.h#L154

Conchylicultor:
Fix #219

### DIFF
--- a/src/caffelib.cc
+++ b/src/caffelib.cc
@@ -2269,7 +2269,10 @@ namespace dd
       }
     
     out.add("nclasses",nclasses);
-    out.add("bbox",bbox);
+    if (bbox)
+      {
+	out.add("bbox",bbox);
+      }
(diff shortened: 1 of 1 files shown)


beniz:
#220 was best.

PR #221, opened 2016-12-02
Fix _best overwritten by bbox

Conchylicultor:
Probably a better solution for #219 than #220  (you can choose :) )


beniz:
Yes, better solution, thank you!

Issue #50, opened 2016-01-27
Model repository does not read correctly on XFS filesystem


beniz:
As explained in http://nerdfortress.com/2008/09/19/linux-xfs-does-not-support-direntd_type/ XFS does not support direntd and everything appears as a regular file. 
This one post http://stackoverflow.com/a/29094555 also reports possible workaround by testing for DT_UNKNOWN and using `lstat` on every file.

cchadowitz-pf:
The first link seems to be broken - here's a cached version: https://web.archive.org/web/20160315093719/http://nerdfortress.com/2008/09/19/linux-xfs-does-not-support-direntd_type/

PR #228, opened 2016-12-27
fix XFS issue

uhfband:
hi
On Centos 7 with XFS,  regular files has DT_UNKNOWN type, so cannot be listed by fileops::list_directory()

### DIFF
--- a/src/utils/fileops.hpp
+++ b/src/utils/fileops.hpp
@@ -71,10 +71,23 @@ namespace dd
 			      std::unordered_set<std::string> &lfiles)
     {
       DIR *dir;
+      DIR *subdir;
       struct dirent *ent;
       if ((dir = opendir(repo.c_str())) != NULL) {
 	while ((ent = readdir(dir)) != NULL) {
-	  if ((files && (ent->d_type == DT_REG || ent->d_type == DT_LNK || ent->d_type == DT_UNKNOWN))
+	  
+	  if(ent->d_type == DT_UNKNOWN)
+	    {
+	    	std::string ef = std::string(repo) + "/" + std::string(ent->d_name);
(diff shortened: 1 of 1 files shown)


beniz:
Thanks, have you looked at #50 ? If yes, can you report on ut_caffeapi unit tests passing?

uhfband:
Sorry, I missed it.
On my machine (Centos 64-bit  XSF) for weird reason stat() also returns the same bitmask in st_mode for both directory and regular file. I have made  the check with opendir(). It's ugly, but works. 
Now ut_caffeapi passes image training, but fails on service_train_txt
[testlog.txt](https://github.com/beniz/deepdetect/files/675532/testlog.txt)

beniz:
I see, thanks. The test error is likely to be due to your changes. This is a test call failing, and that reads data from the model directory.

uhfband:
I thought failure was due to wrong prediction
`5: INFO - 17:57:51 - Ignoring source layer drop0joutstr={"status":{"code":200,"msg":"OK"},"head":{"method":"/predict","service":"my_service"},"body":{"time":146646.0,"measure":{"accp":0.054912988115449919,"recall":0.05155933172795253,"precision":0.0534174967770092,"f1":0.0524719688263868}}}
5: /root/src/deepdetect/tests/ut-caffeapi.cc:591: Failure
5: Value of: jd["body"]["measure"]["f1"].GetDouble() >= 0.6
5:   Actual: false
5: Expected: true
5: [  FAILED  ] caffeapi.service_train_txt (306638 ms)`
Thanks, I will try another way

beniz:
Yes that's a predict call with measure that reverts into a test call (since there's no test call per se in the API).

Issue #234, opened 2017-01-11
Support for Tensorflow 1.0 (starting with alpha)


beniz:
Lot of changes, see https://github.com/tensorflow/tensorflow/releases/tag/v1.0.0-alpha so this may take some time.

beniz:
PR #255 will be merged when TF 1.0 stable is released.

beniz:
Merged.

PR #255, opened 2017-02-05
support for TF 1.0

beniz:
This fullfills #234 but merging will occur once TF 1.0 stable is released.


beniz:
Tested with TF 1.0 final.

Issue #289, opened 2017-04-12
Building DeepDetect with GPU/CUDA only for Tensorflow, CPU for Caffe


cchadowitz-pf:
#### Configuration
- Version of DeepDetect:
    - [x] Locally compiled on:
        - [x] Ubuntu 16.04 LTS, in Docker
- Commit (shown by the server when starting): `4a73d6c6b3e4667872f12d85b303ce1ceb7d1504`

#### Your question / the problem you're facing:
As a temporary workaround to the issue described in #230, I am trying to add a compilation variable `USE_CAFFE_CPU_ONLY` that will compile the DeepDetect version of Caffe with GPU/CUDA disabled (i.e. CPU only) while still compiling Tensorflow with GPU. That way I will be able to create both Caffe and TF services, letting the TF library utilize the GPU and Caffe simply use the CPU. Since TF defaults to GPU if built with CUDA and the GPU is available, there is no way to simply force it to use CPU.

Since the DeepDetect API for the Caffe mllib includes a `gpu` parameter (and a `gpuid` parameter), I originally thought that could be a simple way to disable GPU use with Caffe, but it appears that Caffe still uses some GPU memory regardless as I still get the `StreamExecutor context` error while using the `"gpu": false` parameter.

See this [gist](https://gist.github.com/cchadowitz-pf/8ff95a2f6585aeb9cc629820fd745111) for the diff of the changes I made to DeepDetect to try to build Caffe without CUDA (so, in CPU_ONLY mode) while letting TF build against CUDA. Even with these changes, I have the same `StreamExecutor context` error, and if I `watch -n0.5 nvidia-smi` while starting up the various services, the GPU memory climbs while loading Caffe models, indicating that Caffe still somehow seems to be using CUDA.

```
F tensorflow/stream_executor/cuda/cuda_driver.cc:334] current context was not created by the StreamExecutor cuda_driver API: 0x3228000; a CUDA runtime call was likely performed without using a StreamExecutor context
Received status code at openimagesinceptionv3 (TF) test classification: 000
./startup.sh: line 402:    15 Aborted                 (core dumped) ${CURRDIR}/dede -host 0.0.0.0 -port ${SVCPORT}
```

cchadowitz-pf:
[...]

Issue #230, opened 2017-01-03
Unsafe use of Tensorflow and Caffe on same DeepDetect server / GPU


beniz:
Tensorflow running call conflicts on Cuda initialization context with that of other libraries, see:

- https://github.com/tensorflow/tensorflow/issues/916
- https://github.com/tensorflow/tensorflow/issues/2210
- https://github.com/mosessoh/CNN-LSTM-Caption-Generator/issues/2
- https://github.com/mdeff/cnn_graph/issues/1

Tensorflow devs indicate that there's no solving of this in the roadmap.

The `StreamExecutor context` issue is confirmed with DeepDetect when using Tensorflow and Caffe services on the same GPU. More qualification expected in the future.

Current solution: build Tensorflow backend to support CPU only (i.e. with no GPU support built-in).

abhiguru:
Are there performance optimisations for CPU Only ? Intel or ARM, I see a spike on all 8 cores when predicting

beniz:
Caffe and TF backend are optimized for CPU, with parallel operations using all cores. It is always best to use batches.

byronyi:
I don't think sharing 1 GPU between multiple application is generally a good idea. Given TF takes control of the global memory allocation, and all other resources that requires exclusive access, it is better to share your GPU in some other ways, for example, time-sharing :P

PR #292, opened 2017-04-20
Dockerfiles for CPU or GPU builds with Caffe+TF

cchadowitz-pf:
I took the standard CPU and GPU Dockerfiles already provided and added my changes to build with both Caffe and TF backend libs. Since I don't use XGBoost myself and haven't tested the Dockerfiles with it included, I removed that.

The GPU versions build with the following NVIDIA configurations (Ubuntu 16.04, CUDA 8.0, cuDNN 5.1):
- GeForce GTX 1070 (driver 375.20), which I believe is compute capability 6.1
- GeForce GTX TITAN X (driver 375.26), which I believe is compute capability 5.2

This may need changes to be more generalized for other compute capabilities, etc.

Also note this is still affected by #230, #289.

### DIFF
--- a/docker/cpu-caffe-tf/Dockerfile
+++ b/docker/cpu-caffe-tf/Dockerfile
@@ -1,0 +1,108 @@
+FROM ubuntu:16.04
+
+RUN echo 'building CPU DeepDetect image'
+
+MAINTAINER Emmanuel Benazera "beniz@droidnik.fr"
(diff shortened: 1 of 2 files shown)


beniz:
Apologies for the delay, thanks for the dockerfiles. The default Cuda compute numbers might need to be edited based on user hardware, but that's a detail.

PR #327, opened 2017-06-21
Travis with no test

jsaksris:
Add Travis to deepdetect with no test script. 

fixing eigen3 problem based on 
https://github.com/beniz/deepdetect/pull/276


beniz:
Thanks. Several things before integration:

- Please squash all your commits into a single one
- Remove the `.travis.yml.orig` file that seems to be useless
- Can you explain your changes to `Cuda.cmake` ?
- Could you explain your changes to `CMakeLists.txt` ? I see hard-coded paths such as `/usr/include` and I don't think there should be there, unless Travis platforms were complaining in some ways. In that latter case, please update this PR with the Travis error messages.

jsaksris:
Removing hard-coded paths resolved in following error. 
This fix was part of the changes suggested by https://github.com/beniz/deepdetect/pull/276

screen shot 2017-06-26 at 11 45 38 am

PR #276, opened 2017-03-14
Fix eigen3 cmake

victorv:
eigen3 uses pkg_config rather than a module to find eigen3.

PR #328, opened 2017-06-27
add Travis with no test script, fix eigen3

jsaksris:
Add added Travis file.
fixed eigen3 issue according to #276 
omitted changes in Cuda.cmake of #276 as it's not necessary.


jsaksris:
found regression issue prevent the program from building in ubuntu 16.04

close for now.

PR #334, opened 2017-07-12
add service creation option in dd_bench.py

jsaksris:
add service creation option in dd_bench.py

usage
--create [model's folder name]
--nclasses [number of classes; default to 1000]
--auto-kill 

This options are for pre-trained NN.
Meant to streamline the benchmarking process.
Invocation of the server is still required. 

--auto-kill  would allow the user to change the models easily without manually killing the service


beniz:
Thanks!

Issue #343, opened 2017-08-24
Empty SVM line makes DD dump


YaYaB:
I have an issue in prediction when I put an out of bounds svm index in the CURL request.
It makes DD dump.

#### Configuration
- Version of DeepDetect:
    - Locally compiled on Ubuntu 16.04 LTS
- Commit (shown by the server when starting):
    - Branch `Master`, 24f40b6993dedbedd223ced45f3b88dd0762da63
- GPUS:
    - 1 x Nvidia GTX Titan X (Maxwell)

#### Your question / the problem you're facing:
I have an issue in prediction when I put an out of bounds svm index in the CURL request. DD core dumps in that case.

First I launch DeepDetect
INPUT
```
INPUT
./dede --port 8080
```
OUTPUT
```
DeepDetect [ commit 95f235e4f9c9b827e434a65854b962e49b533f3f ]
INFO - 07:35:16 - Running DeepDetect HTTP server on localhost:8080
```

Second, I create the service that I name svm_test.
INPUT
``
curl -X PUT 'http://localhost:8080/services/svm_test' -d '{"mllib":"caffe","description":"svm_test classifier","type":"supervised","parameters":{"input":{"connector":"svm","characters":true},"mllib":{"nclasses":2}},"model":{"repository":"/home/svm_test/dd_models/model_svm"}}'
```

OUTPUT
```
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0824 07:38:43.746436 21717 svminputfileconn.cc:229] loaded SVM vocabulary of size=20147

INFO - 07:38:49 - Initializing net from parameters: 

INFO - 07:38:49 - Creating layer / name=inputl / type=MemorySparseData
INFO - 07:38:49 - Creating Layer inputl
INFO - 07:38:49 - inputl -> data
INFO - 07:38:49 - inputl -> label
INFO - 07:38:49 - Setting up inputl
INFO - 07:38:49 - Top shape: 1 20146 (20146)
INFO - 07:38:49 - Top shape: 1 (1)
INFO - 07:38:49 - Memory required for data: 80588
INFO - 07:38:49 - Creating layer / name=fc_data / type=SparseInnerProduct

...
```

Then I try to make a prediction with a valid index between 0 and 20146.
```
curl -X POST "http://localhost:8080/predict" -d "{\"service\":\"svm_test\",\"data\":[\"20146:0.18\"]}"
```

It works and I obtain the following:
```
[...]

PR #344, opened 2017-08-24
Catch error when svm line is empty

YaYaB:
When doing aprediction on a SVM, indexes of SVM examples should be bounded. When and index is out of those bounds DD core dumps (see the #343)
This PR attempts to correct it by catching an error after reading an SVM line.
Basically it checks if after reading one line from SVM any value was added to the variable `vals`.
At this point in the code, if nothing was caught that means the SVM line seems correct however `vals` might be empty If the index is out of bounds (see line 96 of https://github.com/beniz/deepdetect/blob/master/src/svminputfileconn.cc). If the example has only indexes out of bounds `vals` will be empty and it makes DD core dumps. To avoid that, verifying if `vals` is not empty might be a solution.

### DIFF
--- a/src/svminputfileconn.cc
+++ b/src/svminputfileconn.cc
@@ -106,6 +106,8 @@ namespace dd
 	    throw InputConnectorBadParamException("error reading SVM format line: " + content);
 	  }
       }
+      if (vals.empty())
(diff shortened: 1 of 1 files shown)


beniz:
Thanks for the fix!

PR #384, opened 2018-03-08
Fix calculation of estimated remain_time if remain_time is not available

YaYaB:
If 'remain_time' is not obtained yet, DeepDetect throws an error and stops the training.

By doing this we do not calculate the estimated remaining time when we do not have this information

### DIFF
--- a/src/mllibstrategy.h
+++ b/src/mllibstrategy.h
@@ -234,13 +234,15 @@ namespace dd
     void est_remain_time(APIData &out)
     {
       APIData meas = out.getobj("measure");
-      int est_remain_time = static_cast<int>(meas.get("remain_time").get<double>());
-      int seconds = est_remain_time % 60;
-      int minutes = (est_remain_time / 60) % 60;
-      int hours = (est_remain_time / 60 / 60) % 24;
-      int days = est_remain_time / 60 / 60 / 24;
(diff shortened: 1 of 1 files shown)


src/mllibstrategy.h
```diff
@@ -234,13 +234,15 @@ namespace dd
...
+      if (meas.has("remain_time")){    
+        int est_remain_time = static_cast<int>(meas.get("remain_time").get<double>());
+        int seconds = est_remain_time % 60;
+        int minutes = (est_remain_time / 60) % 60;
+        int hours = (est_remain_time / 60 / 60) % 24;
+        int days = est_remain_time / 60 / 60 / 24;
+        std::string est_remain_time_str = std::to_string(days) + "d:" + std::to_string(hours) + "h:" + std::to_string(minutes) + "m:" + std::to_string(seconds) + "s";
+        meas.add("remain_time_str",est_remain_time_str);
+      }
       out.add("measure",meas);
```
beniz: You don't need to rewrite measure here since it hasn't been changed. Move it within the brackets and it should work the same.

beniz:
Great thanks, see review for a fix.

src/mllibstrategy.h
YaYaB: Yep, indeed.
it's corrected

beniz:
Thanks!

PR #407, opened 2018-05-15
eigen version of multi label soft proba metrics

fantes:
distance correlation is not eigened, I am not sure it can be parallelized, will do different pull request if so


beniz:
Thanks! It'd be best to remove the old code instead of commenting it out.

PR #418, opened 2018-05-22
Audio data training via float storage in cv::Mat

beniz:
Allows taking audio spectrograms as input, and process them through the image pipeline.
Typically, data is stored into fixed size `.exr` files.

New parameters:
- `unchanged_data` to `input` connector to specify the data should not be encoded nor changed when stored in OpenCV mat

New features:
- training of audio classification task
- prediction

Should allow building speech to text models right out of the box via `ctc` layer and loss.


src/backends/caffe/caffeinputconns.cc
```diff
@@ -278,26 +291,29 @@ namespace dd
     std::unique_ptr<db::Transaction> txn(db->NewTransaction());
     
     // Storing to db
-    Datum datum;
     int count = 0;
     const int kMaxKeyLength = 256;
     char key_cstr[kMaxKeyLength];
     
     for (int line_id = 0; line_id < (int)lfiles.size(); ++line_id) {
+      Datum datum;
```
xoolive: Is it necessary to create a new Datum for each file?

src/backends/caffe/caffeinputconns.h
```diff
@@ -342,11 +342,15 @@ namespace dd
...
-	      images_to_db(_uris,_model_repo + "/" + _dbname,_model_repo + "/" + _test_dbname);
+	      if (!this->_unchanged_data)
+		images_to_db(_uris,_model_repo + "/" + _dbname,_model_repo + "/" + _test_dbname);
+	      else images_to_db(_uris,_model_repo + "/" + _dbname,_model_repo + "/" + _test_dbname,
+				"lmdb",false,"");
 	      
 	      // compute mean of images, not forcely used, depends on net, see has_mean_file
-	      compute_images_mean(_model_repo + "/" + _dbname,
-				  _model_repo + "/" + _meanfname);
+	      if (!this->_unchanged_data)
```
xoolive: from the above comment, I am not sure the test is enough... maybe a `todo` ?

beniz: Good catch on the `computer_mean`, it should be active for images. Not sure we want to remove the mean on spectrograms. 

The test is not thorough but should be enough to avoid the mean computation when data should not be unchanged I guess.

src/backends/caffe/caffeinputconns.cc
beniz: Datum is the caffe inner representation of a sample before it is loaded up as a Blob object in  cpu/gpu memory.

src/backends/caffe/caffeinputconns.h
xoolive: so in the future, should it be active on many-channel images?

beniz: Yes I've removed the `compute_mean` since it was using uint8 operations and I should reactivate it before merge!

xoolive: so, the test should be on `float` type rather than on the `unchanged` flag, am I right?

beniz: We need a single flag that says don't touch the data, it's not a true image inside, i.e. don't encode it, etc... Whether it's float or not actually. We already have something similar inside caffe when we're storing labels into images (e.g. for segmentation).

PR #422, opened 2018-05-23
Travis debug gpu 2

fantes:
only one arch, still timeouts 
deactivate cuda build


beniz:
Deactivation due to timeouts.

PR #424, opened 2018-05-28
Image classification with caffe2 backend

chichaj:
This is the first PR for Caffe2 integration into DD.

 - [x] GPU support (i.e. tag blobs in graph for execution on GPU backend) -> best at service creation
 - [x] Add support for `inputlayer` and `outputlayer` parameters in `mllib` to specify the input and output layers
 - [x] Set batch size automatically to the number of elements in the `data` object
 - [x] unit test for predict with caffe2
 - [x] input image mean values (RGB) subtraction

- Features:

Compile with -DUSE_CAFFE2=ON
Clone https://github.com/caffe2/models to get the templates
A corresp.txt file can be placed inside the model repository 
(Use https://gist.github.com/yrevar/942d3a0ac09ec9e5eb3a for most models
and https://gist.github.com/aaronpolhamus/964a4411c0906315deb9f4a3723aac57 for inception_V2.
Warning: They need to be changed into DD format)

```
       "mllib":"caffe2",
       "description":"image classification service",
       "type":"supervised",
       "parameters":{
         "input":{
           "connector":"image"
         },
         "mllib":{
           "template":"resnet50"
         }
       },
       "model":{
         "templates":"$PATH_TO_CAFFE2_MODELS",
         "repository":"$PATH_TO_AN_EXISTING_DIRECTORY"
       }
     }'

       "service":"imageserv",
       "parameters":{
         "input":{
           "width":224,
           "height":224
         },
         "output":{
           "best":3
         }
       },
       "data":["http://i.ytimg.com/vi/0vxOhd4qlnA/maxresdefault.jpg"]
     }'
```

Should output something like:

```
{
  "status": {
    "msg": "Created", 
    "code": 201
  }
}

{
  "status": {
    "msg": "OK", 
    "code": 200
  }, 
  "body": {
    "predictions": [
      {
        "classes": [
          {
            "prob": 0.24751701951026917, 
            "cat": "football helmet"
          }, 
          {
            "prob": 0.1827472597360611, 
            "cat": "gasmask, respirator, gas helmet"
          }, 
          {
            "last": true, 
            "prob": 0.17197568714618683, 
            "cat": "oxygen mask"
          }
        ], 
        "uri": "http://i.ytimg.com/vi/0vxOhd4qlnA/maxresdefault.jpg"
      }
    ]
  }, 
  "head": {
    "method": "/predict", 
    "service": "imageserv", 
    "time": 1407.0
  }
}
```

Note: When creating a resnet50 service, you need to add "std : 255" to the image connector parameters.


src/backends/caffe2/caffe2inputconns.h
```diff
@@ -0,0 +1,80 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2014-2016 Emmanuel Benazera
```
beniz: Not coded by me!!

src/backends/caffe2/caffe2lib.cc
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
+  Caffe2Lib<TInputConnectorStrategy,TOutputConnectorStrategy,TMLModel>::~Caffe2Lib()
+  {
+  }
+
+  template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
+  void Caffe2Lib<TInputConnectorStrategy,TOutputConnectorStrategy,TMLModel>::
+  instantiate_template(const APIData &ad)
```
beniz: `create_model` instead ? We can figure out caffe2 templates later on.

src/backends/caffe2/caffe2lib.cc
```diff
@@ -0,0 +1,219 @@
...
+    std::string predict_net_path, init_net_path;
+    copy("predict_net.pb", predict_net_path);
+    copy("init_net.pb", init_net_path);
+
+    //
+    CAFFE_ENFORCE(caffe2::ReadProtoFromFile(init_net_path, &_init_net));
+    CAFFE_ENFORCE(caffe2::ReadProtoFromFile(predict_net_path, &_predict_net));
+    CAFFE_ENFORCE(_workspace.RunNetOnce(_init_net));
+    //TODO Create all inexistents external inputs
+    _workspace.CreateBlob("gpu_0/data");
```
beniz: Use `inputlayer` instead.

src/backends/caffe2/caffe2lib.cc
```diff
@@ -0,0 +1,219 @@
...
+      inputc.transform(ad);
+    } catch (std::exception &e) {
+      throw;
+    }
+
+    // int batch_size = inputc.test_batch_size();
+    //TODO Get the batch size and the input tensor from value_info.json
+    int batch_size = 1;
+    std::vector<APIData> vrad;
+    auto &tensor_input = *_workspace.GetBlob("gpu_0/data")->GetMutable<caffe2::TensorCPU>();
```
beniz: Use `inputlayer` instead

src/backends/caffe2/caffe2lib.cc
```diff
@@ -0,0 +1,219 @@
...
+
+      float loss(0);
+      std::vector<std::vector<float> > results(image_count);
+      auto result_size(0);
+
+      //Running the net
+      try {
+	CAFFE_ENFORCE(_workspace.RunNetOnce(_predict_net));
+	//TODO Get the external output from the predict_net
+	auto &tensor = _workspace.GetBlob("gpu_0/softmax")->Get<caffe2::TensorCPU>();
```
beniz: CPU / GPU here?

beniz: Maybe avoid `auto` to show types

src/backends/caffe2/caffe2model.cc
```diff
@@ -0,0 +1,65 @@
...
+	if (read_from_repository(ad.get("repository").get<std::string>(),spdlog::get("api")))
+	  throw MLLibBadParamException("error reading or listing Caffe2 models in repository " + _repo);
+      }
+    read_corresp_file();
+  }
+
+  int Caffe2Model::read_from_repository(const std::string &repo,
+					const std::shared_ptr<spdlog::logger> &logger)
+  {
+    static const std::string &corresp = "corresp";
```
beniz: Look for `.pb` files here.

beniz:
FYI, at the moment I'm getting the following error on service creation:

```
{"status":{"code":500,"msg":"InternalError","dd_code":1007,"dd_msg":"[enforce fail at workspace.cc:226] net_def->has_name(). Net definition should have a name. "}}
```

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -29,6 +29,14 @@ namespace dd
   }
 
   void ImgCaffe2InputFileConn::transform(const APIData &ad) {
+    if (_train) {
+      transform_train(ad);
+    } else {
+      transform_test(ad);
```
beniz: `transform_predict` would be more appropriate

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
     _ids = _uris;
   }
 
+  void ImgCaffe2InputFileConn::transform_train(const APIData &) {
+    _shuffle = true;
```
beniz: no need at this stage

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
...
+      num = _images.size(); // No fixed batch_size
     }
-    auto w = _images[0].cols;
-    auto h = _images[0].rows;
+
+    // Resize and prefill with 0s
     tensor.Resize(std::vector<caffe2::TIndex>({num, 3, h, w}));
-    auto data = tensor.mutable_data<float>();
+    float *data = tensor.mutable_data<float>();
     std::fill(data, data + tensor.size(), 0);
```
beniz: I don't believe we need to pre-fill, the tensor should be the exact size already.

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
...
-    auto idx(0), size(0);
+
+    int idx(0), size(0);
```
beniz: rename `size` variable

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
...
       std::vector<cv::Mat> chan(3);
       cv::split(*it, chan);
       it = _images.erase(it);
-      for (auto &ch : chan) {
-	for (auto row = 0; row < h; ++row) {
-	  for (auto col = 0; col < w; ++col) {
+      for (cv::Mat &ch : chan) {
+	for (int row = 0; row < h; ++row) {
+	  for (int col = 0; col < w; ++col) {
 	    data[idx++] = ch.at<unsigned char>(row, col) / 255.0;
```
beniz: Let's not divide by 255 here.

beniz: try `uint8_t`

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
...
+
   int ImgCaffe2InputFileConn::get_tensor_test(caffe2::TensorCPU &tensor, int num) {
     if (!_images.size()) {
-      return 0;
+      return 0; // No more data
+    }
+    int w = _images[0].cols;
+    int h = _images[0].rows;
+    if (num < 0) {
+      num = _images.size(); // No fixed batch_size
```
beniz: Get real tensor batch size at this stage here.

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -37,23 +45,34 @@ namespace dd
...
-    auto data = tensor.mutable_data<float>();
+    float *data = tensor.mutable_data<float>();
     std::fill(data, data + tensor.size(), 0);
-    auto idx(0), size(0);
+
+    int idx(0), size(0);
     for (auto it = _images.begin(); num - size && it != _images.end(); ++size) {
       std::vector<cv::Mat> chan(3);
       cv::split(*it, chan);
       it = _images.erase(it);
```
beniz: Let's erase outside of the loop, avoiding relocation of vector elements

src/backends/caffe2/caffe2lib.cc
```diff
@@ -19,93 +19,194 @@
...
-
-    //
-    CAFFE_ENFORCE(caffe2::ReadProtoFromFile(init_net_path, &_init_net));
-    CAFFE_ENFORCE(caffe2::ReadProtoFromFile(predict_net_path, &_predict_net));
-    CAFFE_ENFORCE(_workspace.RunNetOnce(_init_net));
-    //TODO Create all inexistents external inputs
-    _workspace.CreateBlob("gpu_0/data");
-    CAFFE_ENFORCE(_workspace.CreateNet(_predict_net));
+  create_model() {
+    if (_state.has_changed()) {
```
beniz: maybe put this logic outside of the function

src/backends/caffe2/caffe2lib.cc
```diff
@@ -140,73 +248,84 @@ namespace dd
...
-	  data += result_size;
-	}
-      } catch(std::exception &e) {
-	this->_logger->error("Error while proceeding with supervised prediction forward pass, not enough memory? {}",e.what());
-	throw;
+    try {
+      CAFFE_ENFORCE(_workspace.RunNetOnce(_predict_net));
+      std::string output_blob = _output_blob;
+      if (output_blob.empty()) {
+	output_blob = _predict_net.external_output()[0];
```
beniz: Explain in a short `//` comment

beniz:
Thanks for the update!

At the moment it fails for me with the following error (even with `gpu:false` at service creation):

```
{"status":{"code":500,"msg":"InternalError","dd_code":1007,"dd_msg":"[enforce fail at operator.cc:42] blob != nullptr. op Conv: Encountered a non-existing input blob: gpu_0/data "}}
```

Server output:

```
[2018-05-31 23:02:29.886] [imageserv] [info] Using GPU 0
[2018-05-31 23:02:29.886] [api] [info] ::1 "PUT /services/imageserv" 201 1
[2018-05-31 23:02:35.375] [caffe] [warning] Blob gpu_0/data not in the workspace.
[2018-05-31 23:02:35.375] [imageserv] [error] Error creating model
[2018-05-31 23:02:35.375] [imageserv] [error] prediction call failed
[2018-05-31 23:02:35.375] [api] [error] ::1 "POST /predict" 500 3382
```

Note that the error `gpu_0/data` remains identical even when using `gpuid:2`.

chichaj:
> Thanks for the update!
> 
> At the moment it fails for me with the following error (even with gpu:false at service creation):
> 
> {"status":{"code":500,"msg":"InternalError","dd_code":1007,"dd_msg":"[enforce fail at operator.cc:42] blob != nullptr. op Conv: Encountered a non-existing input blob: gpu_0/data "}}
> 
> Server output:
> 
> [2018-05-31 23:02:29.886] [imageserv] [info] Using GPU 0
> [2018-05-31 23:02:29.886] [api] [info] ::1 "PUT /services/imageserv" 201 1
> [2018-05-31 23:02:35.375] [caffe] [warning] Blob gpu_0/data not in the workspace.
> [2018-05-31 23:02:35.375] [imageserv] [error] Error creating model
> [2018-05-31 23:02:35.375] [imageserv] [error] prediction call failed
> [2018-05-31 23:02:35.375] [api] [error] ::1 "POST /predict" 500 3382
> 
> Note that the error gpu_0/data remains identical even when using gpuid:2.

I forgot to put a comment about that.
Right now the input layer isn't automatically detected and should be precised at the service creation :

```
"mllib": { "inputlayer" : "gpu_0/data" }
```

Concerning the blob names, they stay the same as in the .pb file. Thus it's still gpu_0/... even after being affected on an other device.

beniz:
> "mllib": { "inputlayer" : "gpu_0/data" }

Thanks, it works perfectly with this added to the service creation call! Congrats on getting caffe2 working seamlessly!

beniz:
Two remarks:

- The inception V2 model from https://github.com/caffe2/models/tree/master/inception_v2 uses `data` as the `inputlayer`. Do you believe we could detect the input layer automatically ?

- The model is first loaded upon the first predict call. Since the model is already within the directory at creation call, I believe it should be loaded up at service creation anyways. This is what we do with caffe I think.

fantes:
please add 
OPTION(USE_CAFFE2 "build caffe2 backend") 
OPTION(USE_TF "use TF backend")

on top (3rd line) of main CMakeLists.txt

fantes:
if wrong repo is given at service creation, dede ouputs "Created" and log does not output anything special in logfile. error message is delayed until predict, which is a bit weird. Maybe the service should not be created in such case?

beniz:
> This is also the case with caffe backend, is there a reason for that ?

It's true that with caffe and caffe2 a model directory should either be automatically populated with a template, either already contain the net architecture, etc... This is not true with XGBoost for instance. So the underlying idea is that the service creation is not erroneous, but the training or predict will be / is if some information is missing. Therefore the error is de facto reported to the training or prediction call.

It could be argued otherwise. At minima it makes the service creation identical for all libs...

beniz:
@chichaj Could you modify the README.md so that the caffe2 building steps are included ? Thanks.

beniz:
Thanks for the test. Can you apply this patch to the PR to allow the unit test to run on GPU:

```
diff --git a/tests/ut-caffe2api.cc b/tests/ut-caffe2api.cc
index aea21b0..f5202b3 100644
--- a/tests/ut-caffe2api.cc
+++ b/tests/ut-caffe2api.cc
@@ -47,7 +47,10 @@ using namespace dd;
          "height": 224,                                \
          "width": 224,                                 \
          "std": std                                    \
-       }                                               \
+       },                                              \
+       "mllib": {                                      \
+         "gpu":true                                    \
+        }                                              \
       }                                                        \
     })
```
It works fine for me.

PR #432, opened 2018-06-07
Removed extra section that subtracted mean scalar values from the dat…

cchadowitz-pf:
…um after it was already subtracted from the Mat before it was converted to a datum.

### DIFF
--- a/src/backends/caffe/caffeinputconns.h
+++ b/src/backends/caffe/caffeinputconns.h
@@ -226,20 +226,6 @@ namespace dd
 			}
 		  datum.clear_data();
 		}
-	      else if (_has_mean_scalar)
-		{
-		  int height = datum.height();
-		  int width = datum.width();
-		  for (int c=0;c<datum.channels();++c)
-		    for (int h=0;h<height;++h)
-		      for (int w=0;w<width;++w)
-			{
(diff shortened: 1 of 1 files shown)


beniz:
Hi, thanks. So in fact we opted for reverting our bad commit. We figured that using `cv::subtract` before the conversion to `caffe::Datum` could yield erroneous results. This is because the Datum conversion from caffe (io.cpp / `CVMatToDatum`) uses an 8bit conversion per pixel value. With potential negative values (due to early mean subtraction), this would try to store the value into an unsigned int8...

cchadowitz-pf:
Ahh I see. But couldn't the per-pixel subtraction still produce negative values (at least stored in a float, but still negative)? What's the appropriate behavior in the case of a negative value after subtracting the mean? Clamping to 0, leaving it as a negative, or taking the absolute value?
```
float datum_element = static_cast<float>(static_cast<uint8_t>(datum.data()[data_index]));	
datum.add_float_data(datum_element - _mean[c]);
```

Thanks!

beniz:
Since it's stored in the float array, that's how it reaches the net's first layer, and this is fine.

cchadowitz-pf:
So it's fine for the net's first layer to take in negative values (assuming that's what it's been trained with)? Got it 👍

PR #438, opened 2018-06-13
Generalized dice loss deepdetect api

fantes:
works with new custom caffe

add ability tu use generalized dice (F1) loss for image segmentation, multilabel and optionally weighted

option to mllib: 
- `"loss": "dice"`
- `"loss": "dice_multiclass"`
- `"loss": "dice_weighted"`

this option, used with `"template":"deeplab_vgg16"` or `"template":"unet"`  and `"segmentation":true`  (`input` parameter) generates corresponding architectures with dice loss 
mllib option `ignore_label` is also usable


beniz:
Idem than for #423, would be great to rebase to master, thanks!

PR #452, opened 2018-06-26
Forcing caffe2's compiled protobuffers inside caffe_dd

chichaj:
When using caffe2, caffe_dd does't use its own .proto files, but use the one compiled in libcaffe2_protos.a instead


beniz:
Thanks for this fix + simplified CMakeLists! https://github.com/jolibrain/caffe/pull/25 is now merged, could you update the CAFFE_DD_USER and other variables accordingly ? Tests should then be updated. Thanks!

PR #455, opened 2018-06-28
correctly aggregate means

fantes:
change aggregation of means

- the idea is to compute mean (over test images) of mean (over labels) of average precision (of bounding boxes)

- bug: the APs of one image (for every labels) were summed to previous mAP (supposedly over all images in test set) then divided by the number of labels in the current image, giving a weighted sum (the first map was divided n_test_images times, the second n_test_images -1 by the number of label of current image)

- Now the map is computed over one image (mean over labels) then summed to global mmap, and divided in the end by the number of image, so we have a mean mAP

### DIFF
--- a/src/supervisedoutputconnector.h
+++ b/src/supervisedoutputconnector.h
@@ -1307,14 +1307,16 @@ namespace dd
     
     static double ap(const APIData &ad)
     {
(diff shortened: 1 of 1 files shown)


beniz:
Thank you!

PR #459, opened 2018-07-05
Better error handling for dlib-internal errors

cchadowitz-pf:
And for general uncaught std::exception errors

Also added `mltype` to dlib backend


src/backends/dlib/dliblib.cc
```diff
@@ -37,6 +37,7 @@ namespace dd {
             :MLLib<TInputConnectorStrategy, TOutputConnectorStrategy, DlibModel>(std::move(cl)) {
         this->_libname = "dlib";
         _net_type = cl._net_type;
+        this->_mltype = "detection/classification (" + _net_type + ")";
```
beniz: I recommend we keep the `mltype` identical to the existing ones, e.g. from the Caffe backend.

cchadowitz-pf: 👍 I'll change this to simply `detection`. I noticed afterwards that a `GET /services/<serviceName>/info` returns service-specific parameters anyways, which will show the `_net_type` there. This `_mltype` I believe is shown for `GET /services/info`.

src/services.h
```diff
@@ -529,6 +529,13 @@ namespace dd
 	  pout._status = -3;
 	  throw;
 	}
+	  catch (const std::exception &e)
```
beniz: OK this adds more in-depth granularity to error messages. As a rule-of-thumb, separate PRs when you are modifying core fles. Will merge this one as is though.

PR #460, opened 2018-07-05
Mkdir repo

fantes:
create directory for repository if it does not exist
makes sense only in conjunction with "template" parameter 
```
curl -X PUT "http://localhost:9999/services/calibration" -d '{
       "mllib":"caffe",
       "description":"clothes calibration",
       "type":"unsupervised",
       "parameters":{
         "input":{
           "connector":"image",
           "height":300,
           "width":300
         },
         "mllib":{
           "nclasses":1000, "template": "resnet_50"
         }
       },
       "model":{
         "repository":"/home/infantes/test/calibration",
         "create_repository": true
       }
     }'
```

### DIFF
--- a/src/mlmodel.h
+++ b/src/mlmodel.h
@@ -29,15 +29,24 @@
 #include <fstream>
 #include <iostream>
 #include <unordered_map>
+#include "apidata.h"
+#include "utils/fileops.hpp"
+#include "mllibstrategy.h"
 
 namespace dd
 {
   class MLModel
   {
   public:
-    MLModel() {};
-    MLModel(const std::string &repo)
-      :_repo(repo) {
+    MLModel() {}
+    MLModel(const APIData &ad)
+      {
+        init_repo_dir(ad);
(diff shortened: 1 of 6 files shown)


src/mlmodel.h
```diff
@@ -29,16 +29,31 @@
...
+#include "utils/fileops.hpp"
+#include "mllibstrategy.h"
 
 namespace dd
 {
   class MLModel
   {
   public:
-    MLModel() {};
+    MLModel() {}
```
beniz: Do we really need an empty constructor ?

src/mlmodel.h
```diff
@@ -29,16 +29,31 @@
 #include <fstream>
 #include <iostream>
 #include <unordered_map>
+#include "apidata.h"
+#include "utils/fileops.hpp"
+#include "mllibstrategy.h"
```
beniz: Do we need to include mllibstrategy ?

PR #464, opened 2018-07-12
Added training with Caffe2's backend

chichaj:
Features (with caffe2):

- Net training:
  - From a template ("mllib": {"template": ... })
  - From an already existing model:
     - Use the flags {"mllib": {"finetuning": true}, "model": {"weights": ...}} where weights is a init_net.pb file
     - The last weights of the net are reset/reshaped to fit the requested number of classes
  - Supervised & Image only
  - Creation of a formatted classification dataset from an image directory
  - Snapshot ("mllib": {"solver": {"snapshot" : ... }}):
     - Backup the training state at regular interval
     - Can be resumed ("mllib": {"resume": ...})

- Net testing ("output" :  {"measure" : ...}):
  - Supervised only
  - With a single 'predict' call
  - At regular interval during training

- GPU Parallelization

Examples :

Creation
```
  "model": {
    "repository": "./some_repository"
  }, 
  "type": "supervised", 
  "mllib": "caffe2", 
  "parameters": {
    "input": {
      "connector": "image", 
      "width": 224, 
      "std": 255.0, 
      "height": 224
    }, 
    "mllib": {
      "gpu": true, 
      "gpuid": [ 0, 1, 2 ],
      "nclasses": 1000
    }
  }, 
  "description": "myclassifier"
}'
```

Train
```
curl -X POST "http://localhost:8080/train" -d '{
  "async": true, 
  "data": [
    "./folder_with_images"
  ], 
  "parameters": {
    "input": {
      "shuffle": true, 
      "db": true, 
      "test_split": 0.1
    }, 
    "mllib": {
      "net": {
        "test_batch_size": 32, 
        "batch_size": 32
      }, 
      "solver": {
        "base_lr": 0.001, 
        "snapshot": 20, 
        "iterations": 30, 
        "test_interval": 20, 
        "gamma": 0, 
        "solver_type": "sgd"
      }
    }, 
    "output": {
      "measure": [
        "acc"
      ]
    }
  }, 
  "service": "imageserv"
}
```

To resume, the same call should be made, but with ``` "resume": true ``` inside "mllib"

Informations
```
curl -X GET "http://localhost:8080/train?service=imageserv&job=1&timeout=20"
```


src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -19,72 +19,447 @@
...
+      } catch (...) {
+	_logger->warn("Could not load {}", file.first);
+	continue;
+      }
+
+      // Put in db
+      auto txn = db->NewTransaction();
+      int length = snprintf(key_cstr, kMaxKeyLength, "%08d_%s", count, file.first.c_str());
+      txn->Put(std::string(key_cstr, length), out);
+      txn->Commit();
```
beniz: Don't commit after every addition, as it is too slow. Instead use the 1000 images check and commit there 1000 images at a time. See how it is done in `caffeinputconns.cc`.

src/backends/caffe2/caffe2inputimg.cc
```diff
@@ -19,72 +19,447 @@
...
+	_logger->error("Missing training inputs");
+	throw;
+      }
+
+      _db = _default_db;
+      _is_testable = fileops::file_exists(_db);
+    }
+
+    compute_db_sizes();
+    compute_images_mean();
```
beniz: You may want to check whether the `mean.pb` file already exists.

src/backends/caffe2/caffe2lib.cc
```diff
@@ -87,176 +65,416 @@ namespace dd {
...
+    _state.set_is_training(train);
+    if (force_init) {
+      _state.force_init();
     }
-    CAFFE_ENFORCE(!ids.empty());
-    for (int id: ids) {
-      this->_logger->info("Using GPU {}", id);
+  }
+
+  //XXX When classifying, the last layer is currently not reshaped based on _nclasses
```
beniz: The layer before the softmax needs to be reshaped based on `_nclasses`. Otherwise the model templates cannot be used for problems that expect a number of classes not exactly the same as the template.
I believe the reshape is mandatory, unfortunately.

beniz:
Hi, I've started testing :) Basic training with SGD seems to work successfully :)

I can't get the other optimizers (rmsprop, adam, adagrad, ...) to work properly (loss doesn't decrease). I've tried various learning rates, and this changes nothing, and thus the problem might be elsewhere.

Could you check that all optimizers are correctly wired in ? (From a quick look at the code, it seems it's fine, but I'm not certain). In your own tests, including using Python + Caffe2, can you test the other optimizers and report ? 

Thanks!

beniz:
Wonderful job and thanks for the fixes and the finetuning! This works fine for me. Merging it now. Subsequent fixes and improvements can be achieved through small additional PRs. Also merging allows moving forward with #490.

PR #472, opened 2018-07-17
deeplab_vgg16 finetune

fantes:
add special case for finetuning deeplab_vgg16 
tested ok


fantes:
unable to reproduce error... 
i can rename more layers (ie norm, and final_interp) but it should not be needed as they do not save any parameters

PR #488, opened 2018-08-21
Add -DUSE_RPI3 cmake flag for RPi3 compatibility

CorentinB:


beniz:
Can you replace the ARM flag by a RPI3 flag ? Jetson is also an arm based platform and we have a JETSON flag already, so having flags by board sounds better at the moment. Thanks!

CorentinB:
It's done, but maybe the same libraries paths can be used for other version of the RPi, such as the 1 and 2, so maybe with more testing this flag could be the RPI flag.

CorentinB:
Closing this pull request, I'm going to reopen it a cleaner way.

PR #489, opened 2018-08-27
CNN autoencoder support for Caffe backend

beniz:
Support for convolutional autoencoder on images, including:
- Deconvolution neural net generator layer
- Upsampling neural net generator layer
- Hourglass architectures with control from API


beniz:
This should be good to go, @fantes if you could review!

PR #491, opened 2018-09-03
Replaced "TensorProtoDBInput" operators by "ImageInput"s

chichaj:
Caffe2 offers several ways to do data augmentation via the "ImageInput" operator.

https://caffe2.ai/docs/operators-catalogue.html#imageinput

That's why the ImgCaffe2InputFileConn class now appends an "ImageInput" into the training net.

The following flags are supported (you can find their description and default value in the link above)

- color_jitter (bool)
- img_saturation (float)
- img_brightness (float)
- img_contrast (float)
- color_lighting (bool)
- color_lighting_std (float)
- scale_jitter_type (int)
- mirror (bool)

They can be set in the "img_aug" componant of the input connector:

```
"parameters": {
  "input": {
    "connector": "image",
    "img_aug": {
      "mirror": true,
      "color_jitter": true,
      "img_saturation": 0.7
    }
  }
}
```

The format used when creating a database is also changed to allow this new operator to load the data.


beniz:
Minor point: `scale_jitter_type` should be a bool I believe.

PR #494, opened 2018-09-10
Add USE_HDF5 flag

CorentinB:
USE_HDF5 flag for compilation and usage without `libhdf5`.

Note: need merging of https://github.com/jolibrain/deepdetect/pull/493 and https://github.com/jolibrain/caffe/pull/28 before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,13 +71,15 @@ set(eigen_archive_hash "50812b426b7c")
 include_directories("${EIGEN3_INCLUDE_DIR}")
 
 # hdf5
-if (RPI3)
-  set(HDF5_LIB /usr/lib/arm-linux-gnueabihf/hdf5/serial)
-else()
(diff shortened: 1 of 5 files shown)


beniz:
@fantes @chichaj is hdf5 still on by default ? We don't want to modify the default build values here.

chichaj:
> @fantes @chichaj is hdf5 still on by default ? We don't want to modify the default build values here.

If we want it by default we must explicitly write at the beginning of the file :
```
option(USE_HDF5 "use HDF5" ON)
```
https://cmake.org/cmake/help/v3.0/command/option.html

CorentinB:
> > @fantes @chichaj is hdf5 still on by default ? We don't want to modify the default build values here.
> 
> If we want it by default we must explicitly write at the beginning of the file :
> 
> ```
> option(USE_HDF5 "use HDF5" ON)
> ```
> 
> https://cmake.org/cmake/help/v3.0/command/option.html

Thanks @chichaj I fixed that in the last commit.

PR #504, opened 2018-10-08
Train images models from txt files

YaYaB:
This PR allows to train images using data from a text file corresponding to the following:
```
path/to/img1 label1
path/to/img2 label2
path/to/img3 label3
```
The main reason of this PR is to facilitate the management of datasets by avoiding having duplicates data when we want to train on several parts of a whole big dataset.
Instead of creating symlinks or copies of images into a folder having the following needed tree;
```
-DATA
  - label1
     - Image1
     - Image2
     - etc.
  - label2
     - Image1
     - Image2
     - etc.
  - etc.
```
One can store images in a fix place and work with txt files containing paths to the images.

The calls to DD remain the same, the only change that need to be made is in the field **data**. Instead of indicating a path of a *root_folder* containing a sub directory per class you'll just have to put the path to the text file

### DIFF
--- a/src/backends/caffe/caffeinputconns.cc
+++ b/src/backends/caffe/caffeinputconns.cc
@@ -166,8 +166,7 @@ namespace dd
 				std::istringstream iss(line);
 				string filename;
 				string label;
-				iss >> filename;
-				iss >> label;
+				iss >> filename >> label;
 
 				int label_int = cl;
 				std::unordered_map<std::string,int>::const_iterator it;
--- a/src/backends/caffe/caffeinputconns.cc
+++ b/src/backends/caffe/caffeinputconns.cc
@@ -84,11 +84,15 @@ namespace dd
   }
   
   // convert images into db entries
-  // a root folder must contain directories as classes holding image
-  // files for each class. The name of the class is the name of the directory.
-  int ImgCaffeInputFileConn::images_to_db(const std::vector<std::string> &rfolders,
+  // a path must:
+	// - either contain directories as classes holding image
+  //   files for each class. The name of the class is the name of the directory.
+	// - either be a file containing in each line the path to an image
+	//   and the label associated
+  int ImgCaffeInputFileConn::images_to_db(const std::vector<std::string> &rpaths,
 					  const std::string &traindbname,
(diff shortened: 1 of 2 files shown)


src/backends/caffe/caffeinputconns.cc
```diff
@@ -145,7 +153,42 @@ namespace dd
...
+			std::ifstream infile(rpaths.at(0));
+			std::string line;
+			int line_num = 1;
+			int cl = 0;
+			while (std::getline(infile, line))
+			{
+				std::istringstream iss(line);
+				string filename;
+				string label;
+				CHECK(iss >> filename) << "Error reading line " << line_num;
```
beniz: If this is the CHECK from Caffe, let's avoid it (as it's overriden across backends), and use a straight `if`.

src/backends/caffe/caffeinputconns.h
```diff
@@ -385,10 +385,14 @@ namespace dd
 		return;
 	      }
 	      // create db
-	      if (!this->_unchanged_data)
-		images_to_db(_uris,_model_repo + "/" + _dbname,_model_repo + "/" + _test_dbname);
-	      else images_to_db(_uris,_model_repo + "/" + _dbname,_model_repo + "/" + _test_dbname,
-				"lmdb",false,"");
+				// Check if the indicated uri is a folder
+				bool dir_images = true;
+				bool exists= fileops::file_exists(_uris.at(0), dir_images);
```
beniz: Not sure what this variable is used for, as I don't see it elsewhere in the diff. Do we need it ?

beniz:
Thanks, see my few comments inline. Also, it'd be great if you could fix the indent so that it fits the existing one!

src/backends/caffe/caffeinputconns.h
YaYaB: Indeed, it's not used, I deleted it

src/backends/caffe/caffeinputconns.cc
YaYaB: I removed the CHECK as it will be catched anyway.

Issue #505, opened 2018-10-10
Finetuning impossible


YaYaB:
I have an issue trying to finetune a model.

#### Configuration
- Version of DeepDetect:
    - Locally compiled on Ubuntu 16.04 LTS
- Commit (shown by the server when starting):
    - Branch `Master`, c67f7338c7db276146347d2f3cd3dc1d0c5cd497
- GPUS:
    - 1 x Nvidia GTX Titan X (Maxwell)

#### Your question / the problem you're facing:
I am not able to finetune a model using DeepDetect. I've updated my DD version that was a bit old and now when I try to finetune a model I get the following error

```
[2018-10-10 12:33:32.333] [img] [info] instantiating model template googlenet
[2018-10-10 12:33:32.333] [img] [info] source=../templates/caffe/googlenet/
[2018-10-10 12:33:32.333] [img] [info] dest=/home/repo/googlenet.prototxt
[2018-10-10 12:33:32.339] [img] [error] service creation mllib bad param: finetuning requires specifying an existing weights file
```

However in my request I specify the **weights**. I did an investigation and I tried a build version from older commits and I finally found the commit that broke this feature.
It is the following one f4d27300633b7fa4a3920bc100ecee7dde013890.

I don't understand the purpose of that commit, because it seems to replace the **weights** and **state** argument given in the CURL request even by something that is empty.

fantes:
http://github.com/fantes/deepdetect.git debug_renewal (PR'ed) 
should add correct check

beniz:
@fantes Not sure #506 does the job. The issue as far as I can understand it, is that they may have a model repository with multiple checkpoints in it. They try selecting one for testing or finetuning typically focing it using the  `weights` API parameter. 

__That being said, I do not understand how the finetuning can fail though, since a checkpoint file should be detected whatsover, unless the file specified by `weights` is not within the model directory.__

@fantes in your PR  `weightsf` grabs the most recent checkpoint and overrides any other option.

My initial commit was introduced to allow renewal of the loaded model, typically loading up the last model checkpoint automatically, replacing finetuned weights.

In practice, testing multiple checkpoints is somewhat a corner case since we now have the automated handling of the `best_model.txt` file from which we can get the best model checkpoint in a single lookup. 

However, API-wise, we may want the `weights` value to stick, but this is more involved.

@fantes @YaYaB let me know if I'm missing something.

YaYaB:
@beniz This is not the issue we had.
[...]

PR #506, opened 2018-10-11
Debug renewal

fantes:
replace weights from request only if most recent weights are already present in the repo


beniz:
Not appropriate at the moment, see #505

PR #507, opened 2018-10-17
Added mask support for detectron models

chichaj:
Added a ```'mask': true``` flag.

More detail in the [README.md](https://github.com/chichaj/deepdetect/tree/detectron_mask/examples/caffe2/detectron)

A complete example in the [masks.md](https://github.com/chichaj/deepdetect/tree/detectron_mask/examples/caffe2/detectron/masks.md)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,7 +185,16 @@ if (USE_CAFFE2)
     ${CMAKE_BINARY_DIR}/patches/caffe2/eigen.patch
     ${CMAKE_BINARY_DIR}/patches/caffe2/logging.patch
     ${CMAKE_BINARY_DIR}/patches/caffe2/VLOG.patch
-    )
+    ${CMAKE_BINARY_DIR}/patches/caffe2/collect_proposals.patch
+  )
+  list(APPEND CAFFE2_OPS
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/bbox_to_roi_op.h
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/bbox_to_roi_op.cc
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/segment_mask_op.h
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/segment_mask_op.cc
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/multi_level_roi_op.h
+    ${CMAKE_BINARY_DIR}/patches/caffe2/custom_ops/multi_level_roi_op.cc
+  )
 
   list(APPEND PYTORCH_FLAGS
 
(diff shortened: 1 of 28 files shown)


chichaj:
> Do you believe we could get the conversion from pkl to pb and mask conversion as well working without a third-party install of Detectron, and using the caffe2 and pytorch compiled from within DD ? If some detectron stuff would be missing, would it be possible to use the minimum stuff so that the conversion scripts work ?
> 
> Typically I get a
> 
> ```
> No module named caffe2.python.utils
> ```
> 
> error when no Detectron is install, and the error suggests that caffe2 Python utils might be enough.

*About Detectron itself*

The conversion from pkl to pb uses several tools that cannot be found outside of the Detectron repository. Three of them are particularly important:

The first is a large set of functions able to generate a fast-rcnn, a mlp, a fpn, ...

Inside a detectron .yml configuration file is stored the names and paths of theses functions (and not some kind of protobuf-like scheme). It's by executing them in the right order (and with the help of a custom 'Model Helper') that the model is generated, and then converted into protobuf.

The second is as global dictionay that contains most of the model's configuration. What is stored inside the .yml file is only what differs from this configuration.

And the third is a set of function able to interpret what is stored inside the .pkl (that contains python object structures).

None of theses are directly present inside the script.

*About the need to build Detectron*

If one just want to use the conversion script, there is indeed no need to install or build detectron. But two types of import will break : the cocoapi backend and the cython files that where not compiled into .so.

As they are not used, just commenting out (or removing) the lines importing 'detectron.utils.boxes' or 'pycocotools' allows the script to be run without an install.

*About Deepdetect*

When building pytorch, the entirety of the python modules is controlled by ```-DBUILD_PYTHON=OFF```. By removing this flag from DeepDetect's CMakeLists.txt, all the files needed by Detectron are compiled (at least those needed by the script we want to run).

*Misc*

Here is a short script to convert a detectron model, by using only a build of pytorch (no install of COCOAPI, Detectron, etc.).

```
#!/bin/sh

PYTORCH=/home/jchicha/dev/pytorch/build # Path to a build of pytorch
DETECTRON=/home/jchicha/tmp/nobuild/detectron # Where detectron will be downloaded

CFG=/home/jchicha/dev/caffe2-tests/detectron/e2e_mask_rcnn_R-50-FPN_2x.yaml # yml file
WEIGHTS=/home/jchicha/dev/caffe2-tests/detectron/model_final.pkl # pkl file
OUTPUT=out # Where the model will be converted

git clone 'https://github.com/facebookresearch/detectron' $DETECTRON
PATTERN="^\(from\|import\).*\(detectron\.utils\.boxes\|pycocotools\)" # Broken imports
grep -rl "$PATTERN" $DETECTRON | xargs sed -i "/$PATTERN/d"

PYTHONPATH=$PYTORCH:$DETECTRON python \
	  $DETECTRON/tools/convert_pkl_to_pb.py --out_dir $OUTPUT --cfg $CFG \
	  DOWNLOAD_CACHE $DETECTRON TRAIN.WEIGHTS $WEIGHTS TEST.WEIGHTS $WEIGHTS
```

beniz:
By testing the PR: masks need to be rescaled with nearest-neighbor upsampling after they are produced. This is because when the input image is not exactly the size required by the network, the output mask need to be scaled up (or down). We already do this for bounding boxes, and so we should do it for masks as well!

PR #509, opened 2018-10-23
Fixed weights parameter upon resume and finetuning

beniz:
Fix to the `weights` parameter that was overridden by newest weight file, see #505


YaYaB:
Tested it, sounds good to me!

beniz:
Upon a second review I don't believe this was correct. A fix has been pushed.

YaYaB:
Tested and works as well on my side to finetune

PR #510, opened 2018-10-24
Change pytorch version

chichaj:
Now compiles with pytorch v1.0rc1


beniz:
Build fails with
```
ImportError: No module named yaml
```
Updating JenkinsFile should do it.

PR #517, opened 2018-11-09
long awaited timeseries

fantes:
- time series input connector in the form of directories of csv files (1 csv per time serie)
- db or not db depending on the size of data (user pref)
- generator of multi stack lstm network
- multi dimensional regression  (ie predict several columns with other as inputs)
- no support for discrete column values (not yet)

API for service creation
```
curl -X PUT "http://localhost:8080/services/ts" -d '{
  "mllib":"caffe",
  "description":"CSV TS test",
  "type":"supervised",
  "parameters":{
       "input":{
          "connector":"csvts",
          "db": false,
          "label" : ["target0", "target1"],
          "ignore": ["date"]
       },
       "mllib":{
            "template":"recurrent", 
            "layers":["L10","L10","L5"], 
            "init": "uniform",
            "dropout":0.0, 
            "regression":true, 
            "loss":"L1",
            "db": false
        }
  },
  "model":{
      "repository":"/home/infantes/test/csvts", "templates":"/home/infantes/deepdetect/templates/caffe"
  }
}'
```

- `label` : list of columns names to predict
- `ignore` : list of column names not to take into account
- `loss`: L1 or L2 (euclidean L1 stringly recommended)
- `layers` gives the shape of the network: L50 L50 L50 for 3 layers of lstms with internal hidden state of size 50, can be R for simple RNN
- `init` let user choose the initial filler for weights. supproted values are from caffe values : `uniform` (bounds auto-computed as in pytorch); `xavier`; `msra` 
- `dropout` is a vector of optional dropouts between recurrent layers, or a single double  can be used if same value of dropout everywhere

API for training:
```
curl -X POST "http://localhost:8080/train" -d '{
       "service":"ts",
       "async":true,
       "parameters":{
         "mllib":{
           "gpu":true,
            "resume":false,
            "timesteps" : 200,
           "net":{
             "batch_size":100,"test_batch_size":10
           },
           "solver":{
             "test_interval":10,"snapshot":1000,
             "iterations":10000,
             "base_lr":0.0001,
             "solver_type":"RMSPROP", "test_initialization":false
           }

         },
         "input":{
           "shuffle": true, "separator":",",
           "db": false, 
           "offset": 100,
           "scale": true,
           "min_vals": [-1.0, -2.0, 4.0, 15.0],
           "max_vals":[1.0, -1.0, 10.0, 15.5]
         },
         "output":{
           "measure":["L1"]
         }
       },
       "data":["/home/infantes/deepdetect/build/examples/all/sinus/train", "/home/infantes/deepdetect/build/examples/all/sinus/test"]
}'
```
- `timesteps` is the length of series given to the network, all considered independent, ie no dependencies needing more than timesteps history can be learned
- `offset` is the offset beween start point of sequences, here the connector will generate sequences of length 200 starting from 0, starting from 100, starting from 200 starting from 300 and so on
- `scale` forces the connector to normalize data so that the model processes internally only values between 0 and 1. It dumps min and max values to file `bounds.dat` for prediction. 
- `min_vals` and `max_vals` are to be given only for non-ignored columns, if not given they are computed from training data. 

API for predict:
```
       "service":"ts",
       "parameters":{
         "input":{
             "connector":"csvts",
             "scale": true,    
             "db": false,
            "timesteps":100,
              "separator":";",
               "continuation": false
         },
         "output": {},
          "mllib":{"test_batch_size":1}},
          "data":["/home/infantes/deepdetect/build/examples/all/sinus/predict"]
}'
```
In order to predict a sequence of arbitrary length N: 
set `batch_size`  to 1 (a new series is considered the following of a previous one), 
set `timesteps` to any exact divisor of N (input connectors repeats the end of the sequence  in case of inexact division in order to avoid empty datums in learning) 

If `scale` is set to true, the connectors reads bounds.dat from the repository, scales inputs and unscale outputs

In order to predict several times (ie. call predict as data comes in a loop without forgetting previous hidden state)
same as above + set `continuation` to true


beniz:
Please rebase on master!

fantes:
not yet ready to merge, will rebase later

beniz:
Thanks, that's what I'd call a Deep-PR :)

One tiny remark before I embark on a review of the code: where do the `max_vals`, `min_vals` values come from in the `POST /train` call ?

src/backends/caffe/caffelib.cc
```diff
@@ -1823,6 +1890,11 @@ namespace dd
 	  ad_res.add("regression",_regression);
 	if (inputc._ctc)
 	  ad_res.add("net_meas",true);
+       if (typeid(inputc) == typeid(CSVTSCaffeInputFileConn))
+         {
+           ad_res.add("timeserie",true);
+           ad_res.add("timeseries",nout);
```
beniz: Do we need a special name for the output ? If it's a vector for instance, should be reuse names we use elsewhere, typically `vals` for an array of float/double ? Or is `nout` a special object, e.g. with timestamps ?

src/backends/caffe/caffelib.cc
```diff
@@ -2304,6 +2419,65 @@ namespace dd
 		  vrad.push_back(outseq);
 		}
 	    }
+           else if (typeid(inputc) == typeid(CSVTSCaffeInputFileConn)) // timeseries
+             {
+               int slot = findOutputSlotNumberByBlobName(_net,"OUTPUT");
```
beniz: Same as above, maybe find a better name than `OUTPUT`

src/backends/caffe/caffelib.cc
```diff
@@ -2304,6 +2419,65 @@ namespace dd
...
+                         // new seq -> push old one
+                         {
+                           if (series.size() > 1)
+                             {
+                               APIData out;
+                               if (!inputc._ids.empty())
+                                 out.add("uri",inputc._ids.at(serieNum++));
+                               else out.add("uri",std::to_string(serieNum++));
+                               out.add("series", series);
+                               out.add("probs",std::vector<double>(series.size(),1.0));
```
beniz: Do we get confidences for every timestep prediction ? That'd be great! If only useful part of the time we may want to (later) reuse the `confidences` API parameter from PR #530

src/csvinputfileconn.cc
```diff
@@ -413,11 +410,11 @@ namespace dd
 	}
 
       // shuffle before possible test data selection.
-      shuffle_data(ad);
+      shuffle_data(_csvdata, forbid_shuffle);
```
beniz: That's a bit weird from the outside, call `shuffle`, but `forbid` it via an option :) We'd better control the call from the outside, no ?

src/csvinputfileconn.h
```diff
@@ -277,44 +292,33 @@ namespace dd
...
-	      g = std::mt19937(ad.get("seed").get<int>());
-	    }
-	  else
-	    {
-	      std::random_device rd;
-	      g = std::mt19937(rd());
-	    }
-	  std::shuffle(_csvdata.begin(),_csvdata.end(),g);
-	}
+      if (_shuffle && !forbid_shuffle)
```
beniz: I don't like it too much, let's control the call to `shuffle_data()` instead, if possible.

src/csvinputfileconn.cc
```diff
@@ -257,9 +257,35 @@ namespace dd
     if (!_id.empty() && !has_id)
       throw InputConnectorBadParamException("cannot find id column " + _id);
   }
+
+
+  void CSVInputFileConn::find_min_max(std::ifstream &csv_file)
```
beniz: When calling this, we should be careful we're looking for bounds over the training *and* testing sets. This is because otherwise some values within the test set can be found to be out of bounds at test time.

src/csvtsinputfileconn.cc
```diff
@@ -0,0 +1,290 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2015 Emmanuel Benazera
+ * Author: Emmanuel Benazera <beniz@droidnik.fr>
```
beniz: I am not the author, you are ! :)

src/csvtsinputfileconn.cc
```diff
@@ -0,0 +1,290 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2015 Emmanuel Benazera
```
beniz: Let's put 2018-2019

src/csvtsinputfileconn.h
```diff
@@ -0,0 +1,129 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2015 Emmanuel Benazera
```
beniz: 2018-2019

src/csvtsinputfileconn.h
```diff
@@ -0,0 +1,129 @@
...
+ * Author: Emmanuel Benazera <beniz@droidnik.fr>
```
beniz: Actually beautiful, writing code in my sleep :)

src/csvtsinputfileconn.h
```diff
@@ -0,0 +1,129 @@
...
+    /*   tokenizer<boost::escaped_list_separator<char> tok(s,els); */
+    /*   for (const auto &t : tok) */
+    /*     std::cout << t << '\n'; */
+    /* } */
+
+    void response_params(APIData &out);
+
+
+    void push_csv_to_csvts(DDCsv &ddcsv);
+    virtual void push_csv_to_csvts(bool is_test_data=false);
```
beniz: Let's kill the virtuality here :)

src/generators/net_caffe.cc
```diff
@@ -184,13 +185,13 @@ namespace dd
     lparam = CaffeCommon::add_layer(this->_net_params,"",top,"inputl",
 				    inputc._sparse ? "MemorySparseData" : "MemoryData",label); // test layer
     caffe::MemoryDataParameter *mdparam = lparam->mutable_memory_data_param();
-    mdparam->set_batch_size(32); // dummy value, updated before training
+    mdparam->set_batch_size(batch_size); // dummy value, updated before training
```
beniz: Thanks.

src/generators/net_caffe.cc
```diff
@@ -309,6 +310,36 @@ namespace dd
     //fparam->set_value(0.2); //TODO: option
   }
 
+
+
+  void NetLayersCaffe::add_lstm(caffe::NetParameter *net_param,
```
beniz: Thanks for the addition, this is directly useful to other tasks such as OCR etc...

src/generators/net_caffe_recurrent.cc
```diff
@@ -0,0 +1,371 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2014-2016 Emmanuel Benazera
```
beniz: 2018-2019

src/generators/net_caffe_recurrent.cc
```diff
@@ -0,0 +1,371 @@
...
+ * Author: Emmanuel Benazera <beniz@droidnik.fr>
```
beniz: That guy is a beast!! Writing so much code :)

src/generators/net_caffe_recurrent.cc
```diff
@@ -0,0 +1,371 @@
...
+    if (dropout_ratio ==0)
+      lparam->add_top(top);
+    else
+      lparam->add_top(ctype+std::to_string(id)+"_undropped");
+    lparam->add_bottom(bottom_seq);
+    lparam->add_bottom(bottom_cont);
+    caffe::RecurrentParameter *rparam = lparam->mutable_recurrent_param();
+    rparam->set_num_output(num_output);
+    caffe::FillerParameter *weight_filler_param = rparam->mutable_weight_filler();
+    weight_filler_param->set_type("uniform");
```
beniz: OK, that's not used very often these days, is that fine initialization for recurrent layers ?

src/generators/net_caffe_recurrent.h
```diff
@@ -0,0 +1,83 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2014-2016 Emmanuel Benazera
```
beniz: 2018-2019

src/generators/net_caffe_recurrent.h
```diff
@@ -0,0 +1,83 @@
...
+ * Author: Emmanuel Benazera <beniz@droidnik.fr>
```
beniz: I might go see my doctor and skip work today, I can't remember writing this piece!

src/backends/caffe/caffelib.cc
```diff
@@ -870,12 +897,13 @@ namespace dd
...
     else // model template instantiation is defered until training call
       {
+
 	update_deploy_protofile_softmax(ad); // temperature scaling
 	create_model(true);
       }
 
     // the first present measure will be used to snapshot best model
-    _best_metrics = {"map", "meaniou", "mlacc", "delta_score_0.1", "bacc", "f1", "net_meas", "acc"};
+    _best_metrics = {"map", "meaniou", "mlacc", "delta_score_0.1", "bacc", "f1", "net_meas", "acc", "eucll"};
```
beniz: You are using L1 loss most often as it seems. Do we want `eucll` as best metrics here ?

fantes:
> Thanks, that's what I'd call a Deep-PR :)
> 
> 

One tiny remark before I embark on a review of the code: where do the `max_vals`, `min_vals` values come from in the `POST /train` call ?
either from external preprocessing (useful in order for min an max to be computed from all data, dede can compute them on train data only, not on test data), or from some previous call to train

src/backends/caffe/caffeinputconns.h
```diff
@@ -792,6 +796,153 @@ namespace dd
...
+      CSVTSInputFileConn::fillup_parameters(ad_input);
+      if (ad_input.has("timesteps"))
+        {
+          _timesteps= ad_input.get("timesteps").get<int>();
+          _offset = _timesteps;
+        }
+      if (ad_input.has("continuation"))
+        _continuation= ad_input.get("continuation").get<bool>();
+      if (ad_input.has("offset"))
+        _offset= ad_input.get("offset").get<int>();
```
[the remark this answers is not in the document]
fantes: if no _offset is given, it should default to _timesteps
basically, the connector builds sequences of length _timesteps, and starting points are at i*_offset (for instance sequences of length 100 for offset of 10:  [0:100] [10:110] [20:120] ...
default is to build non overlapping sequences, [0:100][100:200]... , ie offset = timesteps

src/backends/caffe/caffelib.cc
```diff
@@ -58,6 +59,7 @@ namespace dd
     _nclasses = cl._nclasses;
     _regression = cl._regression;
     _ntargets = cl._ntargets;
+    _targets = cl._targets;
```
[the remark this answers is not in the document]
fantes: _targets contains the colums number of targets, it can be [5,7,9] meaning that columns 5,7 and 9 are all outputs of the LSTM

src/backends/caffe/caffelib.cc
```diff
@@ -1571,17 +1615,21 @@ namespace dd
...
 	      slot = 0; // flatten output
            else if (_regression && _ntargets == 1 && typeid(inputc) == typeid(CSVCaffeInputFileConn))
              {
-               slot = findOutputSlotNumberByBlobName(net, "ip_losst");
+               slot = findOutputSlotNumberByBlobName(net, "ip_loss");
              }
 
+           if (typeid(inputc) == typeid(CSVTSCaffeInputFileConn))
+             {
+               slot = findOutputSlotNumberByBlobName(net,"OUTPUT");
```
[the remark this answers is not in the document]
fantes: it is the one given by generator/net_caffe_recurrent.cc, we can change it if needed

src/backends/caffe/caffelib.cc
fantes: it is a vector of vectors, [num_outputs * length of timeseries] so it seemed that a different name may avoid confusion

src/generators/net_caffe_recurrent.cc
fantes: it is the one used in pytorch ...

src/csvinputfileconn.cc
fantes: _adconf was only needed to pass "shuffle" parameter, but shuffling has been refactored a lot in order to allow to shuffle series themselves (ie order of series) but not their content (at it follows time), also to shuffle cvs (if not timeries) hence its removal here

src/backends/caffe/caffeinputconns.cc
```diff
@@ -1175,7 +1177,7 @@ namespace dd
   {
     if (_cifc)
       {
-	_cifc->read_csv(_adconf,fname);
```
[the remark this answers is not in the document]
fantes: _adconf was used only to pass "shuffle" parameter, but shuffling has been under major refactoring in order to be able to shuffle timeseries order but not their content sugin the cvs connector functions (whereas cvs connector shuffle lines, ie. points)

src/backends/caffe/caffelib.cc
fantes: L1 metric is not yet implemented ...

fantes: putting in my todo list

src/backends/caffe/caffelib.cc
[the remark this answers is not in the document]
fantes: changed to pred_rnn

src/backends/caffe/caffelib.cc
fantes: changed to rnn_pred

src/backends/caffe/caffelib.cc
fantes: no we dont, we are in regression mode, and nothing that can be used confidence comes out from the LSTM layers. Confidence can be simulated with dropouts (to see possible amount of variation, and variational version of lstm, if they exist :)

src/backends/caffe/caffelib.cc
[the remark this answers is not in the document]
beniz: OK, we're using `labels` instead, see https://github.com/jolibrain/deepdetect/blob/master/tests/ut-caffeapi.cc#L1391 with `"labels":["name_col1",...]` as an array. This requires reading the CSV header.

src/backends/caffe/caffelib.cc
beniz: Right, do we need to return `probs` ?

src/generators/net_caffe_recurrent.cc
beniz: OK, for OCR we use Xavier, can we fill the value with a parameter to cover other values ?

src/backends/caffe/caffelib.cc
beniz: Agreed to have a dedicated PR to it.

src/generators/net_caffe_recurrent.cc
```diff
@@ -0,0 +1,371 @@
...
+    caffe::FillerParameter *bfparam = cparam->mutable_bias_filler();
+    bfparam->set_type("xavier");
+  }
+
+
+  void NetLayersCaffeRecurrent::configure_net(const APIData &ad_mllib)
+  {
+
+    std::vector<std::string> layers = {"L","L"}; // default
+    std::vector<double> dropouts; // default
```
beniz: I suggest we keep using the existing `dropout` parameter and have the code detect whether this is an array or not. You may need to first check for singleton with `try` and `catch`, then array.

beniz:
FYI, tests fail with crash error:

```
5: [2019-01-25 18:58:08.417] [my_service_csvts] [info] Net total flops=450 / total params=450
5: [2019-01-25 18:58:08.417] [my_service_csvts] [info] detected network type is regression
5: [2019-01-25 18:58:08.417] [caffe] [info] Opened lmdb csvts_db/test.lmdb
5/9 Test #5: ut_caffeapi ......................***Exception: SegFault1150.93 sec
```

See log.

src/backends/caffe/caffelib.cc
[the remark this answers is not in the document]
fantes: done

src/backends/caffe/caffelib.cc
fantes: in supervisedouputconnector::add_results (l.190) , results are iterated based on probs.size(), and stuff like (prob[i], result[i]) are added to _vvcats, so not putting probs needs a bit of refactoring

src/generators/net_caffe_recurrent.cc
fantes: done

src/generators/net_caffe_recurrent.cc
fantes: until which level should the parameter be exposed ? it is now in add_basic_block, but not in configure_net (having it in configure_net means exposing it at API level)

beniz:
> `unscale_outputs` allows to unscale predictions according to `min_vals` and `max_vals`
> the values are the indexes of outputs in vectors `min_vals`, `max_vals`. ie in example, output 0 (the only one) will be scaled by *(15.5-15.0) + 15.0

In don't correctly understand this I think. Also in the examples, why do we need the `min_vals` and `max_vals` to be specified again ? In most if not all cases, they'd be computed internally and returned at the end of training, no ?

fantes:
> > `unscale_outputs` allows to unscale predictions according to `min_vals` and `max_vals`
> > the values are the indexes of outputs in vectors `min_vals`, `max_vals`. ie in example, output 0 (the only one) will be scaled by *(15.5-15.0) + 15.0
> 
> In don't correctly understand this I think. Also in the examples, why do we need the `min_vals` and `max_vals` to be specified again ? In most if not all cases, they'd be computed internally and returned at the end of training, no ?


Internally, inputs and outputs are scaled between 0 and 1 so the model can ouptut only values between 0 and 1. ; `unscale_outputs` allows dede to give original (unscaled) values by doing output' = output * (max-min)+min, but it needs min and max at predict time for doing so.  
If min_vals and max_vals are not given, they are computed internally, but only on current request data. This might be ok for training (as one is supposed to give data that covers extreme cases), but definitely not for predict, as it is likely to be asked for one particular sequence.

beniz:
Hello, I'd need the list of all new API parameters and their definition, e.g. I am missing `offset` and possibly others. Thanks.

fantes:
> Hello, I'd need the list of all new API parameters and their definition, e.g. I am missing `offset` and possibly others. Thanks.

updated to add offset. Also updated scale and min/max_vals for autounscale

beniz:
> set `batch_size` to 1 (a new series is considered the following of a previous one),

`test_batch_size` ?

beniz:
Instead of 
```
"layers":["L","L","L"],
"hidden": 10
```
Let's do like we do for convolutional layers:

```
"layers":["L10","L10","L10"]
```
unless it does not make sense to have stacked LSTMs with different sizes.

Another advantage is that it removes an API parameter.

beniz:
> * `timesteps` is the length of series to give to caffe (no overlap!!)

Let's throw a `BadParam` error if undefined while using the `cvsts` input connector.

fantes:
> > set `batch_size` to 1 (a new series is considered the following of a previous one),
> 
> `test_batch_size` ?

it seems to me that at predict time, batch_size is used and not test_batch_size, and this comment was for predict. But I may be wrong :)

fantes:
> Instead of
> 
> ```
> "layers":["L","L","L"],
> "hidden": 10
> ```
> 
> Let's do like we do for convolutional layers:
> 
> ```
> "layers":["L10","L10","L10"]
> ```
> 
> unless it does not make sense to have stacked LSTMs with different sizes.
> 
> Another advantage is that it removes an API parameter.

done

beniz:
> * `init` let user choose the initial filler for weights. supproted values are from caffe values : `uniform` (bounds auto-computed as in pytorch); `xavier`; `msra`

I believe layer initialization from API needs another PR, can we force `uniform` for LSTMs for now, and remove the `init` exposition to API ? Thanks.

PR #518, opened 2018-11-12
Add support for NCNN backend

CorentinB:
This pull request add support for [NCNN](https://github.com/Tencent/ncnn) as a backend for DD.

Compilation of NCNN can be performed with `-DUSE_NCNN=ON` and [NCNN's RPi3 support](https://github.com/Tencent/ncnn/pull/640) is triggered with `-DRPI3=ON`.

Caffe models can be converted to NCNN models with [caffe2ncnn](https://github.com/Tencent/ncnn/tree/master/tools/caffe), and should produce a `.bin` weights file, and a `.param` for the params file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ DeepDetect relies on external machine learning libraries through a very generic 
 - the deep learning libraries [Caffe](https://github.com/BVLC/caffe), [Tensorflow](https://tensorflow.org), [Caffe2](https://caffe2.ai/) and [Dlib](http://dlib.net/ml.html)
 - distributed gradient boosting library [XGBoost](https://github.com/dmlc/xgboost)
 - clustering with [T-SNE](https://github.com/DmitryUlyanov/Multicore-TSNE)
-- similarity search with [Annoy]()
+- similarity search with [Annoy](https://github.com/spotify/annoy/)
 
 #### Machine Learning functionalities per library (current):
 


src/backends/ncnn/ncnninputconns.h
```diff
@@ -0,0 +1,89 @@
...
+            }
+            catch(const std::exception& e)
+            {
+                throw;
+            }
+            cv::Mat bgr = this->_images.at(0);
+            //_in = ncnn::Mat::from_pixels_resize(bgr.data, ncnn::Mat::PIXEL_BGR, bgr.cols, bgr.rows, width(), height());
+
+            _in = ncnn::Mat::from_pixels(bgr.data, ncnn::Mat::PIXEL_BGR, bgr.cols, bgr.rows);
+            //const float mean_vals[3] = {104.f, 117.f, 123.f};
```
beniz: Please remove this commented line.

src/backends/ncnn/ncnnlib.cc
```diff
@@ -0,0 +1,232 @@
...
+			  std::istream_iterator<std::string>(),
+			  std::string("processor"));
+    }
+  
+    unsigned int hardware_concurrency()
+    {
+        unsigned int cores = std::thread::hardware_concurrency();
+        if (!cores)
+            cores = my_hardware_concurrency();
+        if (cores > 4)
```
beniz: This is hacky, let's remove it. I suggest we add a `threads` parameter to the `mllib` object at service creation. If unset, the `hardware_concurrency` value applies, otherwise the specified `threads` value applies.

src/backends/ncnn/ncnnlib.cc
```diff
@@ -0,0 +1,232 @@
...
+      _net->load_model(this->_mlmodel._weights.c_str());
+      if (ad.has("nclasses"))
+	_nclasses = ad.get("nclasses").get<int>();
+      /*_blob_pool_allocator.clear();
+	_workspace_pool_allocator.clear();*/
+      _blob_pool_allocator.set_size_compare_ratio(0.0f);
+      _workspace_pool_allocator.set_size_compare_ratio(0.5f);
+      ncnn::Option opt;
+      opt.lightmode = true;
+      opt.num_threads = hardware_concurrency();
```
beniz: Let's apply the `threads` value here if it exists.

beniz:
Thanks, great addition!
See my comments, then let's consider squashing the many commits once the new `threads` flag is added.

beniz:
Also, you may want to add `-DUSE_NCNN=ON` to the `Jenkinsfile` (https://github.com/jolibrain/deepdetect/blob/master/Jenkinsfile) so that the CI takes it into account.

src/backends/ncnn/ncnninputconns.h
```diff
@@ -0,0 +1,87 @@
...
+            {
+                ImgInputFileConn::transform(ad);
+            }
+            catch(const std::exception& e)
+            {
+                throw;
+            }
+            cv::Mat bgr = this->_images.at(0);
+
+            _in = ncnn::Mat::from_pixels(bgr.data, ncnn::Mat::PIXEL_BGR, bgr.cols, bgr.rows);
```
chichaj: The flag may need to be ```PIXEL_GRAY``` depending on the API request
This is can be checked using ```this->_bw```
(```bool _bw; /**< whether to convert to black & white. */```)

src/backends/ncnn/ncnnlib.cc
```diff
@@ -0,0 +1,234 @@
...
+        std::vector<APIData> vrad;
+        std::vector<double> probs;
+        std::vector<std::string> cats;
+        std::vector<APIData> bboxes;
+        APIData rad;
+	
+        // Get confidence_threshold
+        double confidence_threshold = 0.0;
+        if (ad_output.has("confidence_threshold")) {
+	  try {
```
chichaj: This try/catch can be replace with
apitools::get_float(ad_output, "confidence_threshold", confidence_threshold);

chichaj:
It seems that the openmp parallelization isn't working. From what I found it is caused by the ```-fopenmp``` flag not being forwarded correctly at compilation time.

A simple fix is to add ```set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")``` in NCNN's CMakelists.txt.

Here is a more explicit patch:

```
From 314da24c254cf1185cb3fafeac0c7d601022f2d6 Mon Sep 17 00:00:00 2001
From: Julien CHICHA <julien.chicha@epitech.eu>
Date: Wed, 21 Nov 2018 15:49:42 +0100
Subject: [PATCH 1/1] -fopenmp

---
 CMakeLists.txt | 1 +
 1 file changed, 1 insertion(+)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 7ca537b..ac94a4b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ option(NCNN_CMAKE_VERBOSE "print verbose cmake messages" OFF)
 
 if(NCNN_OPENMP)
     find_package(OpenMP)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
     # For CMake < 3.9, we need to make the target ourselves
     if(NOT TARGET OpenMP::OpenMP_CXX)
         find_package(Threads REQUIRED)
-- 
2.19.1
```

beniz:
@CorentinB please see the patch above, to be added to the PR. Thanks!

CorentinB:
Because it's related to NCNN's CMakeLists.txt, maybe it would be better to open a PR on NCNN?

It would be cleaner than applying a patch during DeepDetect build.

CorentinB:
Just added the patch to this PR, but while looking for making the PR on NCNN side, I just discovered that they patched this issue about openmp [in this commit](https://github.com/Tencent/ncnn/commit/a2611f47206ef0fe74f140954bd4c0534fb47a50) 2 weeks ago. We should fix DeepDetect build to a more recent NCNN commit.

beniz:
Need to rebase on master due to PR having been merged earlier. Thanks!

Jenkinsfile
```diff
@@ -6,7 +6,7 @@ pipeline {
         sh '''
 mkdir -p build
 cd build
-cmake .. -DBUILD_TESTS=ON -DUSE_CUDNN=ON -DUSE_SIMSEARCH=ON -DUSE_TSNE=ON -DUSE_XGBOOST=ON -DUSE_CAFFE2=ON -DCUDA_ARCH="-gencode arch=compute_61,code=sm_61"
+cmake .. -DBUILD_TESTS=ON -DUSE_CUDNN=ON -DUSE_SIMSEARCH=ON -DUSE_TSNE=ON -DUSE_XGBOOST=ON -DUSE_CAFFE2=ON -DUSE_NCNN -DCUDA_ARCH="-gencode arch=compute_61,code=sm_61"
```
beniz: It should be `-DUSE_NCNN=ON`. Please fix it, and see whether tests are passing.

CorentinB: Yes, fixed.

CorentinB:
@beniz it looks like Jenkins doesn't execute NCNN's tests, actually when I look at the log file, the command executed for the tests doesn't include -DUSE_NCNN=ON for some reasons..

beniz:
You need to merge master again for this PR to pass the tests.

CorentinB:
New PR: https://github.com/jolibrain/deepdetect/pull/527

PR #519, opened 2018-11-15
Fix ut conn

fantes:


beniz:
Thanks, fix due to website update.

PR #520, opened 2018-11-15
Making Caffe optional in builds

beniz:
This allows building DD without building the Caffe backend. Useful for embedded devices, especially wrt to forthcoming #518


src/services.h
```diff
@@ -55,33 +58,51 @@
...
     MLService<CaffeLib,CSVCaffeInputFileConn,SupervisedOutput,CaffeModel>,
     MLService<CaffeLib,TxtCaffeInputFileConn,SupervisedOutput,CaffeModel>,
     MLService<CaffeLib,SVMCaffeInputFileConn,SupervisedOutput,CaffeModel>,
     MLService<CaffeLib,ImgCaffeInputFileConn,UnsupervisedOutput,CaffeModel>,
     MLService<CaffeLib,CSVCaffeInputFileConn,UnsupervisedOutput,CaffeModel>,
     MLService<CaffeLib,TxtCaffeInputFileConn,UnsupervisedOutput,CaffeModel>,
+    MLService<CaffeLib,SVMCaffeInputFileConn,UnsupervisedOutput,CaffeModel>
+#endif
 #ifdef USE_CAFFE2
+    #ifdef USE_CAFFE
```
chichaj:
```
#ifndef PREV_SERVICE
    #define PREV_SERVICE
#else
    ,
#endif
```
With this kind of directive, the MLServices can be position-independent in the code

src/CMakeLists.txt
```diff
@@ -38,4 +41,6 @@ if (USE_DLIB)
   list(APPEND ddetect_SOURCES backends/dlib/DNNStructures.h backends/dlib/dliblib.cc backends/dlib/dliblib.h backends/dlib/dlibmodel.cc backends/dlib/dlibmodel.h backends/dlib/dlibinputconns.h)
 endif()
 add_library(ddetect ${ddetect_SOURCES})
-add_dependencies(ddetect caffe_dd)
+if (USE_CAFFE)
+  add_dependencies(ddetect caffe_dd)
+endif()
```
chichaj: An error happen when compiling with CAFFE2 and not CAFFE:
pytorch is set as a dependencie of caffe_dd (to ensure they share the same protobuf files)
But if CAFFE is not built, pytorch still need to be compiled.

Adding a
```add_dependencies(ddetect pytorch)```
should solve this

File: README.md [hunk 3]
```diff
@@ -346,6 +346,13 @@ Specify the option via cmake:
 cmake .. -DUSE_CAFFE2=ON
 ```
 
+#### Build without Caffe
+
+Caffe remains the default backend for DeepDetect though it can be deactivated with cmake. However, at least one library needs to be specified:
+```
+cmake .. -DUSE_CAFFE=NO -DUSE_XGBOOST=ON
```
CorentinB: Maybe it should be `-DUSE_CAFFE=OFF` instead of NO, to match other flags, and also avoid typing NO instead of ON or ON instead of NO.

beniz: Thanks, this is fixed!

PR #521, opened 2018-11-16
Automated model installation from compressed archive at service creation

beniz:
This PR allows the automatic loading and installation of a model as a compressed archive.

If the model directory (and thus the tarball) contains a `config.json` file with the exact JSON content of a `PUT /services` call, the model parameters are loaded automatically. 

Thus in the example below the number of classes `nclasses` and `width`/`height` parameters do not need to be specified anymore.

Note: **when the `config.json` file exists, its `parameters` object overrides the full `parameters` object specified via the API.**

* API addition
`init` parameter in `model` on `PUT /services`

Example:

- service creation:

```
       "mllib":"caffe", 
       "description":"image detection service", 
       "type":"supervised", 
       "parameters":{ 
         "input":{ 
           "connector":"image" 
         }, 
         "mllib":{ 
         } 
       }, 
       "model":{ 
         "templates":"../templates/caffe/", 
         "repository":"/home/beniz/tmp/test_auto",
         "init":"/path/to/detection_600.tar.gz" 
       } 
     }' 
```

- service predict:

```
 "service": "imageserv",
 "parameters": {
  "input": {},
  "output": {
   "confidence_threshold": 0.2,
   "bbox": true
  },
  "mllib": {
   "gpu": true
  }
 },
 "data": [
  "https://s.hdnux.com/photos/76/76/11/16507462/9/727x0.jpg"
 ]
}'
```

### DIFF
--- a/src/backends/caffe2/caffe2model.cc
+++ b/src/backends/caffe2/caffe2model.cc
@@ -25,8 +25,9 @@
 
 namespace dd {
 
-  Caffe2Model::Caffe2Model(const APIData &ad)
-    :MLModel()
+  Caffe2Model::Caffe2Model(const APIData &ad, APIData &adg,
+			   const std::shared_ptr<spdlog::logger> &logger)
+    :MLModel(ad, adg, logger)
   {
     std::map<std::string, std::string *> names =
       {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -519,7 +519,7 @@ set(COMMON_LINK_DIRS
   ${DLIB_LIB_DIR})
 if (USE_HDF5)
   set(COMMON_LINK_LIBS
-    ddetect ${CUDA_LIB_DEPS} glog gflags ${OpenCV_LIBS} curlpp curl hdf5_cpp ${Boost_LIBRARIES}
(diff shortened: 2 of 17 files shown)


src/backends/caffe2/caffe2model.cc
```diff
@@ -25,8 +25,9 @@
 
 namespace dd {
 
-  Caffe2Model::Caffe2Model(const APIData &ad)
-    :MLModel()
+  Caffe2Model::Caffe2Model(const APIData &ad, APIData &adg,
+			   const std::shared_ptr<spdlog::logger> &logger)
+    :MLModel(ad, adg, logger)
```
chichaj: Now that MLModels are created with a logger, could a reference be keeped so
```read_from_repository(const std::shared_ptr<spdlog::logger> &logger)```
and
```update_from_repository(const std::shared_ptr<spdlog::logger> &logger)```
can use it ?

beniz: Good point! will look at it.

chichaj: Indeed, the context are different, and so should be the loggers.
But maybe we should at least change the constructors, so the call
```read_from_repository(ad.get("repository").get<std::string>(), spdlog::get("api"))```
becomes
```read_from_repository(ad.get("repository").get<std::string>(), logger)```

PR #523, opened 2018-11-23
make template='' behave like no template

fantes:
previously template="" was acting trying to instantiate template with empty name. Now template="" does not try to instantiate template.


beniz:
Make template great again :) thanks!

PR #539, opened 2019-01-30
Dice loss parameters + contours : clean API 

fantes:
example curl at service creation:
```
curl -X PUT "http://10.10.77.61:"$1"/services/seg" -d '{                                                                                                                                                           
  "mllib":"caffe",                                                                                                                                                                                                 
  "description":"seg dice",                                                                                                                                                                                        
  "type":"supervised",                                                                                                                                                                                             
  "parameters":{                                                                                                                                                                                                   
       "input":{                                                                                                                                                                                                   
          "connector":"image",                                                                                                                                                                                     
          "db": false,                                                                                                                                                                                                 
          "height": 480,                                                                                                                                                                                               
          "width": 480,                                                                                                                                                                                                
           "segmentation": true,                                                                                                                                                                                       
           "bw":false                                                                                                                                                                                                  
          },                                                                                                                                                                                                          
       "mllib":{                                                                                                                                                                                                   
           "template": "pspnet_vgg16",  
           ...
          "loss": "dice",                                                                                                                                                                                              
          "dice_param": {                                                                                                                                                                                              
              "class_weighting": {                                                                                                                                                                                     
                  "compute_on":"all",                                                                                                                                                                                  
                  "weight" : "equalize_classes"                                                                                                                                                                      
               },                                                                                                                                                                                                       
             "contour" : {                                                                                                                                                                                            
                 "shape" : "simple",                                                                                                                                                                                   
                 "size": 3,                                                                                                                                                                                            
                 "amplitude": 20.5                                                                                                                                                                                     
              }                                                                                                                                                                                                        
           }                                                                                                                                                                                                             
        },                                                                                                                                                                                                         
       "output": {                                                                                                                                                                                                 
      "store_config": true                                                                                                                                                                                         
    }                                                                                                                                                                                                              
  },                                                                                                                                                                                                               
  "model":{                                                                                                                                                                                                        
      "repository":"/data1/infantes/dice_debug", "templates":"/home/infantes/deepdetect/templates/caffe"                                                                                                           
  }                                                                                                                                                                                                                
 }'              
```

- ` "loss" = "dice" `
- `dice_param` is a container for all parameters. If omitted, neither class nor contour weighting is applied.
- `dice_param.class_weighting` controls generalized (weighted multiclass) dice. Omit this param for multiclass non weighted dice (auto selected based on nclasses)
-  `"compute_on"` can be 
       - `"all"` [default] to compute class weights on all data,  
       - `"batch"` to compute class weights on batch only, 
       - `"image"` to compute class weight on current image only.
- `"weight"` can be:
        -  `"equalize_classes"`:  do not take class frequency into account, should be the same as non weighted,  
        - `"inverse_volume"`: [default] correct weight so that more frequent classes contributes less, 
        - `"extra_small_volumes"` put even more focus on rare classes. 
- `dice_param.contour` controls extra weights given to contours of patches. Omit for no contour weighting. 
- `"shape"` : select how weights are propagated around contours
         -  `"simple"` gives correct results for smooth shapes, but gives small weights to majoritary region in case of sharp shapes; 
         - `"sharp"`  [default] tries to correct previous behavior in case of sharp regions (both side of sharp contour are given high weights, even majoritary region).                                                                                          
- `"size"`: [default: `7`] is the size of convolution used to compute weights, it is roughly the half of the width of the weighted region. As an example: if size is 3, around 3 pixels will be weighted at both sides of the frontier. Weight are very quickly  decreasing with distance to the frontier, so 3 is a bit small.                                                                                                                                                                                          
- `"amplitude"`: [default: `5.0`]  is the relative amplitude of the weight. If 10 is given;  weighting will give at most ten times more weight to some pixels compared to others.  This max is reached in very rare case, ie. 1 isolated pixel.

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -208,25 +208,22 @@ namespace dd
 	caffe::ReadProtoFromTextFile(dest_net,&net_param); //TODO: catch parsing error (returns bool true on success)
 	caffe::ReadProtoFromTextFile(dest_deploy_net,&deploy_net_param);
 
-       if (this->_loss.compare(0, 4, "dice",0,4) == 0)
+       if (this->_loss == "dice")
       // dice loss!!
       {
-        int ignore_label = -1;
-        if (ad.has("ignore_label"))
-          ignore_label = ad.get("ignore_label").get<int>();
+        if (_nclasses > 2)
+          update_protofiles_one_hot(net_param);
 
         if (net_param.name().compare("deeplab_vgg16")==0
 	    || net_param.name().compare("pspnet_vgg16")==0
 	    || net_param.name().compare("pspnet_50")==0
 	    || net_param.name().compare("pspnet_101")==0)
           {
-            update_protofiles_dice_one_hot(net_param, this->_loss, _nclasses);
-            update_protofiles_dice_deeplab_vgg16(net_param, deploy_net_param, this->_loss, ignore_label);
+            update_protofiles_dice_deeplab_vgg16(net_param, deploy_net_param, ad);
 	  }
         else if (net_param.name().compare("unet") == 0)
           {
-            update_protofiles_dice_one_hot(net_param, this->_loss, _nclasses);
-            update_protofiles_dice_unet(net_param,deploy_net_param, this->_loss, ignore_label);
+            update_protofiles_dice_unet(net_param,deploy_net_param, ad);
 	  }
 
       }
@@ -900,7 +897,7 @@ namespace dd
     if (ad.has("loss"))
       {
         _loss = ad.get("loss").get<std::string>();
-        if (this->_loss.compare(0, 4, "dice",0,4) == 0)
+        if (this->_loss == "dice")
           if (! this->_inputc._segmentation)
             throw MLLibBadParamException("asked for  dice loss without segmentation");
       }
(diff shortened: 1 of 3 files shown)


src/backends/caffe/caffelib.cc
```diff
@@ -208,25 +208,22 @@ namespace dd
 	caffe::ReadProtoFromTextFile(dest_net,&net_param); //TODO: catch parsing error (returns bool true on success)
 	caffe::ReadProtoFromTextFile(dest_deploy_net,&deploy_net_param);
 
-       if (this->_loss.compare(0, 4, "dice",0,4) == 0)
+       if (this->_loss == "dice")
       // dice loss!!
       {
-        int ignore_label = -1;
-        if (ad.has("ignore_label"))
```
beniz: I don't believe we should remove `ignore_label`. In practice it is very useful with at least the softmax and weighted softmax losses, to ignore some classes. One drawback is that we can ignore one class only in the current implementation, which is due to Caffe initial implementation.

fantes: In case of dice loss, I have not been able so far to find a satisfactory way to ignore a class despite having tested several ways to do so. I will put that in my todo list

beniz: Ok but it should remain available for the softmax loss, right ?

fantes: I think it does , we are inside the if (this->_loss == "dice")   block   if I do not fool myself

PR #542, opened 2019-02-11
fix : do not output best confidence map if not asked for 

fantes:
correctlty set default of boolean conf_best so that it best confidence map is not output by default


beniz:
@fantes can you double check that activation is correct (it was always on beforehand) ? We don't see it on the platform at the moment. Thanks.

PR #554, opened 2019-03-14
ncnn timeseries for predictions 

fantes:
- input + output of timeseries for ncnnlib 
- add templates instanciations for ncnn timeseries
- correclty format input and output for timeseries
- add continuation option for ncnn timeseries, so that it can use saved internal state of lstm

- example for service instanciation : 
```
curl -X PUT "http://10.10.77.61:8080/services/dd_valve" -d '{
{"mllib":"ncnn", "description":"dd valve timeserie simulator",
    "type":"supervised",
    "parameters":{
         "input":{
              "connector":"csvts", "db": false,
              "label" : ["PRV.port_b.p","PRV.der_alpha.u","PRV.der_alpha.y"],
              "ignore": ["time"] },
         "mllib":{ "regression":true  }  },
    "model":{ "repository":"/home/infantes/ncnn_ts/model/"  }}'
```

- example for prediction: 
```
curl -X POST "http://10.10.77.61:8080/predict" -d '{
{"service":"dd_valve",
      "parameters":{
         "input":{
              "connector":"csvts", "db": false,
                "timesteps":1,
                "separator":";",
                "scale":true,
                "continuation":true},         "output": {},
          "mllib":{"net":{"test_batch_size":1}}},
          "data":["time;PRV.monitoring.inlet.m_flow;PRV.monitoring.inlet.p;PRV.monitoring.inlet.T;domain.SAP;PRV.SENSLINE.p;PRV.port_b.p;PRV.der_alpha.u;PRV.der_alpha.y","0.0;9.317033900850000e-01;4.599366323670000e+05;4.731500000000000e+02;1.013250000000000e+05;2.388701844060000e+05;0.0;0.0;0.0"]}}'
```

### DIFF
--- a/src/backends/ncnn/ncnnlib.cc
+++ b/src/backends/ncnn/ncnnlib.cc
@@ -55,7 +55,8 @@ namespace dd
 	_net = tl._net;
 	tl._net = nullptr;
 	_nclasses = tl._nclasses;
-    _threads = tl._threads;
+       _threads = tl._threads;
+       _timeserie = tl._timeserie;
     }
 
     template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
@@ -79,6 +80,11 @@ namespace dd
         else
             _threads = dd_utils::my_hardware_concurrency();
 
+        if (typeid(this->_inputc) == typeid(CSVTSNCNNInputFileConn))
+          {
+            _timeserie = true;
+          }
+
         _blob_pool_allocator.set_size_compare_ratio(0.0f);
         _workspace_pool_allocator.set_size_compare_ratio(0.5f);
         ncnn::Option opt;
@@ -87,7 +93,7 @@ namespace dd
         opt.blob_allocator = &_blob_pool_allocator;
         opt.workspace_allocator = &_workspace_pool_allocator;
         ncnn::set_default_option(opt);
-	model_type(this->_mlmodel._params,this->_mltype);
+        model_type(this->_mlmodel._params,this->_mltype);
     }
 
     template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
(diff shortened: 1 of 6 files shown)


src/backends/ncnn/ncnnlib.cc
```diff
@@ -116,8 +116,20 @@ namespace dd
         } catch (...) {
             throw;
         }
+        if (_old_height != -1 && _old_height != inputc.height())
+          {
+            // in case of new prediction with different timesteps value, clear old net
```
beniz: Is there any hope of not re-creating the net ? Do we have an potential example of time-series in prediction with variable time-step ?

src/backends/ncnn/ncnnlib.cc
```diff
@@ -140,14 +152,21 @@ namespace dd
 		  blank_label = ad_output.get("blank_label").get<int>();
 	      }
 	  }
-	
+       bool timeserie = false;
+       if (typeid(inputc) == typeid(CSVTSNCNNInputFileConn))
```
beniz: I believe this should be set once and for all, and `timeserie` be a class variable instead since the input connector cannot be changed on a living service.

src/csvinputfileconn.cc
```diff
@@ -83,6 +83,7 @@ namespace dd
 	else _cifc->add_train_csvline(std::to_string(_cifc->_csvdata.size()+1),vals);
 	++l;
       }
+    _cifc->update_columns();
```
beniz: Is this a bug fix ?

src/backends/ncnn/ncnnlib.cc
fantes: the net is re created only in case of change of height / timesteps. Generally speaking it will be created only once: is is created only in here, not at init_mllib. Here we just check in case height / timestep has changed (which can happen especially during debug :) .

src/backends/ncnn/ncnnlib.cc
fantes: putting timeserie as a mltype is on my task list

src/csvinputfileconn.cc
fantes: yes, columns were not updated in case of csv data passed directly in json

PR #567, opened 2019-04-16
skip bad bbox on Jetson nano GPU with Caffe

beniz:
CUDA execution of DetectionOutput Caffe layer appears to sometimes return overflows in bounding boxes output coordinates.
~~~At this stage, this is unclear why this is never happening on other GPUs, nor has it been reported by Jetson Nano customers that we know use DD, though we checked with them.~~~
This PR makes sure these bounding boxes are removed by DD.

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -2411,6 +2411,18 @@ namespace dd
 			probs.push_back(detection[2]);
 			cats.push_back(this->_mlmodel.get_hcorresp(detection[1]));
 			APIData ad_bbox;
+			bool skip = false;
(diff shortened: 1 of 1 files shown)


YaYaB:
Hey!
Thanks for the PR. I have the same issue on a 1080Ti and on CPU.
However I would avoid removing the bounding box that do not fit in the image but just regularize the bounding boxes with the size of the image

YaYaB:
Here is the modification to keep those "bad bboxes" while regularizing them #594

beniz:
A workaround for the Caffe / Nano issue on detection models is to use the `tensorrt` backend instead, that automatically converts the Caffe models and improves performances.

PR #570, opened 2019-04-30
move init of attributes to construction instead of declaration

fantes:
for some reason, visual studio does not seem to init properly aatributes when their value is given at declaration and not at construction

### DIFF
--- a/src/backends/ncnn/ncnnlib.cc
+++ b/src/backends/ncnn/ncnnlib.cc
@@ -42,7 +42,7 @@ namespace dd
   
     template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
     NCNNLib<TInputConnectorStrategy,TOutputConnectorStrategy,TMLModel>::NCNNLib(const NCNNModel &cmodel)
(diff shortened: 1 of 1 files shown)


beniz:
I believe this is redundant, values are already initialized here: https://github.com/jolibrain/deepdetect/blob/master/src/backends/ncnn/ncnnlib.h#L62

r2d3:
I have to recheck why I got _thread==0 on Windows...
is it a problem with the ctor, the copy/move ctor, ...

r2d3:
I confirm that the problem does not come from init but from 
https://github.com/jolibrain/deepdetect/blob/master/src/backends/ncnn/ncnnlib.cc#L85

`dd_utils::my_hardware_concurrency()` returns 0 on Windows because it uses `/proc/cpuinfo` to count CPU cores

https://github.com/jolibrain/deepdetect/blob/master/src/utils/utils.hpp#L53

Forcing threads:1 in the init string is the correct solution for Windows as there is a bug in NCNN fork and OpenMP

fantes:
nice work ! I am closing this PR . Thanks a lot

PR #559, opened 2019-03-26
Selection of API at build time

fantes:
template instantiation respects compile flags
source file list also respect compile flags 

example use 
- in order to use http_server
cmake .. -DUSE_HTTP_SERVER  (default ON)

- in order to build command line json api
cmake .. -DUSE_COMMAND_LINE  (default ON)

- in order to build neither one nor the other but only the internal json api in order to link against
cmake .. -DUSE_JSON_API  (default ON as it is needed by the two others)

PR #572, opened 2019-04-30
use USE_HTTP to condionally include httpjsonapi.h, correct not defined to !defined

r2d3:
use USE_HTTP to condionally include httpjsonapi.h
correct not defined to !defined to get correct behavior on Windows

### DIFF
--- a/main/dede.cc
+++ b/main/dede.cc
@@ -22,7 +22,9 @@
 #include "deepdetect.h"
 #include "commandlineapi.h"
 #include "commandlinejsonapi.h"
+#ifdef USE_HTTP
 #include "httpjsonapi.h"
+#endif
 #include "imginputfileconn.h"
 #include "outputconnectorstrategy.h"
 #ifdef USE_XGBOOST
--- a/src/deepdetect.cc
+++ b/src/deepdetect.cc
@@ -22,7 +22,9 @@
 #include "deepdetect.h"
 #include "commandlineapi.h"
 #include "commandlinejsonapi.h"
+#ifdef USE_HTTP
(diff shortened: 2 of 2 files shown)


beniz:
@fantes @r2d3  can you check that this works on Linux builds ? Typically, some of the connectors rely on http retrieval functions, and I believe typically that https://github.com/jolibrain/deepdetect/blob/master/src/inputconnectorstrategy.h#L56 should be deactivated at compile time when USE_HTTP is not defined.

r2d3:
The current semantic of `USE_HTTP` as defined in #559 is only for the server part.

So I don't think that httpclient::get_call should be impacted. You should be able to do http request while using only commandline or JSON API (i.e. when USE_HTTP is undefined/0/off)

My understanding is that httpjsonapi is only to serve the API via a http server.

fantes:
Hi

I agree with david on this. Indeed the in https://github.com/jolibrain/deepdetect/pull/559  the _cmake_ option USE_HTTP_SERVER activates  _compile_ option -DUSE_HTTP ,
so  USE_HTTP is only related to server API, not to httpclient

beniz:
> i can push USE_HTTP -> USE_HTTP_SERVER for compile option, if you think that would make things clearer in the future

OK yes let's have this, it's a good idea.

main/dede.cc
```diff
@@ -22,7 +22,9 @@
 #include "deepdetect.h"
 #include "commandlineapi.h"
```
beniz: According to #559 these should be controlled with #ifdef as well, no ? cc @fantes

PR #573, opened 2019-04-30
Do not use httpclient class on Windows

r2d3:


beniz:
Do we want empty functions, or do we rather want that the whole class is not used throughout the code when building for Windows ? My preference goes to the later, since empty functions may lead to undefined behavior, whereas deactivating httpclient throughout the code removes the underlying functionalities. This is related to #572 and #559 I believe. CC @fantes

r2d3:
#572 abd #559 are more related on the http server part.

Here httpclient is only used in 2 places :
https://github.com/jolibrain/deepdetect/blob/master/src/inputconnectorstrategy.h#L56
https://github.com/jolibrain/deepdetect/blob/master/src/mlmodel.h#L198

So, if you want to disable the http client part, you should add `USE_HTTP_CLIENT` option in CMakeLists and rename the current `#def` to `USE_HTTP_SERVER` and `USE_HTTP_CLIENT`.

In terms of testing, you will increase the combinations of options to test.

As httpclient should normally be there, I propose to not use it on windows and make `mlmodel.h`/`inputconnectorstrategy.h` returns an error/assert on Windows when trying to retrieve a URL based ressource.

r2d3:
So I implemented the "httpclient disable" another way. I could squash the 2 commits if this new version is OK to you.

But in no case, httpclient has a link with USE_HTTP_SERVER

fantes:
Hi 
i prefer the new patch, I am okay with merging it, because indeed we are in a case of 'not implemented' 
see https://github.com/jolibrain/deepdetect/pull/572#issuecomment-488586330 for clarification of semantics of USE_HTTP (spoiler alert: david is right, but maybe i can change option name in order to avoid future confusion)

PR #578, opened 2019-05-14
Tensor RT 5.0 backend

fantes:
TensorRT backend for prediction. Caffe models are automatically converted and optimized on the first `POST /predict` call. The converted weights are stored on disk, and can be kept  and reused even after the service has been stopped.

- tested on jetson TX2, AGX, nano / jetpack / tensor RT 5.0

- tested for detection (on squeezenet-ssd) and image classification (on vgg16, googlenet and resnet_18)
- API for service creation:
```
curl -X PUT http://localhost:8080/services/squeezenet_ssd_voc -d '{
 "description": "Squeezenet SSD",
 "model": {
  "repository": "/path/to/model_ssd"
 },
 "mllib": "tensorrt",
 "type": "supervised",
 "parameters": {
  "input": {
   "connector": "image",
   "width":300,
   "height":300
  },
  "mllib" : {
     "datatype": "fp32",
     "maxBatchSize": 4,
     "maxWorkspaceSize": 1024
  }
 }
}'
```
- API for prediction: 
```
  "service": "squeezenet_ssd_voc",
  "parameters": {
    "input": {
    "height": 300,
    "width": 300
   }, 
    "output": {
      "confidence_threshold": 0.1,
      "bbox": true
    },
    "mllib": {
      "gpu": true
    }
  },
  "data": [
    "/home/nvidia/detect_600.jpg"
  ]
}'
```

- do not set `bbox: true` to do classification

## API options:
- `mllib.maxBatchSize`: tensorRT precompiles net using a maximum batch size, this option allows to reduce the batch size to recompile de model. It a precompiled engine is present, this previous engine is used, overriding this option
- `mllib.maxWorkspaceSize`: [default: 1024] in MB: tensorrt maximul allowed memory for precompilation of models. Usefull mainly on nano : 1024 is GB and may cause dede to be sigkilled if not enough memory. 256 allows to limit memory consumption and create low batchsize nets
- `mllib.dla` [default: -1] : id of dla coprocessor to use [-1 means no dla]
- `mllib.datatype` [default: `fp32`] : options are: `int8`, `fp16`: allows tensortRT to try optimization that allow datasize reduction inside the network. `int8` may require an unimplemented   EntropyCalibrator and thus is strongly discouraged as of 2019/05/14
- `mllib.tensorRTEngineFile` :[default: `TRTengine_bs`] name of precompiled tensorRT engine to read or write
- `mllib.readEngine` [default: true] : read serialized tensorRT engine if available
- `mllib.writeEngine` [default: true]: write serialized tensorRT engine to disk  (if newly built)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ option(USE_TF "use TF backend")
 option(USE_NCNN "use NCNN backend")
 option(USE_HDF5 "use HDF5" ON)
 option(USE_CAFFE "use Caffe backend" ON)
+option(USE_TENSORRT "use TensorRT backend" OFF)
 
 if (USE_CAFFE)
   add_definitions(-DUSE_CAFFE)
@@ -133,7 +134,14 @@ if (USE_CAFFE)
 else()
   set(USE_SIMSEARCH OFF)
 endif()
-  
+
+if (USE_TENSORRT)
+  add_definitions(-DUSE_TENSORRT)
+  message(STATUS "TensorRT backend requires CUDA + CUDNN, enabling GPU, CUDA and CUDNN")
+  set(USE_CPU_ONLY OFF)
+  set(USE_CUDNN ON)
+endif()
+
 # optional packages
 if (USE_CPU_ONLY)
   set(USE_CAFFE_CPU_ONLY ON)
@@ -160,6 +168,9 @@ if (CUDA_FOUND)
 else()
   set(CUDA_LIB_DEPS "")
   add_definitions(-DCPU_ONLY)
+  if (USE_TENSORRT)
+    message(FATAL_ERROR "Could not find CUDA, unable to build with TensorRT backend, aborting")
+  endif()
 endif()
 
 if (USE_DLIB)
@@ -340,6 +351,15 @@ if (USE_CAFFE2)
       ${CMAKE_BINARY_DIR}/python_path/detectron)
   endif()
 
+  if (USE_TENSORRT)
+    add_custom_command(TARGET pytorch POST_BUILD
+      COMMAND ${CMAKE_BINARY_DIR}/protobuf/bin/protoc
+      --proto_path=${CMAKE_SOURCE_DIR}/src/backends/tensorrt
+      --cpp_out=${CMAKE_BINARY_DIR}/src ${CMAKE_SOURCE_DIR}/src/backends/tensorrt/caffe.proto
+      WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+      COMMENT "generating caffe.pb.*")
+  endif()
+
 endif()
 
 if (USE_TF)
@@ -381,6 +401,14 @@ set(TF_INC_DIR ${TENSORFLOW_CC_DIR}/tensorflow/ ${TENSORFLOW_CC_DIR}/tensorflow/
 set(TF_LIB_DIR ${TENSORFLOW_CC_DIR}/tensorflow/bazel-bin/tensorflow/)
 
 include_directories("${TF_INC_DIR}")
+if (USE_TENSORRT)
+  add_custom_command(TARGET tensorflow_cc POST_BUILD
+    COMMAND ${CMAKE_BINARY_DIR}/protobuf/bin/protoc
+    --proto_path=${CMAKE_SOURCE_DIR}/src/backends/tensorrt
+    --cpp_out=${CMAKE_BINARY_DIR}/src ${CMAKE_SOURCE_DIR}/src/backends/tensorrt/caffe.proto
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    COMMENT "generating caffe.pb.*")
+endif()
 
 endif() # USE_TF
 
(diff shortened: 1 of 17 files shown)


beniz:
FTR, my working `cmake` build call for this TRT branch on Jetson Nano:

```
cmake .. -DUSE_CUDNN=ON -DJETSON=ON -DCUDA_ARCH="-gencode arch=compute_53,code=sm_53" -DCUDA_USE_STATIC_CUDA_RUNTIME=OFF -DUSE_TENSORRT=ON -DUSE_CAFFE=OFF
```

CMakeLists.txt
```diff
@@ -160,6 +168,10 @@ if (CUDA_FOUND)
 else()
   set(CUDA_LIB_DEPS "")
   add_definitions(-DCPU_ONLY)
+  # if (USE_TENSORRT)
+  #   message(STATUS "Could not find CUDA, desabling TensorRT backend")
```
beniz: I believe we may want to generate an error and stop the build if `USE_TENSORRT` is `ON` and CUDA cannot be found ?

CMakeLists.txt
```diff
@@ -620,13 +632,36 @@ if (USE_NCNN)
   include_directories(${NCNN_INC_DIR})
 endif()
 
+if (USE_TENSORRT)
+  find_package(Protobuf REQUIRED)
```
beniz: It seems we are already using Protobuf for Caffe, TF and Pytorch, sometimes by forcing it up as it is included into TF and Pytorch typically. My question would be at this stage, do we want to move the `find_package` somewhere outside the conditional and/or do we need it actually ?

CMakeLists.txt
```diff
@@ -620,13 +632,36 @@ if (USE_NCNN)
...
+  endif()
+endif()
+
 # main library, main & tests
 include_directories ("${PROJECT_SOURCE_DIR}/src")
 add_subdirectory (src)
 
 # Flags used by the subdirectories
-set(COMMON_INCLUDE_DIRS ${Boost_INCLUDE_DIRS})
+set(COMMON_INCLUDE_DIRS ${Boost_INCLUDE_DIRS} ${CUDA_INCLUDE_DIRS})
```
beniz: Maybe we want to add `CUDA_INCLUDE_DIRS` inside a `CUDA_FOUND` conditional ? I guess the assumption here is that `CUDA_INCLUDE_DIRS` is empty otherwise, but this might make funny things when CUDA is found and that a CPU_ONLY build is requested.

src/backends/tensorrt/caffe.proto
```diff
@@ -0,0 +1,2235 @@
+syntax = "proto2";
```
beniz: I understand why we need this, and that's a copy of our Caffe protofile. Can we download it with cmake instead ? Reason is that although it may not change too much anymore, downloading it would avoid maintaining this file when changes do occur.

A download link to the up to date version should be this: https://raw.githubusercontent.com/jolibrain/caffe/master/src/caffe/proto/caffe.proto

src/backends/tensorrt/protoUtils.h
```diff
@@ -0,0 +1,38 @@
...
+#define PROTO_UTILS_H
+
+
+#include <google/protobuf/io/coded_stream.h>                                                            
+#include <google/protobuf/io/zero_copy_stream_impl.h>                                                   
+#include <google/protobuf/text_format.h>                                                                
+
+namespace dd
+{
+  int fixProto(const std::string dest, const std::string source);
```
beniz: It'd be good to have a short head comment to the `fixProto` function so that we can keep track of what it does at a single glance. Other functions seem more straightforward.

src/backends/tensorrt/tensorrtlib.h
```diff
@@ -0,0 +1,150 @@
...
+	  break;
+	case nvinfer1::ILogger::Severity::kWARNING:
+	  this->_logger->warn(msg);
+	  break;
+	case nvinfer1::ILogger::Severity::kINFO:
+	default: 
+	  this->_logger->info(msg);
+	  break;
+	}
+
```
beniz: empty space :)

src/ext/rapidjson/document.h
```diff
@@ -36,8 +36,10 @@ RAPIDJSON_DIAG_OFF(4127) // conditional expression is constant
 #elif defined(__GNUC__)
 RAPIDJSON_DIAG_PUSH
 RAPIDJSON_DIAG_OFF(effc++)
+#if __GNUC__ >=6
+  RAPIDJSON_DIAG_OFF(terminate)
```
beniz: Just curious, related to some difficulties with TRT ?

src/ext/rapidjson/reader.h
```diff
@@ -497,6 +497,7 @@ class GenericReader {
                         RAPIDJSON_PARSE_ERROR(kParseErrorTermination, is.Tell());
                     else
                         return;
+                    break;
```
beniz: To avoid compiler warnings ?

CMakeLists.txt
fantes: ok, done

CMakeLists.txt
fantes: - here I use system protobuf because in most tensorRT cases, it would be the only backend used (an so the only available protobuf). 
- as i use FindProtobuf.cmake, i also use cmake helper protobuf_generate_cpp. So getting rid of it does not work (or needs rewrite of src/CMakeLists)
- compiling caffe + tensorrt works with current cmakelists 

so I am not sure of what we shoud do here :)

CMakeLists.txt
fantes: ok, done

src/backends/tensorrt/caffe.proto
fantes: ok, done

src/backends/tensorrt/tensorrtlib.h
fantes: ok, done

src/ext/rapidjson/reader.h
fantes: exactly

src/ext/rapidjson/document.h
fantes: do not remember, seems ok w/o, removed

src/backends/tensorrt/protoUtils.h
fantes: ok, done

CMakeLists.txt
fantes: finally changed a a lot CMakeLists in order not to use find_package and use the same hacks as for caffe2/TF already included protobufs

fantes: done,  tested ok

YaYaB:
Really nice !
Do you have any benchmark on this? (comparing inference time between caffe and tensor rt)

beniz:
We do, for embedded devices. 

[Jolibrain_DeepDetect_NVidia_Caffe_TensorRT_benchmarks_june_2019.pdf](https://github.com/jolibrain/deepdetect/files/3307316/Jolibrain_DeepDetect_NVidia_Caffe_TensorRT_benchmarks_june_2019.pdf)

These results are to be updated and made public.

PR #581, opened 2019-05-17
check for openCV 2 or 3, but not 4

fantes:
change CMakeList to compile dede against openCV 3 or 2, but do not try OpenCV 4, as there are a lot of API changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -385,7 +385,10 @@ include_directories("${TF_INC_DIR}")
 endif() # USE_TF
 
 # OpenCV
-find_package(OpenCV REQUIRED)
+find_package(OpenCV 3 QUIET)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -385,6 +385,10 @@ include_directories("${TF_INC_DIR}")
 endif() # USE_TF
 
 # OpenCV
+find_package(OpenCV 3 QUIET)
+if (NOT OpenCV_FOUND)
+  find_package(OpenCV 2 REQUIRED)
+endif()
```
beniz: There seems to be yet another find_package for opencv below this line, do we need it ?

fantes: no we dont, fixed

PR #586, opened 2019-06-04
replace db:true by db:false in json files when copying models

fantes:
simple change to replace db:true by db:false in json files when copying models , as reported by laura a few weeks ago

### DIFF
--- a/src/backends/caffe/caffemodel.cc
+++ b/src/backends/caffe/caffemodel.cc
@@ -212,6 +212,10 @@ namespace dd
               {
                 std::vector<std::string> selts = dd_utils::split((*hit), '/');
                 fileops::copy_file((*hit), target_repo + '/' + selts.back());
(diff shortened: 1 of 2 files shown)


src/backends/caffe/caffemodel.cc
```diff
@@ -200,6 +200,8 @@ namespace dd
 	      {
 		std::vector<std::string> selts = dd_utils::split((*hit),'/');
 		fileops::copy_file((*hit),target_repo + '/' + selts.back());
+        if (selts.back().find(".json") != std::string::npos)
+          fileops::replace_string_in_file(target_repo + '/' + selts.back(), "db:true", "db:false");
```
beniz: The `db` from json in file should have quotes. Have you been able to test the PR by publishing a model by any chance ?

fantes: corrected, sorry for that

PR #594, opened 2019-06-16
Regularize bad bbox to the boundaries of the image

YaYaB:
Instead of removing bad bbox, this PR regularizes the size of the bbox to the boundaries of the image.
This can be usefull when an object is not completly present in the image (the bounding box might exceed the size of the image)


EBazarov:
@YaYaB nicely done !

PR #601, opened 2019-07-03
V0.9.1 dlib 19.17 build fix

cchadowitz-pf:
- Updated dlib to v19.17 from v19.13
- Change deps order in cmakelists so that building with dlib enabled doesn't cause link error. *note* I haven't tested this with every other backend lib as I haven't ever built it with a number of them.


beniz:
Thanks

PR #611, opened 2019-07-22
Integration of Libtorch in DeepDetect for image classification

BynaryCobweb:
- Libtorch backend
- Libtorch image input connector
- Image classification using pytorch traced models
- Downloading and compiling libtorch automatically during DeepDetect compilation
- Unit tests on libtorch backend

Building instructions:
```
cmake .. -DUSE_TORCH=ON
```

### DIFF
--- a/src/backends/torch/torchinputconns.h
+++ b/src/backends/torch/torchinputconns.h
@@ -1,0 +1,114 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2019 Jolibrain
+ * Author: Louis Jean <ljean@etud.insa-toulouse.fr>
+ *
+ * This file is part of deepdetect.
+ *
+ * deepdetect is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * deepdetect is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with deepdetect.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef TORCHINPUTCONNS_H
+#define TORCHINPUTCONNS_H
+
+#include "imginputfileconn.h"
+#include "txtinputfileconn.h"
+
+#include <vector>
+
+namespace dd
+{
+    class TorchInputInterface
+    {
+    public:
+      TorchInputInterface() {}
+      TorchInputInterface(const TorchInputInterface &i)
+      {
+
+      }
+
+      ~TorchInputInterface() {}
+    };
+
+    class ImgTorchInputFileConn : public ImgInputFileConn, public TorchInputInterface
+    {
+    public:
+        ImgTorchInputFileConn()
+            :ImgInputFileConn() {}
+        ImgTorchInputFileConn(const ImgTorchInputFileConn &i)
+            :ImgInputFileConn(i),TorchInputInterface(i) {}
+        ~ImgTorchInputFileConn() {}
+
+        // for API info only
+        int width() const
+        {
+            return _width;
+        }
+
+        // for API info only
+        int height() const
+        {
+            return _height;
+        }
+
+        void init(const APIData &ad)
+        {
+            ImgInputFileConn::init(ad);
+        }
+        
+        void transform(const APIData &ad)
+        {
+            try
+            {
+                ImgInputFileConn::transform(ad);
+            }
+            catch(const std::exception& e)
+            {
+                throw;
+            }
+
+            if (ad.has("parameters"))
+            {
+                APIData ad_param = ad.getobj("parameters");
+                if (ad_param.has("input"))
+                {
+                    APIData input_ad = ad_param.getobj("input");
+                    if (input_ad.has("std"))
+                        _std = input_ad.get("std").get<double>();
+                }
+            }
+
+            std::vector<at::Tensor> tensors;
+            std::vector<int64_t> sizes{ _height, _width, 3 };
+            at::TensorOptions options(at::ScalarType::Byte);
+
+            for (const cv::Mat &bgr : this->_images) {
+                at::Tensor imgt = torch::from_blob(bgr.data, at::IntList(sizes), options);
+                imgt = imgt.toType(at::kFloat).permute({2, 0, 1});
+                if (_std != 1.0)
+                    imgt = imgt.mul(1. / _std);
+                tensors.push_back(imgt);
+            }
+
+            _in = torch::stack(tensors, 0);
+        }
+
+    public:
+        at::Tensor _in;
+
+        double _std = 1.0;
+    };
+} // namespace dd
+
+#endif // TORCHINPUTCONNS_H
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -1,0 +1,158 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2019 Jolibrain
+ * Author: Louis Jean <ljean@etud.insa-toulouse.fr>
+ *
+ * This file is part of deepdetect.
+ *
+ * deepdetect is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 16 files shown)


src/backends/torch/torchinputconns.h
```diff
@@ -0,0 +1,109 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2018 Jolibrain
```
beniz: This is 2019 :)

src/backends/torch/torchinputconns.h
```diff
@@ -0,0 +1,109 @@
...
+            }
+            catch(const std::exception& e)
+            {
+                throw;
+            }
+
+            std::vector<at::Tensor> tensors;
+
+            for (const cv::Mat &bgr : this->_images) {
+                _height = bgr.rows;
```
beniz: `width` and `height` are usually set at service creation via the options of the input connector. Since they are class variables we may not want to set them up again here within the loop.

src/backends/torch/torchinputconns.h
```diff
@@ -0,0 +1,109 @@
...
+                _width = bgr.cols;
+
+                std::vector<int64_t> sizes{ _height, _width, 3 };
+                at::TensorOptions options(at::ScalarType::Byte);
```
beniz: Can we put this object outside of the loop since it's common to all images. Since inference can process batches of many hundreds of images, many times, we may want to even lightly optimize as needed here.

src/backends/torch/torchinputconns.h
```diff
@@ -0,0 +1,109 @@
...
+                at::Tensor imgt = torch::from_blob(bgr.data, at::IntList(sizes), options);
+                imgt = imgt.toType(at::kFloat).mul(1./255.).permute({2, 0, 1});
```
beniz: Let's keep the normalization and channel permutation from BGR here, but let's add an option that would default to `true` for now. The reason is that once we start using models that are trained via DD and libtorch, normalization and especially permutation may not bee needed anymore.

With other backends, we typically use `std` in input connector to specify the normalization factor (here 255). 

DD uses OpenCV and thus is BGR. Instead of permuting, we may want to add an `rgb` boolean API parameter to the input connector, and let our main `ImgInputConn` do the conversion ?

src/backends/torch/torchinputconns.h
beniz: `width` and `height` are set at service creation via the input connector parameters, let's not reset them within this loop, as they are global variables.

src/backends/torch/torchinputconns.h
beniz: This is already 2019 :)

src/backends/torch/torchinputconns.h
beniz: With other backends we use the `std` input connector API parameter to specify the normalizing factor, here 255. 

For RGB/BGR, let's do it before hitting Torch, in InputImgConn, by adding an input connector API boolean parameter 'rgb' (default to false).

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2018 Jolibrain
```
beniz: 2019 it is , Blade Runner year!

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
...
+
+    /*- from mllib -*/
+    template <class TInputConnectorStrategy, class TOutputConnectorStrategy, class TMLModel>
+    void TorchLib<TInputConnectorStrategy, TOutputConnectorStrategy, TMLModel>::init_mllib(const APIData &ad) 
+    {
+        bool gpu = false;
+        APIData lib_ad = ad.getobj("parameters").getobj("mllib");
+
+        if (lib_ad.has("gpu")) {
+            gpu = lib_ad.get("gpu").get<bool>() && torch::cuda::is_available();
```
beniz: We may want to add support for `gpuid` here, that allows to place the job on a given GPU.

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
...
+        }
+        if (lib_ad.has("nclasses")) {
+            _nclasses = lib_ad.get("nclasses").get<int>();
+        }
+
+        _device = gpu ? torch::Device("cuda") : torch::Device("cpu");
+
+        _traced = torch::jit::load(this->_mlmodel._model_file);
```
beniz: This code only works for traced model, right ?

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
...
+
+        TInputConnectorStrategy inputc(this->_inputc);
+        TOutputConnectorStrategy outputc;
+        try {
+            inputc.transform(ad);
+        } catch (...) {
+            throw;
+        }
+
+        Tensor output = _traced->forward({inputc._in.to(_device)}).toTensor();
```
beniz: We want to try / catch around the forward call so that we can have a message within the `catch` scope and before the `throw`. 

This is because there are many memory and network related common errors, and we want to signify what the error is. For this reason we may want to `catch(std::exception &e)` report `e.what()` in the error message.

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
...
+        std::vector<APIData> results_ads;
+
+        for (int i = 0; i < output.size(0); ++i) {
+            std::tuple<Tensor, Tensor> sorted_output = output.slice(0, i, i + 1).sort(1, true);
+
+            APIData results_ad;
+            std::vector<double> probs;
+            std::vector<std::string> cats;
+
+            if (output_params.has("best")) {
```
beniz: You should be able to put the test on `best` outside of the loop since it bears the same value for every output.

src/backends/torch/torchlib.cc
```diff
@@ -0,0 +1,142 @@
...
+                    int index = std::get<1>(sorted_output).slice(1, i, i + 1).item<int>();
+                    cats.push_back(this->_mlmodel.get_hcorresp(index));
+                }
+            }
+
+            results_ad.add("uri", inputc._uris.at(results_ads.size()));
+            results_ad.add("loss", 0.0);
+            results_ad.add("cats", cats);
+            results_ad.add("probs", probs);
+            results_ad.add("nclasses", 4);
```
beniz: This should be `_nclasses` ?

beniz:
FYI, a build on a CPU only machine with `cmake -DUSE_CAFFE=OFF -DUSE_TORCH=ON ..` fails with the error below:

```
[ 32%] Built target pytorch
[ 92%] Built target ddetect
make[2]: *** No rule to make target 'pytorch/src/pytorch/torch/lib/libc10_cuda.so', needed by 'main/dede'.  Stop.
CMakeFiles/Makefile2:177: recipe for target 'main/CMakeFiles/dede.dir/all' failed

```

beniz:
FYI, to work on CPU at least, it is required to install `libiomp5`. So we should add it to the README I believe.

BynaryCobweb:
Fixed everything mentioned above, except the libiomp loading error.

BynaryCobweb:
[Script](https://gist.github.com/BynaryCobweb/5ed750850a4880108395cbed4a7bbbd5) to trace resnet and other models.

Requirements:
```
pip3 install --user pytorch torchvision
```

BynaryCobweb:
<!> A segfault may occur when running a model with CUDA

src/backends/torch/torchlib.cc
BynaryCobweb: Yes, only traced models are currently supported.

beniz:
@BynaryCobweb Confirmed that the final linking is fixed on CPU builds!

BynaryCobweb:
## Example: Image classification
* Download the model
```
wget https://www.deepdetect.com/dd/examples/torch/resnet50_torch.tar.gz
tar xvf resnet50_torch.tar.gz
```
* Run dede
* Start the service
```
curl -X PUT "http://localhost:8080/services/torch_resnet" -d '{
    "description": "image classification service",
    "mllib": "torch",
    "model": {
        "repository": "./resnet50_torch/"
    },
    "parameters": {
        "input": {
            "connector": "image"
        }
    },
    "type": "supervised"
}
'
```
* Run predict to guess the class of an image
```
    "service": "torch_resnet",
    "parameters": {
        "input": {
            "width":224,
            "height":224,
            "rgb":true,
            "std":255.0
        },
        "output": {
            "best":3
        }
    },
    "data":["./resnet50_torch/cat.jpg"]
}
'
```

BynaryCobweb:
- [x] squashed all commits in one

PR #617, opened 2019-07-31
TensorRT better exception catching around inference

beniz:


beniz:
I'm still able to crash TRT on a Nano when overloading the GPU, despite the exception catching:
```
#0  0x0000007fae3ed4d8 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x0000007fae3ee8b4 in __GI_abort () at abort.c:79
#2  0x0000007fae3e6b44 in __assert_fail_base (fmt=0x7fae4e20c0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7faff81df8 "status == STATUS_SUCCESS", file=file@entry=0x7faff82618 "nmsPlugin.cpp", line=line@entry=102, function=function@entry=0x7faff82490 "virtual int nvinfer1::plugin::DetectionOutput::enqueue(int, const void* const*, void**, void*, cudaStream_t)") at assert.c:92
#3  0x0000007fae3e6bc4 in __GI___assert_fail (assertion=0x7faff81df8 "status == STATUS_SUCCESS", file=0x7faff82618 "nmsPlugin.cpp", line=102, function=0x7faff82490 "virtual int nvinfer1::plugin::DetectionOutput::enqueue(int, const void* const*, void**, void*, cudaStream_t)") at assert.c:101
#4  0x0000007faff62f6c in  () at /usr/lib/aarch64-linux-gnu/libnvinfer_plugin.so.5
#5  0x0000007fb068ea94 in nvinfer1::rt::ExecutionContext::enqueue(int, void**, CUstream_st*, CUevent_st**) ()
    at /usr/lib/aarch64-linux-gnu/libnvinfer.so.5
#6  0x00000055559d65ac in dd::TensorRTLib<dd::ImgTensorRTInputFileConn, dd::SupervisedOutput, dd::TensorRTModel>::predict(dd::APIData const&, dd::APIData&) (this=this@entry=0x7f69779060, ad=..., out=...)
    at /home/jolibrainjetson/code/deepdetect/src/backends/tensorrt/tensorrtlib.cc:438
#7  0x00000055559b4ad8 in dd::MLService<dd::TensorRTLib, dd::ImgTensorRTInputFileConn, dd::SupervisedOutput, dd::TensorRTModel>::predict_job(dd::APIData const&, dd::APIData&, bool const&) (chain=@0x7f73ffc7a0: true, out=..., ad=..., this=0x7f69779060) at /home/jolibrainjetson/code/deepdetect/src/mlservice.h:445
#8  0x00000055559b4ad8 in dd::visitor_predict::operator()<dd::MLService<dd::TensorRTLib, dd::ImgTensorRTInputFileConn, dd::SupervisedOutput, dd::TensorRTModel> >(dd::MLService<dd::TensorRTLib, dd::ImgTensorRTInputFileConn, dd::SupervisedOutput, dd::TensorRTModel>&) (mllib=..., this=0x7f73ffc730)
    at /home/jolibrainjetson/code/deepdetect/src/services.h:177
```

It seems `nvinfer` sends a SIG_ABRT, see https://stackoverflow.com/questions/3413166/when-does-a-process-get-sigabrt-signal-6

fantes:
beware, this PR has chain_basics2 as target

fantes:
I am not suprised by such behavior of TRT, I seemed to me that I had the same at some point, TRT does not use exceptions, (because it is mainly C) and does not propagate error codes (as it is mainly very low quality C) and the worst of all is that we cannot change this sh***y behavior. 
At this point, the only thing to do is to write something on nvidia forums and hope for a corrected behavior in some future release

fantes:
indeed there is something we can do in this very precise case : the assert that causes the sigabrt is in nmsPlugin.cpp and we have the source of it (as it is part of the few released trt code), so we can link against our own plugin. This is what is done in the unmerged/not_even_PRed_yet fantes/trt_lstm branch

PR #636, opened 2019-09-10
make dede tencent/ncnn/master compatible

fantes:
this PR makes deepdetect compatible with ncnn with our modifs + tencent head

- expose lightmode option, defautls to true:  ncnn maximize inplace ops and releases mem as soon as possible in case of light mode, needs linear net, does not work in case of resnets, for instance.


beniz:
I believe you may need to rebase to master

fantes:
integration failed because disk full on jenkins'host ...

PR #640, opened 2019-09-13
input image interpolation method selection via API

beniz:
This PR allows to control OpenCV interpolation method via the API:

```json
{
"parameter": {
  "input": {"connector":"image", "interp": "linear"}
 }
}
```
The `interp` API parameter takes values from `"linear","nearest","cubic","area","lanczos4"`. 

Fastest methods are `"nearest"` then `"linear"`, default remains `"cubic"`.


src/imginputfileconn.h
```diff
@@ -594,6 +612,7 @@ namespace dd
     int _scale_min = 600;
     int _scale_max = 1000;
     bool _keep_orig = false;
+    std::string _interp = "cubic";
```
fantes: same as cuda_resize : i think this _interp should be copyed at copy_contruction time

PR #642, opened 2019-09-16
OpenCV GPU support with CUDA for image input resizing

beniz:
This PR adds OpenCV GPU support (NVidia only) for image input resizing operation. This functionality **requires an OpenCV builds with CUDA support**.

API changes
- `input.cuda` boolean to resize input images on GPU

Build flag
- `-DUSE_CUDA_CV=ON` passed to `cmake`

Results:
- Early tests on Jetson TX2 shows up to 20% speed gain on single image with large size (e.g. 4000x4000)
- Speedup may vary depending on GPU

### DIFF
--- a/src/imginputfileconn.h
+++ b/src/imginputfileconn.h
@@ -27,6 +27,13 @@
 #ifdef USE_CUDA_CV
 #include <opencv2/cudaimgproc.hpp>
 #endif
+#if CV_VERSION_MAJOR >= 3
+#define CV_LOAD_IMAGE_COLOR cv::IMREAD_COLOR
+#define CV_LOAD_IMAGE_GRAYSCALE cv::IMREAD_GRAYSCALE
+#define CV_LOAD_IMAGE_UNCHANGED cv::IMREAD_UNCHANGED
+#define CV_BGR2RGB cv::COLOR_BGR2RGB
+#define CV_INTER_CUBIC cv::INTER_CUBIC
+#endif
 #include "ext/base64/base64.h"
 #include "utils/apitools.h"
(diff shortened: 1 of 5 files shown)


src/imginputfileconn.h
```diff
@@ -613,6 +650,9 @@ namespace dd
     int _scale_max = 1000;
     bool _keep_orig = false;
     std::string _interp = "cubic";
+#ifdef USE_CUDA_CV
+    bool _cuda = false;
```
fantes: should not be this attribute value copied during ImgInputFileConn copy construction, somewhere around line 386?

beniz: Ohlala, good catch, thanks!

src/imginputfileconn.h
```diff
@@ -69,22 +79,45 @@ namespace dd
...
       {
 	std::vector<unsigned char> vdat(str.begin(),str.end());
-	cv::Mat img = cv::Mat(cv::imdecode(cv::Mat(vdat,true),
+	cv::Mat img = cv::Mat(cv::imdecode(cv::Mat(vdat,false),
                                      _unchanged_data ? CV_LOAD_IMAGE_UNCHANGED :
                                      (_bw ? CV_LOAD_IMAGE_GRAYSCALE : CV_LOAD_IMAGE_COLOR)));
-        if (_rgb)
-          cv::cvtColor(img, img, CV_BGR2RGB);
-	if (_keep_orig)
+      	if (_keep_orig)
```
BynaryCobweb: Was support for bgr to rgb conversion removed here ?

PR #646, opened 2019-10-06
fixed ambiguous var with TF input connector

beniz:


src/backends/tf/tfinputconns.h
```diff
@@ -44,14 +44,13 @@ namespace dd
   public:
     TFInputInterface() {}
     TFInputInterface(const TFInputInterface &tii)
-    :_dv(tii._dv),_ids(tii._ids){}
+    :_dv(tii._dv)
```
cchadowitz-pf: Get a compile error here, I think you're missing a trailing set of `{}`

beniz: Oops, bad sunday fix. Thanks.

PR #649, opened 2019-10-11
expose overlap_threshold for tests in detection tasks

fantes:
overlap_threshold is a hard threshold over intersection ratio of bboxes. computed bboxes that do not overlap ground truth bboxes more than these ratio are completely discarded.

overlap ratio is IoU (intersection over union):
intersection_size / (bbox1_size + bbox2_size - intersection_size)
see for instance  https://www.pyimagesearch.com/2016/11/07/intersection-over-union-iou-for-object-detection/
1.0 is perfect match
0.0 is complete miss (null intersection)

example usage : 
"service":"detection_service",
"parameters":{
"output":{
"measure":"map"
"overlap_threshold" : 0.7
},
"data":["test_data"]
}'

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -30,6 +30,7 @@
 #include "utils/utils.hpp"
 #include "utils/apitools.h"
 #include "caffe/sgd_solvers.hpp"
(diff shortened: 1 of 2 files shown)


beniz:
As discussed, we'd rather get several map-X measures at once. Closing, to be resumed :)

Issue #647, opened 2019-10-08
Infinite prediction call (prediction stuck because of data retrieval)


YaYaB:
#### Configuration
- Version of DeepDetect:
    - Locally compiled on Ubuntu 16.04 LTS
- Commit (shown by the server when starting):
    - Branch `Master`, 3d65432ecba9f766527aa058c15c6ac199ae9d2a
- GPUS:
    - 1 x Nvidia GTX Titan X (Maxwell)

#### Your question / the problem you're facing:
I have an issue when I want to make a prediction on an image when this image is too heavy or when it is very slow to retrieve.

#### Error message (if any) / steps to reproduce the problem:
- Launch Dede

Api call
```
./dede --port 8080
```
Serveur log output
```
DeepDetect [ commit 3d65432ecba9f766527aa058c15c6ac199ae9d2a ]
[2019-10-08 15:21:36.337] [api] [info] Running DeepDetect HTTP server on localhost:8080
```

- Launch netcat to simulate and image that can not be retrieved.
```
nc -l -p 8888 
```

- Create service

Api call
```
curl -X PUT "http://localhost:8080/services/imageserv" -d '{
      "mllib":"caffe",
      "description":"image classification service",
      "type":"supervised",
      "parameters":{
        "input":{
          "connector":"image",
          "width":224,
          "height":224
        },
        "mllib":{
          "nclasses":101
        }
      },
      "model":{
        "repository":"TODO/classification_model"
      }
    }'
```
Serveur log output
```
{"status":{"code":201,"msg":"Created"}}
```
- Create Prediction

Api call
```
curl -X POST "http://localhost:8080/predict" -d '{
      "service":"imageserv",
      "parameters":{
        "input":{
          "width":224,
          "height":224
        },
        "output":{
          "best":3
        }
      },
      "data":["http://127.0.0.1:8888"]
    }'
```

Serveur log output:
```
It's empty, nothing happens because it is still trying to download the image
```

The issue comes from the fact that no timeout is indicated for the retrieve of the data.
After checking a bit the code I found here https://github.com/jolibrain/deepdetect/blob/20c143f27a110f8fff849347dc2b42b9add92df4/src/imginputfileconn.h#L487 that we try to read the element from the **uri**.
That lead us to https://github.com/jolibrain/deepdetect/blob/9978230e916a9e30a0be4efe2c6678f7139601d9/src/inputconnectorstrategy.h#L61 where we see that we do an **http call** to retrieve the data.
[...]

PR #650, opened 2019-10-14
Added input connector timeout, ref #647

beniz:
This PR adds timeout control to the input connector, via `parameters.input.timeout`:
- at service creation, and stays the default value through the service's life
- can be overridden with every `/predict` call

Additional elements:
- default timeout is set to 600 seconds (10 mins)
- there's a max timeout value set to 36000 seconds (10 hours), any timeout set above that threshold leads to a 400 error.

### DIFF
--- a/src/backends/caffe/caffeinputconns.cc
+++ b/src/backends/caffe/caffeinputconns.cc
@@ -1361,7 +1361,7 @@ namespace dd
     if (ad_input.has("label_offset"))
     _label_offset = ad_input.get("label_offset").get<int>();*/
     
-    DataEl<DDCCsv> ddcsv;
+    DataEl<DDCCsv> ddcsv(this->_input_timeout);
     ddcsv._ctype._cifc = this;
     ddcsv._ctype._adconf = ad_input;
     ddcsv.read_element(_csv_fname,this->_logger);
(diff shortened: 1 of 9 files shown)


YaYaB:
It works very well on my side. I've tested in:
- the service creation alone;
- the predict request alone;
- the predict and the creation requests and the value in the predict call overrides well the value indicated in the creation request.

Few other remarks to understand what's done behind:
- The prediction value is overridden only for this prediction. The default value set in the service creation is kept and used if no timeout is indicated in the prediction call;
- If no value is indicated in the creation service, the prediction's value is used. However the default value still remains as None. If prediction A has a timeout of 5 and prediction B has no timeout then prediction B will be infinite
===> that's a correct behavior for me

I've tried to put a value superior to the max timeout value however I get an incorrect message:
```
{"status":{"code":400,"msg":"BadRequest","dd_code":1005,"dd_msg":"Service Input Error: timeout value is above max default timeout (6000)"}}
```
Indeed it should be **36000** instead of **6000**.

Finally I've tried the default timeout, and it did timeout correctly at 10 minutes.
```
{"status":{"code":400,"msg":"BadRequest","dd_code":1005,"dd_msg":"Service Input Error: Operation timed out after 600001 milliseconds with 0 bytes received"}}
```

beniz:
Thanks for the thorough review @YaYaB 

>     * The prediction value is overridden only for this prediction. The default value set in the service creation is kept and used if no timeout is indicated in the prediction call;

That's correct.

> If prediction A has a timeout of 5 and prediction B has no timeout then prediction B will be infinite

Not infinite since there's a default timeout value.

> Indeed it should be **36000** instead of **6000**.

OK, thanks, we'll fix it.

YaYaB:
Yep indeed, it's not infinite but default timeout (lost in my writing ^^)

PR #652, opened 2019-10-15
Raw detection output if measure = raw

fantes:
WIP : 
we need to decide how to pass the data wrt prediction/test   memorydatalayer/AnnotatedDataLayer

### DIFF
--- a/src/supervisedoutputconnector.h
+++ b/src/supervisedoutputconnector.h
@@ -683,6 +683,13 @@ namespace dd
                  meas_out.add(s,static_cast<double>(ap.second));
 	       }
 	    }
+      bool raw = (std::find(measures.begin(),measures.end(),"raw")!=measures.end());
+      if (raw)
+        {
+          std::vector<std::string> clnames = ad_res.get("clnames").get<std::vector<std::string>>();
+          APIData ap = raw_detection_results(ad_res,clnames);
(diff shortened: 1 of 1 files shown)


src/supervisedoutputconnector.h
```diff
@@ -1542,7 +1549,71 @@ namespace dd
...
+      APIData bad = ad.getobj("0");
+      int pos_count = ad.get("pos_count").get<int>();
+      for (int i=0;i<pos_count;i++)
+        {
+          std::vector<APIData> vbad = bad.getv(std::to_string(i));
+          //std::cerr << "vbad size=" << vbad.size() << std::endl;
+          for (size_t j=0;j<vbad.size();j++)
+            {
+              std::vector<double> tp_d = vbad.at(j).get("tp_d").get<std::vector<double>>();
+              //std::cerr << "tp_d size=" << tp_d.size() << std::endl;
```
beniz: Let's remove comments before merge! tks.

beniz:
Setting `measure:["raw"]` is enough, either for training or predict calls ?

src/supervisedoutputconnector.h
fantes: done

fantes:
> Setting `measure:["raw"]` is enough, either for training or predict calls ?

yes it is !

PR #655, opened 2019-10-23
Random crops action in `/chain`

beniz:
WIP: This PR adds random actions to `/chain`. This is useful for image similarity search.


dgtlmoon:
`This is useful for image similarity search.` Hmmm example? :)

PR #660, opened 2019-10-26
Chains: avoid encoding + serializing crops, passing raw cv::Mat instead

beniz:
This PR should speed up chains with many crops, removing the encoding and serializing steps.

### DIFF
--- a/src/imginputfileconn.h
+++ b/src/imginputfileconn.h
@@ -490,11 +490,44 @@ namespace dd
     {
       return _test_images.size();
     }
+
+    void get_data(const APIData &ad)
+    {
+      // check for raw cv::Mat
+      if (ad.has("data_raw_img"))
+	{
+	  if (ad.has("ids"))
+	    _ids = ad.get("ids").get<std::vector<std::string>>();
+	  if (ad.has("meta_uris"))
+	    _meta_uris = ad.get("meta_uris").get<std::vector<std::string>>();
+	  if (ad.has("index_uris"))
+	    _index_uris = ad.get("index_uris").get<std::vector<std::string>>();
+	  
+	  _images = ad.get("data_raw_img").get<std::vector<cv::Mat>>();
+	  std::vector<cv::Mat> rimgs;
+	  std::vector<std::string> uris;
+	  int i = 0;
+	  for (auto img: _images)
+	    {
+	      cv::Mat rimg;
+	      resize(img,rimg,cv::Size(_width,_height),0,0);
+	      _images_size.push_back(std::pair<int,int>(img.rows,img.cols));
+	      if (_keep_orig)
+		_orig_images.push_back(std::move(img));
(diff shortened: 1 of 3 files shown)


src/imginputfileconn.h
```diff
@@ -490,11 +490,44 @@ namespace dd
...
+      // check for raw cv::Mat
+      if (ad.has("data_raw_img"))
+	{
+	  if (ad.has("ids"))
+	    _ids = ad.get("ids").get<std::vector<std::string>>();
+	  if (ad.has("meta_uris"))
+	    _meta_uris = ad.get("meta_uris").get<std::vector<std::string>>();
+	  if (ad.has("index_uris"))
+	    _index_uris = ad.get("index_uris").get<std::vector<std::string>>();
+
```
fantes: why not replace these 6 lines by InputConnectorStrategy::get_data(ad); ? 
changing the structure a bit would make things a bit more clear :
```
void get_data(ad)
{
   InputConnectorStrategy::get_data(ad);
   if (ad.has("data_raw_img"))
   {
      // all raw img stuff
   }
}
```

beniz: Because if "data" is empty, since data is in "data_raw_img", the `get_data` call fails. 
I agree it's hacky though and that it'd be good to refactor the calls in transform so that they can be reused here.

fantes: ok :)

src/imginputfileconn.h
```diff
@@ -490,11 +491,50 @@ namespace dd
...
       return _test_images.size();
     }
+
+    void get_data(const APIData &ad)
+    {
+      // check for raw cv::Mat
+      if (ad.has("data_raw_img"))
+	{
+	  if (ad.has("ids"))
+	    _ids = ad.get("ids").get<std::vector<std::string>>();
```
fantes: if "ids" is not defined,  we have an exception at caffeinputconnh.h:250 (in transform)
 _imgs_size.insert(std::pair<std::string,std::pair<int,int>>(this->_ids.at(i),this->_images_size.at(i)));     
because ids_ will be empty

beniz: Good catch, could you try the patch below ?
```
diff --git a/src/imginputfileconn.h b/src/imginputfileconn.h
index 91d927b2..e0c8f71b 100644
--- a/src/imginputfileconn.h
+++ b/src/imginputfileconn.h
@@ -523,6 +523,11 @@ namespace dd
 		_orig_images.push_back(std::move(img));
 	      if (!_ids.empty())
 		uris.push_back(_ids.at(i));
+	      else
+		{
+		  _ids.push_back(std::to_string(i));
+		  uris.push_back(_ids.back());
+		}
 	      rimgs.push_back(std::move(rimg));
 	      ++i;
 	    }
```

fantes: ok for me , tested :)

PR #669, opened 2019-12-05
libtorch 1.3.1 

fantes:
WIP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -706,7 +706,7 @@ if (USE_TORCH)
   set(PYTORCH_PATCHES_PATH ${CMAKE_BINARY_DIR}/patches/pytorch)
 
   set(PYTORCH_PATCHES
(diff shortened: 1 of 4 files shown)


beniz:
The CI build fails with this linking error: 

```
/usr/bin/ld: cannot find -liomp5
collect2: error: ld returned 1 exit status
```

We did see this before, this is a library used for CPU builds (Intel OpenMP). It's possible the lib path is wrong, as the `libiomp5` is already installed on the CI machine, and is linking fine with pytorch on the master branch.

fantes:
ca

> The CI build fails with this linking error:
> 
> ```
> /usr/bin/ld: cannot find -liomp5
> collect2: error: ld returned 1 exit status
> ```
> 
> We did see this before, this is a library used for CPU builds (Intel OpenMP). It's possible the lib path is wrong, as the `libiomp5` is already installed on the CI machine, and is linking fine with pytorch on the master branch.

I cannot reproduce this ... building seems correct now

Issue #674, opened 2019-12-12
in detection models (ssd), in predict output :  ymin and ymax are swapped 


fantes:
ie ymax < ymin , 

- change in dd to do here : https://github.com/jolibrain/deepdetect/blob/master/src/backends/caffe/caffelib.cc#L2560  (replace ymax by ymin and reverse) , reference in caffe :  https://github.com/jolibrain/caffe/blob/master/src/caffe/util/bbox_util.cpp#L2105  (same in cuda) 

- change in platform to do here (ymin - ymax)
https://github.com/alx/react-bounding-box/blob/master/src/react-bounding-box.js#L394

YaYaB:
Please make a release when this is pushed to master as it has a big impact for those using detection models.

fantes:
https://github.com/jolibrain/deepdetect/pull/690

alx:
Fixed in `react-bounding-box` v0.5.3: https://www.npmjs.com/package/react-bounding-box/v/0.5.3

fantes:
> Fixed in `react-bounding-box` v0.5.3: https://www.npmjs.com/package/react-bounding-box/v/0.5.3

:+1:

PR #690, opened 2020-01-16
Fixed inversion of ymin and ymax in detection models output

fantes:
FIX https://github.com/jolibrain/deepdetect/issues/674

- ymin and ymax were previoulsy swapped, ie ymin was > ymax 
- this patch correctly set ymin and ymax so that it respects caffe detection_ouput layer semantics as for instance in https://github.com/jolibrain/caffe/blob/master/src/caffe/layers/detection_output_layer.cpp#L368

- when using output of detection models, if ymin and ymax are used as is, then problems are to be exepcted, but if they are used with somthing like height = abs(ymin-ymax) , then this patch will not change anything.

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -2557,9 +2557,9 @@ namespace dd
 			cats.push_back(this->_mlmodel.get_hcorresp(detection[1]));
 			APIData ad_bbox;
 			ad_bbox.add("xmin",static_cast<double>(detection[3]*(cols-1)));
-			ad_bbox.add("ymax",static_cast<double>(detection[4]*(rows-1)));
+			ad_bbox.add("ymin",static_cast<double>(detection[4]*(rows-1)));
 			ad_bbox.add("xmax",static_cast<double>(detection[5]*(cols-1)));
-			ad_bbox.add("ymin",static_cast<double>(detection[6]*(rows-1)));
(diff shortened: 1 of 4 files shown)


beniz:
Thanks. Beware I believe this PR may break image cropping in `chains`, especially this line: https://github.com/jolibrain/deepdetect/blob/master/src/chain_actions.cc#L91

Maybe update the PR so that it becomes:
```
cv::Rect roi(cxmin,cymax,cxmax-cxmin,cymax-cymin);
```

beniz:
Also, this line may need to be modified, https://github.com/jolibrain/deepdetect/blob/master/src/chain_actions.cc#L89

fantes:
> Thanks. Beware I believe this PR may break image cropping in `chains`, especially this line: https://github.com/jolibrain/deepdetect/blob/master/src/chain_actions.cc#L91
> 
> Maybe update the PR so that it becomes:
> 
> ```
> cv::Rect roi(cxmin,cymax,cxmax-cxmin,cymax-cymin);
> ```

done

fantes:
> Thanks. Beware I believe this PR may break image cropping in `chains`, especially this line: https://github.com/jolibrain/deepdetect/blob/master/src/chain_actions.cc#L91
> 
> Maybe update the PR so that it becomes:
> 
> ```
> cv::Rect roi(cxmin,cymax,cxmax-cxmin,cymax-cymin);
> ```

done

PR #696, opened 2020-01-31
engine selection at service creation time : CAFFE of CUDNN

fantes:
this PR allows to select engine (cudnn or caffe (which is CPU or GPU depending on GPU availability)) at service creation time, for layers that can, ie conv, deconv, tanh, relu, sigmoid, pooling, lrn, and softmax.

 This is a workaround to memory over consumption / leakage seen with some models that do a _lot_ of cudnncreate 

if unset, use default, which is cudnn is deepdetect+caffe was compiled with -DUSE_CUDNN=ON , or caffe in other case

parameter is parameters.mllib.engine  , can be set to "CUDNN" or "CAFFE"

example: 
curl -X PUT http://localhost:8080/services/ocr -d '{
 "description": "ocr",
 "model": {
  "repository": "/home/infantes/airbusey_comp_ocr_resnext50_v4"
 },
 "mllib": "caffe",
 "type": "supervised",
 "parameters": {
  "input": {
   "connector": "image",
   "width":250,
   "height":200, "ctc":true, "bw":false, "db":true
  },
  "mllib" : {
     "nclasses":40, "finetuning": true,
     "rotate":false,
     "mirror":false,
     "scale":1,
     "engine": "CUDNN",
     "db":true,
     "noise": {
          "all_effects":true,
          "prob":0.01
     },
    "distort": {
        "all_effects":true,
        "prob":0.01
    },
    "geometry" : {
         "persp_horizontal":true,
         "persp_vertical":true,
         "zoom_out":true,
         "pad_mode":"constant",
         "prob":0.001
     },
     "gpu":true,
     "gpuid":1

  }
 }
}'

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -362,7 +362,10 @@ namespace dd
 	// image data augmentation
 	if (typeid(this->_inputc) == typeid(ImgCaffeInputFileConn))
 	  configure_image_augmentation(ad,net_param);
-	    
+
+    update_protofile_engine(net_param, ad);
+    update_protofile_engine(deploy_net_param, ad);
+
 	// adapt number of neuron output
 	update_protofile_classes(net_param);
 	update_protofile_classes(deploy_net_param);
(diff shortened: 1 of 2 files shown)


beniz:
Maybe let's rebase on master.

PR #707, opened 2020-03-03
fix divide by zero in case of very wrong prediction 

fantes:
fix divide by zero metrics in case of very wrong prediction, typically found in short unit tests like bert


fantes:
this is merged somewhere else

PR #723, opened 2020-04-18
moving travis to xenial

beniz:


sileht:
There is not more travis.

PR #726, opened 2020-04-23
EfficientNet B0 to B5 image classification templates for Caffe

beniz:
Pre-trained models are available from https://www.deepdetect.com/models/efficientnet
Input images require a pre-scaling value around 0.0083. 
The pretrained models may not be too useful outside of transfer learning tasks, due to incompatibilities when converting from TF, and they require an adapted scaling factor. Additionnally, it's proved useful in certain cases to freeze the batchnorm layers for transfer learning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,7 +266,7 @@ if (USE_DLIB)
 endif()
 
 set(CAFFE_DD_USER jolibrain)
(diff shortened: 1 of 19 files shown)


BirdeeChen:
Thanks for the excellent work, can you please upload the prototxt as well?

beniz:
The prototxt are part of this PR, you can download them easily.

Issue #732, opened 2020-05-11
Reorganization Docker files and builds instructions


beniz:
PR #706 greatly improved the docker images. However, we need a second pass to clean up / reorganize some elements. See the list below:

- the `build` dir is often volatile in projects that use `cmake`. I suggest that it is turned into `docker` as before the PR as it hosts the Docker build documentation as well.
- the `*.Dockerfile` should not be at the root of the repository, let's have them into the `docker` directory above.
- if the `build.sh` script is generic enough that it can be used as an alternative to the building steps we outline in the documentation, then it could/should go at the root of the repository. If this script eases all builds, then we could re-write the build documentation accordingly in `README.md`.

PR #733, opened 2020-05-12
Reorganization Docker files and builds instructions

quadeare:
PR related to issue #732 

Features : 
- Rollback Docker directory `build` -> `docker`
- Move generic build script at root directory
- Make some changes on documentation

Fixes : 
- Update cmake to version 3.10 on cpu Docker image


beniz:
Thanks. It's good to have the `build.sh` script at the root, but at the moment it seems it cannot be used as is for generic builds. Two paths from here:
- or we move `build.sh` inside `docker` as it works for docker only
- or we modify the `build.sh` so that it can work for all builds.

For the second path, from what I've tried, it fails because :
- doing `mkdir build`, `cd build`, `cp ../build.sh .`, then `./build.sh` needs `cmake ..` and not `cmake .`. 

Also, it'd be useful to be able to control `CUDA_ARCH` values from the menu options (with a `default` value falling back on what we have now).

What do you think ? My preference goes to the second option (generic 'build.sh' script)

PR #734, opened 2020-05-19
[FIX] dice loss for 2 classes + contour weighting

fantes:
- caffe dice connexion with prototxt was still done as in an old version of dice layer, fix is to use general case (ie >2 classes)
- dice contour weighting has a bug (normalization ??) disabling it for now

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -3525,8 +3525,6 @@ void CaffeLib<TInputConnectorStrategy,TOutputConnectorStrategy,TMLModel>::update
 	    if (lparam->has_convolution_param())
 	      {
             int num_output = lparam->mutable_convolution_param()->num_output();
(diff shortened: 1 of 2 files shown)


src/backends/caffe/caffelib.cc
```diff
@@ -4010,18 +3946,16 @@ void CaffeLib<TInputConnectorStrategy,TOutputConnectorStrategy,TMLModel>::update
                     this->_logger->warn("dice contour amplitude unrecognized, float expected");
                   }
               }
+			this->_logger->warn("dice contour is not working (2020 - 05 - 19) disabling it");
+			contours = caffe::DiceCoefLossParameter::NO;
+
```
beniz, line 3951: Let's fail the creation / job if contour is selected, no ?

beniz:
Does this work for 2 classes then ?

fantes:
> Does this work for 2 classes then ?

yes :)

src/backends/caffe/caffelib.cc
fantes: done

PR #738, opened 2020-06-09
[FIX] image -> single target regression

fantes:
this PR fixes loss used when single target regression is used, previously it was working only with normalized values (or was meant to work only with normalized values, but has never been tested AFAIK)
option is parameters.mlllib.loss    
values are 
- "L1" (smoothL1 with smoothing when error < 10-2) , 
- "L2" (euclidean loss) [default],  
- "sigmoid" (sigmoidcrossentropyloss), to be used only for values between 0 and 1

three options have been tested OK

### DIFF
--- a/src/backends/caffe/caffelib.cc
+++ b/src/backends/caffe/caffelib.cc
@@ -319,28 +319,63 @@ namespace dd
       } // end multi_label
     else if (_regression)
       {
-	int k = net_param.layer_size();
+		int k = net_param.layer_size();
         for (int l=k-1;l>0;l--)
           {
             caffe::LayerParameter *lparam = net_param.mutable_layer(l);
             if (lparam->type() == "SoftmaxWithLoss")
               {
-                lparam->set_type("SigmoidCrossEntropyLoss");
-		lparam->clear_include();
+				if ( _loss.empty() || _loss == "L2")
+				  {
+					lparam->set_type("EuclideanLoss");
+				  }
+				else if (_loss == "L1")
+				  {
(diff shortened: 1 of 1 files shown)


beniz:
Thanks, can you squash / cleanup commits before merge ? Ideally, commit https://github.com/jolibrain/deepdetect/pull/738/commits/9939ccd7ac233fafcbd89ce5faf56e33b187aa52 should go elsewhere, if possible.

beniz:
Two remarks:
- Sigmoid CE loss was used for multi-label classification, are we sure we want to remove it ?
- Do we want an option for using L1 instead of MSE when useful ?

fantes:
> Two remarks:
> 
>     * Sigmoid CE loss was used for multi-label classification, are we sure we want to remove it ?

sigmoidCE is still the default for multi label, this fix only applies for single label

 
>     * Do we want an option for using L1 instead of MSE when useful ?

why not, but L1 in caffe needs more work, as it has to be implemented using several layers, maybe e can do this later if needed. as a workaround, smooth l1 can be added quickly.

fantes:
> Thanks, can you squash / cleanup commits before merge ? Ideally, commit [9939ccd](https://github.com/jolibrain/deepdetect/commit/9939ccd7ac233fafcbd89ce5faf56e33b187aa52) should go elsewhere, if possible.

done

beniz:
Thanks, can you add an image regression call or at least the relevant options that need to be set + basic input dataset format, FTR so that we can point to the PR.

Issue #769, opened 2020-08-10
Different predictions with Tensorrt7 for a refinedet model


YaYaB:
#### Configuration
- Version of DeepDetect:
    - [X] Locally compiled on:
        - [ ] Ubuntu 14.04 LTS
        - [X] Ubuntu 18.04 LTS
        - [ ] Mac OSX
        - [ ] Other:
    - [ ] Docker
    - [ ] Amazon AMI
- Commit (shown by the server when starting):
0eac8a49404292529a6ea1810290af800092902c
#### Your question / the problem you're facing:
I successfully installed a version of DD with tensorrt backend and tensorrt-oss.
A refineddet model is correctly loaded however the results obtained are very very different than what we obtain with the original model in caffe.

The following steps will help reproduce:
#### Download the model
```
PATH_MODEL="PATH_WANTED"
mkdir $PATH_MODEL && cd $PATH_MODEL
wget https://deepdetect.com/models/init/desktop/images/detection/faces_512.tar.gz
wget -O test_image.jpeg https://miro.medium.com/max/6528/1*DYUaxku5bfbZaDLW-4SyWg.jpeg
tar -xvf faces_512.tar.gz
```
#### Error message (if any) / steps to reproduce the problem:
Let us first use the caffe version of this model
**Load the model**

- [ ] list of API calls:
```
curl -X PUT "http://localhost:8080/services/imageserv" -d '{
"mllib":"caffe",
"description":"image classification service",
"type":"supervised",
"parameters":{
    "input":{
    "connector":"image",
    "width":512,
    "height":512
    },
    "mllib": {
       "nclasses": 2
    }
},
"model":{
    "repository":"PATH_MODEL"
}
}'
```
- [ ] Server log output:
```
[2020-08-10 12:15:47.549] [imageserv] [info] Using pre-trained weights from PATH_MODEL/model_iter_20000.caffemodel
[2020-08-10 12:15:47.648] [caffe] [info] Ignoring source layer label_data_1_split
[2020-08-10 12:15:47.670] [caffe] [info] Ignoring source layer arm_loc_ftune_arm_loc_0_split
[2020-08-10 12:15:47.670] [caffe] [info] Ignoring source layer arm_conf_ftune_arm_conf_0_split
[2020-08-10 12:15:47.670] [caffe] [info] Ignoring source layer arm_priorbox_arm_priorbox_0_split
[2020-08-10 12:15:47.671] [caffe] [info] Ignoring source layer arm_loss
[...]

PR #770, opened 2020-08-11
fix(TRT): refinedet

fantes:
this PR fixes refinedet over TensorRT
- inputs blob id were swapped 
- detection output layer / refinedet case was not correctly saved upon serialization / deserialization
- deserialization from existing engine did not work  : bug in upstream tensorrt-oss (namely (de)serialization of normalize layer and (de)serialization of priorboxlayer), fixed upstream since juillet 2nd,  make dede depend on this fixed version


fantes:
fixes https://github.com/jolibrain/deepdetect/issues/769

PR #780, opened 2020-09-04
feat: format the code with clang-format

sileht:
The code currently mixes tab and space indenting.

Having the code style defined into a file is usually good practice, as
we can use tools to do it automatically.

This change proposes use clang-format to do it. It provides predefined
rules following LLVM/Google/Mozilla/GNU code style that can be tweaked.

I try to create a .clang-format configuration that just fix the
tab/space mix and keep the rest as-is as much I can.

Currently all cpp/hpp/h are formatted except those in src/ext.

The code formatting will be checked Jenkins.
It also can be run with: `cmake .. && make clang-format`

cmake also install automatically a pre-commit hook to format the code.
It can be disabled with `INSTALL_GIT_HOOKS=OFF`

llvm provide some basic integration for editor:

https://clang.llvm.org/docs/ClangFormat.html#vim-integration
https://clang.llvm.org/docs/ClangFormat.html#emacs-integration

But you can also use other tools like:

https://github.com/SavchenkoValeriy/emacs-clang-format-plus

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,135 @@
+---
+Language:        Cpp
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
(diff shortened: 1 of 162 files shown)


sileht:
I don't care about the code style rules themself, I just don't want to think about them.

BynaryCobweb:
Add this in `.clang-format`:
```
AllowShortFunctionsOnASingleLine: SFS_None
```
to avoid formatting one-line functions.
Other options are listed on [this page](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

PR #783, opened 2020-09-04
Fix build with syslog

BynaryCobweb:
See https://github.com/jolibrain/caffe/pull/74

### DIFF
--- a/src/apistrategy.h
+++ b/src/apistrategy.h
@@ -22,15 +22,13 @@
 #ifndef APISTRATEGY_H
 #define APISTRATEGY_H
 
+#ifdef USE_DD_SYSLOG
+#define SPDLOG_ENABLE_SYSLOG
+#endif
+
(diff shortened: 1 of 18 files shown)


BynaryCobweb:
As setting SPDLOG_ENABLE_SYSLOG everytime we include spdlog.h, it would be great to create a "wrapper file" like this:
```cpp
// dd_spdlog.h

#ifdef USE_DD_SYSLOG
#define SPDLOG_ENABLE_SYSLOG
#endif

#include <spdlog/spdlog.h>

#if SPDLOG_VER_MAJOR > 0
#ifdef USE_DD_SYSLOG
#include <spdlog/sinks/syslog_sinks.h>
#else
#include <spdlog/sinks/stdout_sinks.h>
#endif
#endif
```
I can include that in this PR on monday or so.

sileht (approved):

PR #795, opened 2020-09-11
feat(torch) : Native torch C++ model interface + NBEATS architecture implementation for time-series 

fantes:
https://github.com/philipperemy/n-beats/

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -41,6 +41,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 
+#include "native/native.h"
+
 using namespace torch;
 
 using google::protobuf::io::CodedInputStream;
@@ -125,9 +127,98 @@ namespace dd
   {
   }
 
+  void TorchModule::to(torch::Device device)
+  {
+    if (_graph)
+      _graph->to(device);
+    if (_native)
+      _native->to(device);
+    if (_traced)
+      _traced->to(device);
+  }
+
+  void TorchModule::to(torch::Dtype dtype)
+  {
+    if (_graph)
+      _graph->to(dtype);
+    if (_native)
+      _native->to(dtype);
+    if (_traced)
+      _traced->to(dtype);
+  }
+
+  void TorchModule::to(torch::Device device, torch::Dtype dtype)
+  {
+    if (_graph)
+      _graph->to(device, dtype);
+    if (_native)
+      _native->to(device, dtype);
+    if (_traced)
+      _traced->to(device, dtype);
+  }
+
+  template <class TInputConnectorStrategy>
+  void TorchModule::post_transform(const std::string tmpl,
+                                   const APIData &template_params,
+                                   const TInputConnectorStrategy &inputc,
+                                   const TorchModel &tmodel,
+                                   const torch::Device &device)
+  {
+    this->_native = std::shared_ptr<NativeModule>(
+        NativeFactory::from_template<TInputConnectorStrategy>(
+            tmpl, template_params, inputc));
+
+    if (_native)
+      if (!tmodel._native.empty())
+        torch::load(_native, tmodel._native, device);
+
+    if (_graph)
+      {
+        std::vector<long int> dims = inputc._dataset.datasize(0);
+        dims.insert(dims.begin(), 1); // dummy batch size
+        _graph->finalize(dims);
+        // reload params after finalize
+        if (!tmodel._traced.empty())
+          torch::load(_graph, tmodel._traced, _device);
+      }
+    to(_device);
+  }
+
+  template <class TInputConnectorStrategy>
+  void TorchModule::post_transform_train(const std::string tmpl,
+                                         const APIData &template_params,
+                                         const TInputConnectorStrategy &inputc,
+                                         const TorchModel &tmodel,
+                                         const torch::Device &device)
+  {
+    post_transform(tmpl, template_params, inputc, tmodel, device);
+  }
+
+  template <class TInputConnectorStrategy>
+  void TorchModule::post_transform_predict(
+      const std::string tmpl, const APIData &template_params,
+      const TInputConnectorStrategy &inputc, const TorchModel &tmodel,
+      const torch::Device &device, const APIData &ad)
+  {
+    post_transform(tmpl, template_params, inputc, tmodel, device);
+    if (_graph)
+      {
+        if (ad.getobj("parameters").getobj("input").has("continuation")
+            && ad.getobj("parameters")
+                   .getobj("input")
+                   .get("continuation")
+                   .get<bool>())
+          _graph->lstm_continues(true);
+        else
+          _graph->lstm_continues(false);
+      }
+  }
+
   c10::IValue TorchModule::forward(std::vector<c10::IValue> source)
   {
-    if (_native) // native modules take only one tensor as input for now
+    if (_graph) // native modules take only one tensor as input for now
+      return _graph->forward(to_tensor_safe(source[0]));
+    if (_native)
       return _native->forward(to_tensor_safe(source[0]));
     if (_traced)
       {
@@ -179,6 +270,8 @@ namespace dd
 
   std::vector<Tensor> TorchModule::parameters()
   {
+    if (_graph)
+      return _graph->parameters();
     if (_native)
       return _native->parameters();
     std::vector<Tensor> params;
@@ -199,8 +292,10 @@ namespace dd
       _traced->save(model._repo + "/checkpoint-" + name + ".pt");
     if (_classif)
       torch::save(_classif, model._repo + "/checkpoint-" + name + ".ptw");
+    if (_graph)
+      torch::save(_graph, model._repo + "/checkpoint-" + name + ".pt");
     if (_native)
-      torch::save(_native, model._repo + "/checkpoint-" + name + ".pt");
+      torch::save(_native, model._repo + "/checkpoint-" + name + ".npt");
   }
 
   void TorchModule::load(TorchModel &model)
@@ -212,37 +307,48 @@ namespace dd
       torch::load(_classif, model._weights, _device);
     if (!model._proto.empty())
       {
-        _native = std::make_shared<CaffeToTorch>(model._proto);
+        _graph = std::make_shared<CaffeToTorch>(model._proto);
         if (!model._traced.empty())
-          torch::load(_native, model._traced, _device);
+          torch::load(_graph, model._traced, _device);
+      }
+    if (!model._native.empty())
+      {
+        std::shared_ptr<NativeModule> m;
+        torch::load(m, model._native);
+        _native = m;
       }
   }
 
   void TorchModule::eval()
   {
-    if (_native)
-      _native->eval();
+    if (_graph)
+      _graph->eval();
     if (_traced)
       _traced->eval();
     if (_classif)
       _classif->eval();
+    if (_native)
+      _native->eval();
   }
 
   void TorchModule::train()
   {
-    if (_native)
-      _native->train();
+    if (_graph)
+      _graph->train();
     if (_traced)
       _traced->train();
     if (_classif)
       _classif->train();
+    if (_native)
+      _native->train();
   }
 
   void TorchModule::free()
   {
-    _native = nullptr;
+    _graph = nullptr;
     _traced = nullptr;
     _classif = nullptr;
+    _native = nullptr;
   }
 
   // ======= TORCHLIB
@@ -277,6 +383,7 @@ namespace dd
     _classification = tl._classification;
     _timeserie = tl._timeserie;
     _loss = tl._loss;
+    _template_params = tl._template_params;
   }
 
   template <class TInputConnectorStrategy, class TOutputConnectorStrategy,
@@ -346,11 +453,14 @@ namespace dd
         this->_mlmodel._proto = dest_net;
       }
 
-    // Create the model
-    if (this->_mlmodel._traced.empty() && this->_mlmodel._proto.empty())
-      throw MLLibInternalException(
-          "Use of libtorch backend needs either traced net or protofile");
+    bool unsupported_model_configuration
+        = this->_mlmodel._traced.empty() && this->_mlmodel._proto.empty()
+          && !NativeFactory::valid_template_def(_template);
 
+    if (unsupported_model_configuration)
+      throw MLLibInternalException("Use of libtorch backend needs either: "
+                                   "traced net, protofile or native template");
+    // Create the model
     this->_inputc._input_format = "bert";
     if (_template == "bert")
       {
@@ -382,7 +492,8 @@ namespace dd
         this->_inputc._input_format = "gpt2";
         _seq_training = true;
       }
-    else if (_template.find("recurrent") != std::string::npos)
+    else if (_template.find("recurrent") != std::string::npos
+             || NativeFactory::is_timeserie(_template))
       {
         _timeserie = true;
       }
@@ -409,6 +520,9 @@ namespace dd
     _best_metrics = { "map", "meaniou",  "mlacc", "delta_score_0.1", "bacc",
                       "f1",  "net_meas", "acc",   "L1_mean_error",   "eucll" };
     _best_metric_value = std::numeric_limits<double>::infinity();
+
+    if (lib_ad.has("template_params"))
+      _template_params = lib_ad;
   }
 
   template <class TInputConnectorStrategy, class TOutputConnectorStrategy,
@@ -536,17 +650,8 @@ namespace dd
     try
       {
         inputc.transform(ad);
-        if (_module._native)
-          {
-            auto batchoptional = inputc._dataset.get_batch({ 1 });
-            TorchBatch batch = batchoptional.value();
-            torch::Tensor td = batch.data[0];
-            _module._native->finalize(td.sizes());
-            // reload params after finalize
-            if (!this->_mlmodel._traced.empty())
-              torch::load(_module._native, this->_mlmodel._traced, _device);
-            _module._native->to(_device);
-          }
+        _module.post_transform_train<TInputConnectorStrategy>(
+            _template, _template_params, inputc, this->_mlmodel, _device);
       }
     catch (...)
       {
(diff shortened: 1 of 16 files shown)


sileht (approved):

beniz:
@fantes Expecting to see more native models being part of the C++ code, what would be your thoughts on having a `torch/native/templates` directory to host them all ?

fantes:
> @fantes Expecting to see more native models being part of the C++ codebase, what would be your thoughts on having a `torch/native/templates` directory to host them all ?
in my mind, the torch/native directory was the place to put them all, because torch/native itself will contains nothing else than the .h  (or we could put native.h in torch/ and native templates in torch/native)

beniz:
> But if your prefer a template dir, I am ok

Let's say templates are part of a type of distribution (could be modulated depending on what templates are needed) whereas native interface is more generic and part of the framework I imagine.

src/backends/torch/torchlib.cc
```diff
@@ -126,7 +128,9 @@ namespace dd
 
   c10::IValue TorchModule::forward(std::vector<c10::IValue> source)
   {
-    if (_native) // native modules take only one tensor as input for now
+    if (_graph) // native modules take only one tensor as input for now
+      return _graph->forward(to_tensor_safe(source[0]));
+    if (_native)
```
beniz, line 133: Can we have both, or do we want an `else` here ?

src/backends/torch/torchlib.cc
```diff
@@ -178,6 +182,8 @@ namespace dd
 
   std::vector<Tensor> TorchModule::parameters()
   {
+    if (_graph)
+      return _graph->parameters();
     if (_native)
```
beniz, line 187: Is an `else` needed ?

src/backends/torch/torchlib.cc
```diff
@@ -211,37 +219,48 @@ namespace dd
...
   {
-    if (_native)
-      _native->eval();
+    if (_graph)
+      _graph->eval();
     if (_traced)
       _traced->eval();
     if (_classif)
       _classif->eval();
+    if (_native)
```
beniz, line 242: Pointing the possible need of `else` statements here, may be inappropriate though, unsure.

src/backends/torch/torchlib.cc
```diff
@@ -346,9 +366,11 @@ namespace dd
       }
 
     // Create the model
-    if (this->_mlmodel._traced.empty() && this->_mlmodel._proto.empty())
-      throw MLLibInternalException(
-          "Use of libtorch backend needs either traced net or protofile");
+    if (this->_mlmodel._traced.empty() && this->_mlmodel._proto.empty()
```
beniz, line 369: I suggest we wrap this long condition into a function that should be easier to maintain in the future. Also, we may want to put a nice comment on why every condition is there. We've seen this growing uncontrolled before in the Caffe lib API !

src/backends/torch/torchlib.cc
```diff
@@ -534,15 +561,30 @@ namespace dd
...
+          {
+            CSVTSTorchInputFileConn *ctip
+                = reinterpret_cast<CSVTSTorchInputFileConn *>(&inputc);
+            _module._native = std::make_shared<NBeats>(*ctip, _nbeats_stacks);
+            // load params if any after instanciation
+            if (!this->_mlmodel._native.empty())
+              torch::load(_module._native, this->_mlmodel._native, _device);
+            _module._native->to(_device);
+          }
+        if (_module._graph)
```
beniz, line 575: `else` applicable ?

src/backends/torch/torchlib.cc
```diff
@@ -568,6 +610,10 @@ namespace dd
     // logging parameters
     int64_t log_batch_period = 20;
 
+    if (ad_mllib.has("nbeats_stacks"))
```
beniz, line 613: Maybe we'd like to have the torch native C++ template models to be accomodated in a function, so that when we add more, we have placeholders ready. So that we can keep the generic train calls as a list of high level function calls.

src/backends/torch/torchlib.cc
```diff
@@ -725,12 +773,32 @@ namespace dd
               }
             else if (_timeserie)
               {
-                if (_loss.empty() || _loss == "L1" || _loss == "l1")
-                  loss = torch::l1_loss(y_pred, y);
-                else if (_loss == "L2" || _loss == "l2" || _loss == "eucl")
-                  loss = torch::mse_loss(y_pred, y);
+                if (_template.find("nbeats") != std::string::npos)
```
beniz, line 776: Can we wrap this up somewhere dedicated to the `nbeats` template ? Like `nbeat_native_template_loss_selection()` that would follow some generic signature ? This would allows simplifying the adding of more native templates.

src/backends/torch/torchlib.cc
```diff
@@ -898,21 +968,23 @@ namespace dd
 
     bool lstm_continuation = false;
     TInputConnectorStrategy inputc(this->_inputc);
+    if (_template.find("nbeats") != std::string::npos)
```
beniz, line 971: I'm thinking having a `nbeats_native_template_inputconn_config()` function somewhere. Possibly into the template actually, i.e. outside of `torchlib`.

beniz:
Aside from my small comments, my main remark would be that I believe `torchlib` should be independent from torch native templates we may provide. This meaning that we may want any configuration code touching a given template to be part of the native template code itself. 
Some higher level function call could then take place into `torchlib`. Factory/virtual/policy-based should do it, I'm keen on discussing lower level details actually.

src/backends/torch/torchlib.cc
sileht: Instead of documentation I suggest using explicit variable name like:
```
bool  unsupported_model_configuration = this->_mlmodel._traced.empty() && this->_mlmodel._proto.empty()
```

src/backends/torch/torchlib.cc
fantes: I followed what was previously done with traced net that can followed by a weight net (fully connected) ; This allows for free to chain nets, if you really know what you are doing (typically few things will need to be rewritten around input connector -> first net) 
For know we do not have both, and at some time in the future, this should be superseded by the graph abstraction layer, ie a torchModule will be a node , that can be of one type native / traced / or (sub) graph.
Morality : if you think having a else makes things more clear, I am ok ;)

src/backends/torch/torchlib.cc
fantes: no else because of return :)    this contradicts previsous comment, if we want to chain net for free, we need to cat parameters and not only send first one

src/backends/torch/torchlib.cc
fantes: see first comment : this allows to chain nets for free, unused ATM

src/backends/torch/torchlib.cc
fantes: same as before, one could want to chain a native and a graph, but with current code they take same input

src/backends/torch/torchlib.cc
fantes: agreed, but I am not sure how to do it ATM, do you have an idea ?

src/backends/torch/torchlib.cc
fantes: I am doing this directly in nbeats class, ie there will be here _module._native->loss() here, what do you think ?

src/backends/torch/torchlib.cc
fantes: I am puttinh this as virtual into module._native/ native_net / nbeats , native_net class is there exactlty for doing so

fantes:
> Aside from my small comments, my main remark would be that I believe `torchlib` should be independent from torch native templates we may provide. This meaning that we may want any configuration code touching a given template to be part of the native template code itself.
> Some higher level function call could then take place into `torchlib`. Factory/virtual/policy-based should do it, I'm keen on discussing lower level details actually.

native_net class was put there for doing so: to declare virtual stuff and hide everything, I prefer the inheritance way than policy based way, in case we have a lot of native stuff, and native API should be relatively stable

src/backends/torch/torchlib.cc
fantes: I will propose a mix of both , let me know after my next push

src/backends/torch/torchlib.cc
fantes: ok, I got an idea, see my upcoming push

src/backends/torch/torchlib.cc
```diff
@@ -754,7 +807,9 @@ namespace dd
                   }
                 try
                   {
+                    auto p = _module.parameters();
```
BynaryCobweb, line 810: p is unused ?

fantes: remains from debug code, removed

src/backends/torch/torchlib.cc
```diff
@@ -534,15 +561,30 @@ namespace dd
...
+            _module._graph->finalize(td.sizes());
             // reload params after finalize
             if (!this->_mlmodel._traced.empty())
-              torch::load(_module._native, this->_mlmodel._traced, _device);
+              torch::load(_module._graph, this->_mlmodel._traced, _device);
+            _module._graph->to(_device);
+          }
+        if (_module._native)
+          {
             _module._native->to(_device);
```
BynaryCobweb, line 588: `module._native->to(_device)` called a second time (first time l.572)

It could be great to centralize calls of the `to()` method in `TorchModule`, unless we need to have some models (graph, native, traced) on cpu and some other on gpu. What do you think?

src/backends/torch/torchlib.cc
```diff
@@ -1044,6 +1110,9 @@ namespace dd
           }
         else if (_timeserie)
           {
+            if (_module._native != nullptr)
+              output = _module._native->cleanup_output(output);
+
```
BynaryCobweb, line 1115: Is the call to `cleanup_output` specific to timeseries or can we put it outside of the "if"?

fantes: done

src/backends/torch/torchlib.cc
fantes: ok, almost done :)

tests/ut-torchapi.cc
```diff
@@ -44,6 +44,11 @@ static std::string bert_train_repo
 static std::string bert_train_data
     = "../examples/torch/bert_training_torch_140_transformers_251/data/";
 
+static std::string sinus = "../examples/all/sinus/";
+
+static std::string iterations_nbeats_cpu = "100";
```
beniz, line 49: FYI nbeats tests are not built on CPU because they are enclosed within the scope of a `#if !defined(CPU_ONLY)` that was originally put in place to remove BERT training tests from clogging CPU builds. We may want to add nbeats training to CPU unit tests, what do you think ?

fantes: ok, done

src/backends/torch/torchlib.cc
```diff
@@ -211,37 +249,48 @@ namespace dd
...
-    if (_native)
-      _native->train();
+    if (_graph)
+      _graph->train();
     if (_traced)
       _traced->train();
     if (_classif)
       _classif->train();
+    if (_native)
+      _native->eval();
```
beniz, line 285: Should be `train()` ?

fantes: it should. nice catch!

src/backends/torch/torchlib.cc
```diff
@@ -534,17 +591,25 @@ namespace dd
     try
       {
         inputc.transform(ad);
+
+        _module._native = std::shared_ptr<NativeModule>(
+            NativeFactory::from_template(_template, _template_params, inputc));
```
beniz, line 596: In Caffe backend, we try to instantiate the network template at service init. Do we want to do the same here ? As a side question, probably the template could be instantiated before the input are even transformed ?

src/backends/torch/torchlib.cc
beniz: was it even training ? ;)

src/backends/torch/torchlib.cc
fantes: unfortunately, instanciating nbeats needs the forecast and backcast length to be known, and they are "timesteps" input connector parameter, which can be given only at train/predict time

fantes: this is because for lstm, timestep is only used at timeserie length and lstm simply unroll over this data dimension, but for nbeats this has to be statically fixed

src/backends/torch/torchlib.cc
fantes: mmmm

fantes: train() is passed to all module and they do what they want with it, ie switch on/off dropout batchnorm and so on, most models do not do anything with it

src/backends/torch/torchlib.cc
```diff
@@ -534,17 +591,25 @@ namespace dd
...
             // reload params after finalize
             if (!this->_mlmodel._traced.empty())
-              torch::load(_module._native, this->_mlmodel._traced, _device);
-            _module._native->to(_device);
+              torch::load(_module._graph, this->_mlmodel._traced, _device);
           }
+        _module.to(_device);
       }
     catch (...)
       {
```
BynaryCobweb, line 615: Do we still need the `try { ... } catch (...) { throw; }` ? It seems cumbersome and could be removed with no consequences

fantes: @beniz this is inspired from the try catch around inputc.transform in caffelib https://github.com/jolibrain/deepdetect/blob/5e982df6b71b55e26715f71f304d534daa59dc7d/src/backends/caffe/caffelib.cc#L1352 , do you have any comments ?

beniz: `transform` throws by design, if this is what you ask.

src/backends/torch/torchlib.cc
```diff
@@ -536,17 +593,25 @@ namespace dd
...
         inputc.transform(ad);
+
+        _module._native = std::shared_ptr<NativeModule>(
+            NativeFactory::from_template(_template, _template_params, inputc));
+
         if (_module._native)
+          if (!this->_mlmodel._native.empty())
+            torch::load(_module._native, this->_mlmodel._native, _device);
+
+        if (_module._graph)
```
beniz, line 604: I would recommend that anything that depends on values from within `_module` do actually happen in the `TorchModule` class. This is to avoid spreading the code and complexifying torchlib itself.

src/backends/torch/torchlib.cc
```diff
@@ -900,21 +977,22 @@ namespace dd
...
+    if (_module._native != nullptr)
+      _module._native->update_input_connector(inputc);
+
     TOutputConnectorStrategy outputc(this->_outputc);
-    ;
     try
       {
         inputc.transform(ad);
-        if (_module._native)
+        if (_module._graph)
```
beniz, line 987: Similar to my previous remark, I'd see this block of code going first into `TorchModule` with a function call replacing the block.

Next I'd envison that anything that deals with `graph` internals remains within `graph`, typically the `lstm_continues` stuff, i.e. as far as possible by design.

src/backends/torch/torchlib.cc
```diff
@@ -212,37 +250,48 @@ namespace dd
...
       {
-        _native = std::make_shared<CaffeToTorch>(model._proto);
+        _graph = std::make_shared<CaffeToTorch>(model._proto);
         if (!model._traced.empty())
-          torch::load(_native, model._traced, _device);
+          torch::load(_graph, model._traced, _device);
+      }
+    if (!model._native.empty())
+      {
+        std::shared_ptr<NativeModule> m;
```
beniz, line 259: My understanding is that at this stage, the native template hasn't been set. Before loading the model, we may want to put down a call to `NativeFactory::from_template` ?

src/backends/torch/torchlib.cc
beniz: OK, maybe we can do it within `graph` ?

src/backends/torch/torchlib.cc
fantes: ok, done

src/backends/torch/torchlib.cc
fantes: ok, done

src/backends/torch/torchlib.cc
BynaryCobweb: @beniz What do you mean by that? Is the `try ... catch` here as a remainder for exceptions thrown by `transform`?

src/backends/torch/torchlib.cc
fantes: in case of native module/template, torchmodule::load should be called after factory::from_template, and it seems to me that it is the case in the code, i am not sure to understand your point

beniz: When `TorchModule::load` is called from `init_mllib`, it doesn't seem that at that stage `from_template` has been called. Actually since one call is mandatory before the other, we may want to regroup them into a function ?

fantes: in case of nbeats, it is not possible to instanciate model before having exact data size (length of timeseries), ie before the transform, so `init_mllib` cannot instanciate using `NativeFactory::from_template` . init_mllib loads a native net if available in repo, but it will be replaced by `NativeFactory::from_template` followed by another `TorchModule::load` at train or test time .
All we can do is remove this first model loading during init_mllib

src/backends/torch/torchlib.cc
```diff
@@ -125,9 +127,98 @@ namespace dd
...
+    if (_graph)
+      _graph->to(device, dtype);
+    if (_native)
+      _native->to(device, dtype);
+    if (_traced)
+      _traced->to(device, dtype);
+  }
+
+  template <class TInputConnectorStrategy>
+  void TorchModule::post_transform(const std::string tmpl,
```
beniz, line 161: we may call this function `load` for clarity ?

fantes: in my mind it was more like a hook (that loads in case of lstm and nbeats) or does nothing (in case of traced  becasue already loaded) . We should in the future clarify all this when unifying TorchModule as a graph

BynaryCobweb (approved):

src/backends/torch/torchlib.cc
beniz: OK, great, to be clarified in terms of API in later updates!

beniz (approved):

PR #805, opened 2020-09-14
feat(imginputfile): histogram equalization of input image

### LINKED ISSUE #778: Image histogram equalization by the image input connector
Having OpenCV's [histogram equalization](https://docs.opencv.org/2.4/doc/tutorials/imgproc/histograms/histogram_equalization/histogram_equalization.html) directly executed by the input connector is useful in some cases for both training and inference.

We suggest a `eqhist` boolean parameter to be added to the `input` JSON object on both training and predict calls.

I guess this could go after the `rgb` check here: https://github.com/jolibrain/deepdetect/blob/master/src/imginputfileconn.h#L104

The function to apply from OpenCV should be: 
```
/// Apply Histogram Equalization
cv::equalizeHist( src, dst );
```
Beware that equalization is per channel, so the call above may only work when the image is single channel B&W, which is our motivating use case actually. For RGB images, the equalization may need to be applied per channel.

sileht:
This does a histogram equalization of bw & colored image.

Closes #778


clients/python/dd_bench.py
```diff
@@ -37,6 +37,7 @@
 parser.add_argument('--img-width',help='image width',type=int,default=224)
 parser.add_argument('--img-height',help='image height',type=int,default=224)
 parser.add_argument('--bw',help='whether images are bw',action='store_true')
+parser.add_argument('--histogram-equalization',help='whether we apply an histogram equalization to images', action='store_true')
```
beniz, line 40: Suggestion, maybe abbreviated as `--eqhist`, unless it's considered bad practice!

beniz:
Other than maybe abbreviate the API option, PR is fine by me!

fantes (approved):

clients/python/dd_bench.py
sileht: For API, self explicit name if often better.
For cli tools, it always good to have the explicit option (for usage of the cli in script) and a short alias (for usage in interactive terminal)

Issue #811, opened 2020-09-15
DeepDetect builds with `TENSORRT_OSS` do require CUDA 10.2


beniz:
Otherwise the build error is as follows: 
```
deepdetect/build/tensorrt-oss/src/tensorrt-oss/samples/common/sampleDevice.h:189:56: error: ‘cudaStreamCaptureModeGlobal’ was not declared in this scope
         cudaCheck(cudaStreamBeginCapture(stream.get(), cudaStreamCaptureModeGlobal));
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
deepdetect/build/tensorrt-oss/src/tensorrt-oss/samples/common/sampleDevice.h:189:56: note: suggested alternative: ‘cudaStreamCaptureStatusNone’
         cudaCheck(cudaStreamBeginCapture(stream.get(), cudaStreamCaptureModeGlobal));
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                        cudaStreamCaptureStatusNone
```

FTR this also appears here: https://github.com/triton-inference-server/server/issues/635

Using CUDA 10.2 fixes the issue. A check for CUDA 10.2 is probably needed in the `CMakeFile` to catch the issue much earlier.

YaYaB:
According to the issue on triton-infenrence-server it should work with cuda 10.1 no?

beniz:
dk, 10.1 is deprecated for us.

PR #813, opened 2020-09-16
fix: build with -DUSE_TENSORRT_OSS=ON

beniz:
Sets Jenkins to build with TENSORRT_OSS as default since it's now the
most recent and advanced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -889,9 +889,9 @@ if (USE_TENSORRT)
 	set(TENSORRT_LIBS protobuf nvinfer nvparsers nvinfer_plugin nvonnxparser )
 
   else()
-    set(TENSORRT_LIB_DIR ${CMAKE_BINARY_DIR}/tensorrt-oss/src/tensorrt-oss-build ${TENSORRT_LIB_DIR})
+    set(TENSORRT_LIB_DIR ${CMAKE_BINARY_DIR}/tensorrt-oss/bin/ ${TENSORRT_LIB_DIR})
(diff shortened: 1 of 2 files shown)


beniz:
We still should check for proper CUDA 10.2 for builds with `-DUSE_TENSORRT_OSS=ON`, see #811

CMakeLists.txt
```diff
@@ -889,9 +889,9 @@ if (USE_TENSORRT)
 	set(TENSORRT_LIBS protobuf nvinfer nvparsers nvinfer_plugin nvonnxparser )
 
   else()
-    set(TENSORRT_LIB_DIR ${CMAKE_BINARY_DIR}/tensorrt-oss/src/tensorrt-oss-build ${TENSORRT_LIB_DIR})
+    set(TENSORRT_LIB_DIR ${CMAKE_BINARY_DIR}/tensorrt-oss/src/tensorrt-oss-build ${CMAKE_BINARY_DIR}/tensorrt-oss/bin/ ${TENSORRT_LIB_DIR})
```
fantes, line 892: I think ${CMAKE_BINARY_DIR}/tensorrt-oss/src/tensorrt-oss-build can be dropped as everything should be in ${CMAKE_BINARY_DIR}/tensorrt-oss/bin/  because of -DTRT_OUT_DIR=${CMAKE_BINARY_DIR}/tensorrt-oss/bin set some lines below

fantes (approved):

CMakeLists.txt
beniz: Good intuition! the build goes through when removing this path, I've updated the commit.

PR #817, opened 2020-09-17
Pull request for pragma

sileht:


sileht:
Only 1 kind of warning remaining due to a macro conflict between xgboost/dmlc logger and caffe logger, to be able to enable "-Werror"

beniz:
My only remark I guess: on files that are under heavy development, e.g. anything related to the C++ torch interface with DD, we may want to leave the warnings on `unused-parameters`, signs and types ?

sileht:
The point of such cleaning is to be able to set `-Werror`. Setting this flag doesn't mean you have to fix all warnings all the times. 
It's totally OK, to locally ignore warning or use `__attribute__((unused))`.

The usage of `-Werror` just means you will have all ignored rules explicitly written in the code, instead of having to looks at compliation logs (that nobody reads when a lot of warning are printed).

PR #825, opened 2020-09-23
Pull request for Werror

sileht:
## chore(build): hide maybe-uninitialized for caffe.pb.h

Depending on the version of protoc the warning differ.
So we need to filter all of them.

## chore(build): protobuf::io::CodedInputStream::SetTotalBytesLimit deprecation

second parameter is ignored and deprecated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ if (${PB_GENERATOR_SYSTEM})
 endif()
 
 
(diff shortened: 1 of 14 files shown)


fantes (approved):
Can we make a card/task/ticket for tensort deprecation on createInferBuilder (you can assign it to me unless you want to handle it)

sileht:
I created a card.

PR #827, opened 2020-09-23
Graph lstm autoencoder

fantes:
this PR adds support for lstm autoencoders using graph/torch API

API follows classical LSTM call:
service creation 
```
curl -X PUT "http://localhost:8800/services/autoencoder" -d '{
        "mllib":"torch",
        "description":"lstm_autoencoder",
         "type":"supervised",
         "model": {
             "repository":"/path/to/repo"
         },
        "parameters" : {
             "input": { 
                "connector":"csvts",
                "label":["output"] 
             },
             "mllib" : {
                "template" : "recurrent",
                 "layers":["L10","L10","T","L10"],
                 "loss":"L1"
             }
       }
}'
```
`T` means Tile, original caffe layer name that repeats one blob along a dimension. Here last hidden value of the encoder is  repeated / tiled to serve as input for all decoder inputs. 

exemple training:
```
curl -X POST "http://localhost:8800/train" -d '{
         "service":"autoencoder", 
          "async":true,
          "parameters" : {
              "input" : {
                  "shuffle":true,"                                                                                                                                        
                   "separator":",",
                    "scale":true,
                     "timesteps":200,
                    "label":["output"]  
                },
              "mllib" : { 
                  "gpu":true,
                  "solver" : {
                      "iterations":100, 
                      "test_interval":10,
                      "base_lr":0.01,
                      "snapshot":10,
                      "test_initialization":false
                   },
                   "net": { 
                      "batch_size":100,
                      "test_batch_size":10
                    }
               },
              "output" : {
                  "measure":["L1","L2"] 
               }
          },
         "data":["/path/to/train/data", "/path/to/test/data"]
}                                                                                                                                 
```

### DIFF
--- a/src/basegraph.cc
+++ b/src/basegraph.cc
@@ -257,6 +257,13 @@ namespace dd
           outputdim.push_back(_graph[inputs[0]].dim[i]);
         outputdim.push_back(_graph[producer].num_output);
       }
+    else if (_graph[producer].type == "Tile")
+      {
+        // BIG FAT WARNING : this is a hack for tile in autoencoders only
+        // because exposed hidden ouputs are computed as of full size
+        outputdim.push_back(_graph[inputs[0]].dim[1]); // timesteps
+        outputdim.push_back(_graph[inputs[0]].dim[2]); // hidden_size
+      }
     return std::tie(inputdim, outputdim);
   }
 
@@ -265,6 +272,7 @@ namespace dd
     auto es = boost::in_edges(v, _graph);
     auto eit = es.first;
     BaseGraph::Vertex producer = boost::source(*eit, _graph);
+
     auto newdims = compute_dims_from_producer(producer);
     update_alloc_status(newdims, producer);
     _graph[producer].dim = std::get<1>(newdims);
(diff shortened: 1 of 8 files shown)


src/basegraph.cc
```diff
@@ -257,6 +257,13 @@ namespace dd
           outputdim.push_back(_graph[inputs[0]].dim[i]);
         outputdim.push_back(_graph[producer].num_output);
       }
+    else if (_graph[producer].type == "Tile")
+      {
+        // BIG FAT WARNING : this is a hack for tile in autoencoders only
```
beniz, line 262: Maybe we may condition on an `_autoencoder` boolean ?

beniz:
Nice PR, exciting to see timeseries autoencoders!

src/basegraph.cc
fantes: my first aim was to implement generic Tile layer, independently from autoencoders, but it needs a few changes that I will do later (I have a ticket :)  , so I would prefer to stick with current version here, and clean up things when I will allow different sizes for outputs of a layer ( it is not complicated, but needs to be done carefully)

beniz: OK

beniz (approved):

sileht (approved):

PR #829, opened 2020-09-24
feat(stats): added service statistics mechanism

sileht:
the initial version returns:

    inference_count
    predict_success
    predict_failure
    predict_count
    avg_batch_size
    avg_predict_duration
    avg_transform_duration

### DIFF
--- a/src/mlservice.h
+++ b/src/mlservice.h
@@ -248,6 +248,7 @@ namespace dd
               ad.add("height", this->_inputc.height());
             }
         }
+      this->_stats.to(ad);
       return ad;
     }
 
@@ -283,7 +284,9 @@ namespace dd
                 static_cast<long int>(this->_mem_used_train * sizeof(float)));
       stats.add("data_mem_test",
                 static_cast<long int>(this->_mem_used_test * sizeof(float)));
-      ad.add("stats", stats);
+      ad.add("stats", stats); // FIXME(sileht): deprecated name, delete me when
+                              // platform use the new name
+      ad.add("model_stats", stats);
       ad.add("jobs", vad);
       ad.add("parameters", _init_parameters);
       ad.add("repository", this->_inputc._model_repo);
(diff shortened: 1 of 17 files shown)


src/mlservice.h
```diff
@@ -495,36 +499,29 @@ namespace dd
...
-          boost::shared_lock<boost::shared_mutex> lock(_train_mutex);
-          return this->predict(ad, out);
+          _train_mutex.unlock_shared();
+          this->_stats.predict_end(false);
+          throw;
         }
-      return 0;
+      this->_stats.predict_end(true);
+
+      _train_mutex.unlock_shared();
```
beniz, line 523: Just curious, we need to explicitely unlock the shared mutex ?

src/statistics.cc
```diff
@@ -0,0 +1,87 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with deepdetect.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <chrono>
+
+#include "apidata.h"
+#include "statistics.h"
```
beniz, line 26: The file could be named `service_statistics.h` or `service_stats.h`. Point being that within an ML software, `statistics.h` may well handle statistical computations / algorithms in a more general sense. Also, the class itself is `ServiceStatistics` so I believe it'd make even more sense to rename.

sileht: Yes, renamed.

src/mlservice.h
sileht: I just keep how you rewrote this method in your branch and it doesn't use `std::lock_guard`. I guess this is because we use `unlock_shared()` to unlock it. If we use`std::lock_guard`, it will call `unlock()`  instead and this is not want we want here.

beniz (approved):

PR #830, opened 2020-09-24
feat: tentative refactor of torch native backend interface with templates

beniz:
This PR is an example of a possible refactoring of the torch C++ interface for native models in DD. 

Advantages :
- all compile-time checks of native models regarding required functions
- no virtual functions outside of torch internals (forward, ...)

Drawbacks:
- code is made more complicated via visitors etc.. for the variant
- doesn't fully work at the moment, see below.

At the moment, the `load` & `save` of torch models are not working, for a reason that is not too clear to me (pointer / virtual internal stuff to `torch::load` typically when applied to an object from the `native_variant_type` variant type list. Thus test fails after training a native model since the loading step doesn't work. If the step is removed, since the model is already in memory after training, tests do proceed.

I believe feedback on this alternative to the current solution could be useful in order to properly design DD's torch C++ backend so that it gracefully handles the 3 types of models: graph, torchscript modules, and C++ native.

### DIFF
--- a/src/backends/torch/native/native_factory.h
+++ b/src/backends/torch/native/native_factory.h
@@ -12,15 +12,15 @@ namespace dd
   {
   public:
     template <class TInputConnectorStrategy>
-    static NativeModule *from_template(const std::string tdef,
-                                       const APIData template_params,
-                                       const TInputConnectorStrategy &inputc)
-    {
-      (void)(tdef);
-      (void)(template_params);
-      (void)(inputc);
-      return nullptr;
-    }
+      static native_variant_type from_template(std::string tdef,
+					       APIData template_params,
+					       TInputConnectorStrategy &inputc)
+   {
+     (void)tdef;
+     (void)inputc;
+     (void)template_params;
+     throw std::exception();
+   }
 
     static bool valid_template_def(std::string tdef)
     {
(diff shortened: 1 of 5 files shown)


fantes:
In my opinion , we have two problems to solve here 
- we want to store several types in a container / pointer (ie all native/template types like nbeats)
- we want to call methods from this container w/o  knowing the real type (ie the compiler/code) selects the good method depending on real type

- if all types are unrelated, the only solution is to create an union/variant of all containable types and a (single dispatch) visitor that will call correct method depending on type
- if all types can be considered as a subtype of let say an interface (ie the list of possible methods) that is exactly what does inheritance and late binding do   (ie call the method depending on the type)

fantes:
as all native/template class will share the same interface, (ie to(), forward() and so on) I am not sure to understand the advantages of  union/variant solution. One advantage I see is to have an explicit list of possible subtypes at one place (ie at the declaration of the union/variant). The main advantage of the visitor scheme is to force to put code of a visitor at same place for all dispatched types. (in that case, the visitor is only an indirection to the native/template class corresponding method, so this advantage almost disappears)

fantes:
At very least, if union/variant solution is preferred , I would suggeest to use c++17 std::variant and rewrite completely the visitors so that they do not look like visitors that need to be applied by another function

sileht:
> code is made more complicated via visitors etc.. for the variant

It's a huge drawback:
* we don't understand anymore what the code does and have to read the visitor to get it.

```
visitor_native_output vo;
vo._output = output;
mapbox::util::apply_visitor(vo,*_module._native.get());
output = vo._output;
```
vs 

`output = _module._native->cleanup_output(output);`

It increases a lot the number of lines to write just to describe a interface that class have to implement.

> all compile-time checks of native models regarding required functions

pure virtual also offer that. If the only inheritance is from the interface one that contain only virtual, it's not a big deal to have virtual.

beniz:
> At very least, if union/variant solution is preferred , I would suggeest to use c++17 std::variant and rewrite completely the visitors so that they do not look like visitors that need to be applied by another function

We can't do this now because not all target platforms support c++17, thus at the moment the candidate to replace mapbox/variant if needed is https://github.com/mpark/variant .

sileht:
> > At very least, if union/variant solution is preferred , I would suggeest to use c++17 std::variant and rewrite completely the visitors so that they do not look like visitors that need to be applied by another function
> 
> We can't do this now because not all target platforms support c++17, thus at the moment the candidate to replace mapbox/variant if needed is https://github.com/mpark/variant .

What is the lower targeted platform ?  ubuntu 18.04 ?

Anything lower always doesn't work because of rapidjson and mapbox/variant

fantes:
https://en.cppreference.com/w/cpp/utility/variant/visit

beniz:
> It increases a lot the number of lines to write just to describe a interface that class have to implement.

Agreed, it's ugly, I had to try it.

https://i0.wp.com/magazine.lecranpop.com/wp-content/uploads/2019/07/Les-Visiteurs-3-Jean-Reno-et-Christian-Clavier-se-rejouissent-1.jpg?resize=1170%2C585&ssl=1

> pure virtual also offer that. If the only inheritance is from the interface one that contains only virtual, it's not a big deal to have virtual.

Yes, the advantage of policy-based is avoiding virtual calls, something I especially like on `predict` calls since inference is in millions+ calls. *However* my understanding here is that torch::forward is actually virtual so policy-based / variant combo does not make real sense.

I did try the whole thing as an excercise though, the structure and articulation of `NativeModule`, `TorchModule` and `Graph` remains to be polished.

beniz:
I'm closing this as a bad alternative to current design.

PR #831, opened 2020-09-25
chore: ignore caffe.pb.h warning

sileht:


BynaryCobweb (approved):
No warnings when building with torch only.

PR #832, opened 2020-09-25
refactor: native as pure virtual

sileht:
This uses pure virtual and virtual final to ensure
some methods must be implemented and some can't.

And when you need to use generic/common code it's explicit not magic.


fantes (approved):
PR 853 should make .to methods superfluous and remove them, maybe you want to merge 853 before and remove stuff here ?

sileht:
Include in #853

PR #834, opened 2020-09-25
feat(dede): Training for image classification with torch

BynaryCobweb:
### To be discussed before merge

- [x] Is using a test dataset in unit tests a good idea? It could check that the model is actually learning, and that the test process works, but it introduces a small part of random in unit tests, which may not be a good idea?
- [x] Need to add the .tar archive required for tests

# Training for image classification with torch

Add support of training and test dataset to `TorchImgInputFileConn`, thus allowing to train and finetune image classification models with libtorch backend.
Add a test for training on a simple classification task.

An example of API call for image training could be (do not call with too many data):
```
curl -X PUT "http://10.10.77.61:"$1"/services/img_serv" -d '{
  "description": "image",
  "mllib": "torch",
  "type": "supervised",
  "model": {
    "repository": "/opt/platform/models/training/louisj/private/resnet50_training_torch",
    "create_repository": false
  },
  "parameters": {
    "input": {
      "connector": "image",
      "width": 224,
      "height": 224,
      "db":true
    },
    "mllib":{
       "nclasses":2,
       "finetuning":true,
       "gpu":true
    }
  }
 }'

curl -X POST "http://10.10.77.61:"$1"/train" -d '{
  "service": "img_serv",
  "async": true,
  "parameters": {
    "mllib":{
       "solver":{
          "iterations":100,
          "base_lr":1e-5,
          "iter_size":4,
          "solver_type":"ADAM",
          "test_interval":1000
       },
       "net":{
          "batch_size":4
       },
       "nclasses":2,
       "resume":false
    },
    "input":{
       "db":true,
       "shuffle":true,
       "test_split":0.2
    },
    "output":{
       "measure":[
          "f1",
          "acc",
          "mcll",
          "acc-5"
       ]
    }
  },
  "data": [
    "/opt/platform/examples/dogs_cats/train/"
  ]
}'
```


beniz:
> * Is using a test dataset in unit tests a good idea? It could check that the model is actually learning, and that the test process works, but it introduces a small part of random in unit tests, which may not be a good idea?

Yes, if I understand correctly, you can look at `tests/ut-caffeapi.cc` for how we handle training tests and validate test based on a lower bound on accuracy.

> * [ ]  Test split is not supported, should I include it in this PR?

Yes it should, usually it's a simple shuffle + split of the list of images in the dataset.

src/backends/torch/torchinputconns.cc
```diff
@@ -155,9 +157,8 @@ namespace dd
     _indices.push_back(index);
   }
 
-  void TorchDataset::write_tensors_to_db(
-      std::vector<at::Tensor> data,
-      __attribute__((unused)) std::vector<at::Tensor> target)
+  void TorchDataset::write_tensors_to_db(std::vector<at::Tensor> data,
+                                         std::vector<at::Tensor> target)
```
beniz, line 161: shall we `const` + ref these vectors ?

src/backends/torch/torchinputconns.cc
```diff
@@ -369,6 +370,232 @@ namespace dd
     return new_dataset;
   }
 
+  // ===== ImgTorchInputFileConn
+
+  void ImgTorchInputFileConn::read_image_folder(
```
beniz, line 375: Is this the same function we use when feeding the Caffe backend ? If yes, shall we move it up to the top image connector class ?

src/backends/torch/torchinputconns.cc
```diff
@@ -369,6 +370,232 @@ namespace dd
...
+                _dataset.finalize_db();
+                bool has_test_data = _test_dataset._dbData != nullptr;
+                _test_dataset.finalize_db();
+
+                if (_test_split != 0.0 && !has_test_data)
+                  build_test_datadb_from_full_datadb(_test_split);
+              }
+          }
+
+        if (!_db)
```
beniz, line 497: else ?

src/backends/torch/torchinputconns.cc
```diff
@@ -369,6 +370,232 @@ namespace dd
...
+      }
+  }
+
+  at::Tensor ImgTorchInputFileConn::image_to_tensor(const cv::Mat &bgr)
+  {
+    std::vector<int64_t> sizes{ _height, _width, 3 };
+    at::TensorOptions options(at::ScalarType::Byte);
+
+    at::Tensor imgt = torch::from_blob(bgr.data, at::IntList(sizes), options);
+    imgt = imgt.toType(at::kFloat).permute({ 2, 0, 1 });
```
beniz, line 574: Is this related to BGR vs RGB ? If yes, we have API options for this, let's review how it should be handled.

src/backends/torch/torchinputconns.cc
BynaryCobweb: There is already a function in DDImg that reads the list of files, but we can't use it yet because it also loads all the image data (and we do not want that if we are using a db)
As modifying DDImg can have unexpected consequences, it would be better to do it in another PR, what do you think?

beniz: OK. Loading dataset into db from list of files was a user request, it's useful in some settings. It doesn't load in memory though, unless it's using the ImageInputLayer that is internal to caffe, FTR.

src/backends/torch/torchinputconns.cc
BynaryCobweb: Done

src/backends/torch/torchinputconns.cc
BynaryCobweb: Outdated

src/backends/torch/torchinputconns.cc
BynaryCobweb: > > * Is using a test dataset in unit tests a good idea? It could check that the model is actually learning, and that the test process works, but it introduces a small part of random in unit tests, which may not be a good idea?
> 
> Yes, if I understand correctly, you can look at `tests/ut-caffeapi.cc` for how we handle training tests and validate test based on a lower bound on accuracy.

The test checks that the accuracy is greater than 0.6. This lower bound is low enough for the test not to fail randomly, but greater than 0.5 so that if the model output pure random, we can detect it. I think it's a good compromise.
> 
> > *   Test split is not supported, should I include it in this PR?
> 
> Yes it should, usually it's a simple shuffle + split of the list of images in the dataset.

Done

BynaryCobweb:
> > * Is using a test dataset in unit tests a good idea? It could check that the model is actually learning, and that the test process works, but it introduces a small part of random in unit tests, which may not be a good idea?
> 
> Yes, if I understand correctly, you can look at `tests/ut-caffeapi.cc` for how we handle training tests and validate test based on a lower bound on accuracy.

The test checks that the accuracy is greater than 0.6. This lower bound is low enough for the test not to fail randomly, but greater than 0.5 so that if the model output pure random, we can detect it. I think it's a good compromise.
> 
> > *   Test split is not supported, should I include it in this PR?
> 
> Yes it should, usually it's a simple shuffle + split of the list of images in the dataset.

Done

src/backends/torch/torchinputconns.cc
BynaryCobweb: I used `DDImg::read_file` to read the image, it applies some preprocessing (scaling, cropping...). Is the RGB vs BGR handled here as well?

And I am not sure if black & white images are handled correctly by torch input connector.

fantes (commented):
I would also add  a predict call after train in ut, sometimes the connector may not have exactly the expected behavior if !train

BynaryCobweb:
There are two ways to create the test dataset:
- by giving a second folder for tests in the `data` field
- by specifying a `test_split`

Currently, only the `test_split` approach has been tested. It would be great if the other approach was tested as well.
To do so, we can:
- Write a new test similar to the current one, only changing the way the test set is loaded
- Write tests specifically for input connectors
- Something else?

Which one should we pick?

src/backends/torch/torchinputconns.cc
beniz: Yes, the RGB/BGR is an API option to the image input connector. It's actually lacking in the API doc, I'll fix that ASAP.  What's the `permute` for here ? 

BW images are handled by the image input connector as well, we need to have them working with torch, it's a matter of the number of channels in the first layer inputs.

beniz:
> * Write a new test similar to the current one, only changing the way the test set is loaded

This could do, maybe a file move from train to test in the `tests/CMakeFiles.txt` for the plankton dataset, or anything else that does the job!

beniz:
> There are two ways to create the test dataset:
> 
> * by giving a second folder for tests in the `data` field
> * by specifying a `test_split`

Note there's a third way, through list of files. We may well want to support that as well as it's supported by our Caffe backend input connector.

src/backends/torch/torchinputconns.cc
BynaryCobweb: OpenCV internal representation is stored in BGR, since we read the raw data we need to convert it back to RGB (at least for the models I used)

But maybe we should not permute if the model require BGR images. I will check that.

> BW images are handled by the image input connector as well, we need to have them working with torch, it's a matter of the number of channels in the first layer inputs.
Ok.

BynaryCobweb: Actually I made a mistake, `permute` is used to permute the dimensions of the tensor, so instead of [width, height, channel] we get [channel, width, height]. This has nothing to do with bgr / rgb.

BynaryCobweb:
I added the tests for loading train & test directories, and predict as well.

Lists of files are used typically for detection? Then it could be done on another PR? I have not tested detection models yet.

src/backends/torch/torchinputconns.cc
beniz: > OpenCV internal representation is stored in BGR, since we read the raw data we need to convert it back to RGB (at least for the models I used)

Let's use the API `rgb` option for this, from the image input connector then.

beniz:
> Lists of files are used typically for detection? Then it could be done on another PR? I have not tested detection models yet.

It's for classification models, for large datasets it's not unusual to have images in many different directories, and thus it is useful to provide txt files for train/test that simply list them. The caffe input connector already does it, it should not be too difficult to add to torch, but feel free to make it another PR.

src/backends/torch/torchinputconns.h
```diff
@@ -36,7 +36,7 @@
 #include "backends/torch/db_lmdb.hpp"
 #include "csvtsinputfileconn.h"
 
-#define TORCH_TEXT_TRANSATION_SIZE 100
+#define TORCH_TEXT_TRANSACTION_SIZE 100
 #define TORCH_IMG_TRANSACTION_SIZE 10
```
beniz, line 40: FYI we use 1000 in caffe input connector.

beniz (approved):

PR #839, opened 2020-09-28
refactor: hide visitor complexity inside the visitor itself

sileht:
## refactor: hide visitor complexity inside the visitor itself

This aims to make the code flow looks like we don't have visitor in the
loop.

The idea is to put all visitor of a particular base class, into
dedicated namespace. Each visitor will have a static method in this
namespace where we move all of visitor boilerplate setup. This avoid
creating intermediate object to store the method result. Using a visitor
become a simple as:

  `int status = visitor_mllib::predict_job(mllib, ad_in, ad_out, chain);`

The visitor helper signature looks like the real mllib method.

## refactor: hide visitor complexity (other)

This applies the same logic for other service visitors.

### DIFF
--- a/src/services.h
+++ b/src/services.h
@@ -216,25 +216,28 @@ namespace dd
     APIData _out;
   };
 
-  class visitor_predict
+  namespace visitor_mllib
   {
-  public:
-    visitor_predict()
+    class v_predict_job
     {
-    }
-    ~visitor_predict()
-    {
-    }
+    public:
+      const APIData &_in;
+      APIData &_out;
+      bool _chain;
 
-    template <typename T> output operator()(T &mllib)
-    {
-      int r = mllib.predict_job(_ad, _out, _chain);
-      return output(r, _out);
-    }
+      template <typename T> int operator()(T &mllib)
+      {
+        return mllib.predict_job(_in, _out, _chain);
+      }
+    };
 
-    APIData _ad;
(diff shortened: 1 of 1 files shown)


sileht:
It's just an example with mllib::predict_job, if you like the idea, I will update all other visitors.

BynaryCobweb (approved):

src/services.h
```diff
@@ -636,67 +648,65 @@ namespace dd
...
           throw;
         }
       catch (...)
         {
           llog->error("prediction call failed");
-          pout._status = -1;
+          status = -1;
           throw;
         }
-      out = pout._out;
```
BynaryCobweb, line 692: It seems like `out` could be modified by the predict call, this PR seems to remove this behaviour (`out` is passed by value to `predict_job()` and the modifications are discarded with the visitor).

BynaryCobweb, line 692: It seems like `out` could be modified by the predict call, this PR seems to remove this behaviour (`out` is passed by value to `predict_job()` and the modifications are discarded with the visitor).

BynaryCobweb (dismissed):

src/services.h
sileht: Fixed.

fantes (approved):

beniz (approved):

PR #842, opened 2020-09-28
fix(native) fix no template param, fallback to default

fantes:
this PR allows user not to give template_param to native modules, allowing default from module to be used

### DIFF
--- a/src/backends/torch/native/native_factory.h
+++ b/src/backends/torch/native/native_factory.h
@@ -44,8 +44,10 @@ namespace dd
   {
     if (tdef.find("nbeats") != std::string::npos)
       {
-        std::vector<std::string> p = template_params.get("template_params")
-                                         .get<std::vector<std::string>>();
+        std::vector<std::string> p;
+        if (template_params.has("template_params"))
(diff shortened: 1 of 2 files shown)


BynaryCobweb (approved):

src/backends/torch/native/native_factory.h
```diff
@@ -44,8 +44,10 @@ namespace dd
   {
     if (tdef.find("nbeats") != std::string::npos)
       {
-        std::vector<std::string> p = template_params.get("template_params")
-                                         .get<std::vector<std::string>>();
+        std::vector<std::string> p;
```
beniz, line 47: What happens when template params are not passed ? Shall we have a default `template_params` value here ?

Looking at the `Nbeats` constructor, the values of some parameters appear to be hardcoded into the constructor itself. 

If possible I'd put them into the class variables init directly, e.g. `int _backcast_length = 50;`.

fantes: indeed the default template_params with this syntax is an empty vector 
will do for default values

sileht (approved):

PR #848, opened 2020-10-01
Feat nbeats expose hidden size in api

fantes:
now template_params can be of the form ["t3","s10","g20","b10",'h25"]  where h denotes a hidden size of 25

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -727,7 +727,16 @@ nclasses        | int    | yes      | none    | if set to some int, add a classi
 self_supervised | string | yes      | ""      | self-supervised mode: "mask" for masked language model
 embedding_size  | int    | yes      | 768     | embedding size for NLP models
 freeze_traced   | bool   | yes      | false   | Freeze the traced part of the net during finetuning (e.g. for classification)
-template        | string | yes      | ""      | for language models, either "bert" or "gpt2"
+template        | string | yes      | ""      | for language models, either "bert" or "gpt2", "recurrent" for LSTM-like models (including autoencoder), "nbeats" for nbeats model
+
+
+Model instantiation parameters:
+
+Parameter       | Template  | Type            | Default                      | Description
+---------       | --------- | ------          | ---------------------------- | -----------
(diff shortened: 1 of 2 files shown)


beniz:
Let's update the API doc as well.

fantes:
> Let's update the API doc as well.

ok done

docs/api.md
```diff
@@ -727,7 +727,16 @@ nclasses        | int    | yes      | none    | if set to some int, add a classi
...
-template        | string | yes      | ""      | for language models, either "bert" or "gpt2"
+template        | string | yes      | ""      | for language models, either "bert" or "gpt2", "recurrent" for LSTM-like models (including autoencoder), "nbeats" for nbeats model
+
+
+Model instantiation parameters:
+
+Parameter       | Template  | Type            | Default                      | Description
+---------       | --------- | ------          | ---------------------------- | -----------
+template_params | nbeats    | array of string | ["t2","s8","g3","b3","h10" ] | default means: trend stack with theta = 2, seasonal stack with theta = 8 , generic stack with theta = 3, 3 blocks per stacks, hidden unit size of 10 everywhere
+layers          | recurrent | array of string | []                           | ["L50","L50"] means 2 layers of LSTMs with hidden size of 50. ["L100","L100", "T", "L300"] means an lstm autoencoder with encoder composed of 2 LSTM layers of hidden size 100 and decoder is one LSTM layer of hidden size 300
```
BynaryCobweb, line 738: What does the "T" stand for?

BynaryCobweb (approved):

docs/api.md
fantes: It means Tile, which is the original name of caffe layers that replicates a matrix along some axis. Here we take one 1d output, ie the hidden state vector, replicate/tile along time axis it so that this hidden state vector will be given as every input for each timestep to the decoder

beniz: Can we abstract it away, i.e. having it inserted automatically when needed  ?

fantes: one need to define encoder on one side and decoder on the other, both of which can be whatever user want. So either we add an option `encoder = ["L50"]` and `decoder = ["L100"]` or we have one option `layers = ["L50", "T", "L100"]` (we can change T to whatever, but we need something to denote the botlleneck with such syntax)

fantes: or we can force the user to use same architecture for encoder and decoder, but I am not sure why we should to that (unless if only for simplifying the API)

PR #850, opened 2020-10-01
feat(build): allow to change CUDA_ARCH

sileht:
They is no need to build all arch for some dev/testing.

Since this have a huge impact on the compilation time,
this allow to configure the list.

This also adds cards supported since cuda 10

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,6 +7,15 @@ deepdetect_arch=(cpu gpu)
 deepdetect_cpu_build_profiles=(default caffe-tf armv7)
 deepdetect_gpu_build_profiles=(default tf caffe-tf-cpu caffe-tf caffe2 p100 volta volta-faiss faiss)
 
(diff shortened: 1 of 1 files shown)


beniz:
FTR, the CUDA arch for the local machine the build takes place on is automatically detected by https://github.com/jolibrain/deepdetect/blob/master/cmake/Cuda.cmake 
So maybe we could set the detected arch value into `CUDA_ARCH` automatically...

sileht:
> FTR, the CUDA arch for the local machine the build takes place on is automatically detected by https://github.com/jolibrain/deepdetect/blob/master/cmake/Cuda.cmake
> So maybe we could set the detected arch value into `CUDA_ARCH` automatically...

`build.sh` is mainly used by docker images. My intent to quickly build an image regulary with Jenkins (with a minimal list in CUDA_ARCH), but for release we should have the complete list we want to support.

PR #864, opened 2020-10-06
fix(torch/timeseries): unscale prediction output if needed

fantes:
this PR unscale the output at prediction time, exaclty as in caffe backend. This step was previously wrongly skipped

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -502,6 +502,46 @@ namespace dd
     empty_cuda_cache();
   }
 
+  template <class TInputConnectorStrategy, class TOutputConnectorStrategy,
+            class TMLModel>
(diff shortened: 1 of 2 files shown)


beniz:
Could we put the code into it's own function to keep torchlib.cc as readable as possible ? If this applies more broadly to any regression job (images -> regression is coming up to our torch backend), this would make even more sense.

Even more general, we could have this into some specialization for the torch backend of the output connector.

fantes:
> Could we put the code into it's own function to keep torchlib.cc as readable as possible ? If this applies more broadly to any regression job (images -> regression is coming up to our torch backend), this would make even more sense.
> 
> Even more general, we could have this into some specialization for the torch backend of the output connector.

ok done

beniz (approved):

BynaryCobweb (approved):

PR #865, opened 2020-10-06
fix(torch): load weights only once

fantes:
this PR tries to cleanup model loading in torchlib.cc, previous version was extra cautious , leading to many unnecessary loadings which caused major slowdowns at inference

### DIFF
--- a/src/backends/torch/torchgraphbackend.cc
+++ b/src/backends/torch/torchgraphbackend.cc
@@ -189,6 +189,7 @@ namespace dd
 
   void TorchGraphBackend::allocate_modules()
   {
+    _allocation_done = false;
     for (BaseGraph::Vertex v : _sortedOps)
       {
         if (!_graph[v].alloc_needed)
@@ -220,6 +221,7 @@ namespace dd
             _modules[opname] = AnyModule(m);
             _graph[v].alloc_needed = false;
             _rnn_has_memories[opname] = false;
+            _allocation_done = true;
           }
         else if (optype == "RNN")
           {
@@ -233,6 +235,7 @@ namespace dd
             _modules[opname] = AnyModule(m);
             _graph[v].alloc_needed = false;
             _rnn_has_memories[opname] = false;
+            _allocation_done = true;
           }
         else if (optype == "InnerProduct")
           {
@@ -243,6 +246,7 @@ namespace dd
                 Linear(LinearOptions(dim(v, 0, 2), num_output(v)).bias(true)));
             _modules[opname] = AnyModule(m);
             _graph[v].alloc_needed = false;
+            _allocation_done = true;
           }
         else if (optype == "Tile")
           _graph[v].alloc_needed = false;
--- a/src/backends/torch/torchgraphbackend.h
+++ b/src/backends/torch/torchgraphbackend.h
@@ -177,10 +177,18 @@ namespace dd
       _parameters_used = false;
     }
 
+    /**
+     * tells if some allocation was done (needs to be called just after
+     * set_inputdim or finalize
+     */
+    bool needs_reload()
+    {
+      return _allocation_done;
+    }
+
   protected:
     /**
      * internal torch module allocation, called whithin (finalize)
-     * @param force
      */
     void allocate_modules();
 
@@ -215,8 +223,9 @@ namespace dd
     std::unordered_map<std::string, bool>
         _rnn_has_memories; /**< true if previsous hidden values are available
                             */
-  };
 
+    bool _allocation_done = false;
+  };
 }
 
 #endif
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -163,6 +163,92 @@ namespace dd
       _classif->to(device, dtype);
   }
 
+  void TorchModule::proto_model_load(const TorchModel &model)
+  {
+    _logger->info("loading " + model._proto);
+    try
+      {
+        _graph = std::make_shared<CaffeToTorch>(model._proto);
+      }
+    catch (std::exception &e)
+      {
+        _logger->info("unable to load " + model._proto);
+        throw;
+      }
+  }
+
+  void TorchModule::graph_model_load(const TorchModel &tmodel)
+  {
+    if (!tmodel._traced.empty() && _graph->needs_reload())
+      {
+        _logger->info("loading " + tmodel._traced);
+        try
+          {
+            torch::load(_graph, tmodel._traced, _device);
+          }
+        catch (std::exception &e)
+          {
+            _logger->error("unable to load " + tmodel._traced);
+            throw;
+          }
+      }
+  }
+
+  void TorchModule::native_model_load(const TorchModel &tmodel)
+  {
+    if (!tmodel._native.empty())
+      {
+        _logger->info("loading " + tmodel._native);
+        try
+          {
+            torch::load(_native, tmodel._native);
+          }
+        catch (std::exception &e)
+          {
+            _logger->error("unable to load " + tmodel._native);
+            throw;
+          }
+      }
+  }
+
+  void TorchModule::classif_model_load(const TorchModel &model)
+  {
+    _logger->info("loading " + model._weights);
+    try
+      {
+        torch::load(_classif, model._weights, _device);
+      }
+    catch (std::exception &e)
+      {
+        _logger->error("unable to load " + model._weights);
+        throw;
+      }
+  }
+
+  void TorchModule::classif_layer_load()
+  {
+    if (!_classif_layer_file.empty())
+      {
+        _logger->info("loading " + _classif_layer_file);
+        torch::load(_classif, _classif_layer_file, _device);
+      }
+  }
+
+  void TorchModule::traced_model_load(TorchModel &model)
+  {
+    _logger->info("loading " + model._traced);
+    try
+      {
+        _traced = std::make_shared<torch::jit::script::Module>(
+            torch::jit::load(model._traced, _device));
+      }
+    catch (std::exception &e)
+      {
+        _logger->error("unable to load " + model._traced);
+        throw;
+      }
+  }
+
   template <class TInputConnectorStrategy>
   void TorchModule::post_transform(const std::string tmpl,
                                    const APIData &template_params,
@@ -170,22 +256,26 @@ namespace dd
                                    const TorchModel &tmodel,
                                    const torch::Device &device)
   {
+    _device = device;
     this->_native = std::shared_ptr<NativeModule>(
         NativeFactory::from_template<TInputConnectorStrategy>(
             tmpl, template_params, inputc));
 
     if (_native)
-      if (!tmodel._native.empty())
-        torch::load(_native, tmodel._native, device);
+      {
+        _logger->info("created net using template " + tmpl);
+        native_model_load(tmodel);
+      }
 
     if (_graph)
       {
         std::vector<long int> dims = inputc._dataset.datasize(0);
         dims.insert(dims.begin(), 1); // dummy batch size
         _graph->finalize(dims);
+        if (_graph->needs_reload())
+          _logger->info("net was reallocated due to input dim changes");
         // reload params after finalize
-        if (!tmodel._traced.empty())
-          torch::load(_graph, tmodel._traced, _device);
+        graph_model_load(tmodel);
       }
     to(_device);
   }
@@ -361,11 +451,7 @@ namespace dd
     // First dimension is batch id
     int outdim = to_tensor_safe(forward(input_example)).sizes()[1];
     _classif = torch::nn::Linear(outdim, nclasses);
-
-    if (!_classif_layer_file.empty())
-      {
-        torch::load(_classif, _classif_layer_file, _device);
-      }
+    classif_layer_load();
   }
 
   std::vector<Tensor> TorchModule::parameters()
@@ -401,13 +487,13 @@ namespace dd
   void TorchModule::load(TorchModel &model)
   {
     if (!model._traced.empty() && model._proto.empty())
-      _traced = std::make_shared<torch::jit::script::Module>(
-          torch::jit::load(model._traced, _device));
+      traced_model_load(model);
+
     if (!model._weights.empty())
       {
         if (_classif)
           {
-            torch::load(_classif, model._weights, _device);
+            classif_model_load(model);
           }
         else if (_require_classif_layer)
           {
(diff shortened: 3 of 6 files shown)


beniz:
What happens when loading a model fails ? Loading is such a critical operation that I believe we may want to intercept exceptions of `torch::load`, log on a failed loading attempt, and rethrow so that it gets caught by DD upper levels, as per design.
What are your thoughts ?

src/backends/torch/torchgraphbackend.h
```diff
@@ -103,27 +103,31 @@ namespace dd
     /**
      * \brief allocates torch modules base on real input dimension
      * @param inputdim input tensor dimension, in int64_t
+     * @return true if some allocation was done
      */
-    void finalize(std::vector<int64_t> inputdim);
+    bool finalize(std::vector<int64_t> inputdim);
```
BynaryCobweb, line 108: Instead of returning a `bool`, there could be a flag (`need_reload` or something else) and a method `needs_reload()` to query this flag. This would avoid returning an unclear value (that needs to be documented) and make things more explicit in my opinion. And the graph could keep its "dirty" state in memory, so we do not need to keep track of it externally and it's more flexible to use (e.g. we do not need to save the result of the `finalize` method in a variable or a class field if we want to perform loading later in the code).

src/backends/torch/torchlib.cc
```diff
@@ -622,16 +644,16 @@ namespace dd
...
 
+    // if (!this->_mlmodel._traced.empty())
+    //   this->_logger->info("Loading ml model from file {}.",
+    //                       this->_mlmodel._traced);
+    // if (!this->_mlmodel._proto.empty())
+    //   this->_logger->info("Loading ml model from file {}.",
+    //                       this->_mlmodel._proto);
+    // if (!this->_mlmodel._weights.empty())
+    //   this->_logger->info("Loading weights from file {}.",
+    //                       this->_mlmodel._weights);
```
BynaryCobweb, line 655: I think we can safely remove that

fantes: ok, done

fantes:
> What happens when loading a model fails ? Loading is such a critical operation that I believe we may want to intercept exceptions of `torch::load`, log on a failed loading attempt, and rethrow so that it gets caught by DD upper levels, as per design.
> What are your thoughts ?

ok, done

src/backends/torch/torchgraphbackend.h
fantes: ok done

src/backends/torch/torchlib.cc
```diff
@@ -175,17 +175,44 @@ namespace dd
...
+          _logger->info("net was reallocated due to input dim changes");
         // reload params after finalize
-        if (!tmodel._traced.empty())
-          torch::load(_graph, tmodel._traced, _device);
+        if (!tmodel._traced.empty() && _graph->needs_reload())
+          {
+            _logger->info("loading " + tmodel._traced);
+            try
+              {
+                torch::load(_graph, tmodel._traced, _device);
```
BynaryCobweb, line 208: Maybe add `set_need_reload(false)` or something like this?

src/backends/torch/torchgraphbackend.cc
```diff
@@ -163,7 +163,12 @@ namespace dd
             std::get<1>(std::get<1>(full_output))); // last memory / c  value
         if (_lstm_continuation)
           {
-            _rnn_memories[opname_v] = std::get<1>(full_output);
+            if (_rnn_has_memories[opname_v])
+              _rnn_memories.erase(opname_v);
+            _rnn_memories.insert(
```
beniz, line 168: Just to make sure, this is not related to this PR, right ?

src/backends/torch/torchlib.cc
```diff
@@ -175,17 +175,44 @@ namespace dd
...
+        if (!tmodel._native.empty())
+          {
+            _logger->info("loading " + tmodel._native);
+            try
+              {
+                torch::load(_native, tmodel._native, device);
+              }
+            catch (std::exception &e)
+              {
+                _logger->warn("unable to load " + tmodel._native);
```
beniz, line 189: I'd use `_logger->error` instead.

src/backends/torch/torchlib.cc
```diff
@@ -175,17 +175,44 @@ namespace dd
...
+        if (!tmodel._traced.empty() && _graph->needs_reload())
+          {
+            _logger->info("loading " + tmodel._traced);
+            try
+              {
+                torch::load(_graph, tmodel._traced, _device);
+              }
+            catch (std::exception &e)
+              {
+                _logger->warn("unable to load " + tmodel._traced);
```
beniz, line 212: `_logger->error` ?

src/backends/torch/torchlib.cc
```diff
@@ -175,17 +175,44 @@ namespace dd
...
         dims.insert(dims.begin(), 1); // dummy batch size
         _graph->finalize(dims);
+        if (_graph->needs_reload())
+          _logger->info("net was reallocated due to input dim changes");
         // reload params after finalize
-        if (!tmodel._traced.empty())
-          torch::load(_graph, tmodel._traced, _device);
+        if (!tmodel._traced.empty() && _graph->needs_reload())
+          {
+            _logger->info("loading " + tmodel._traced);
```
beniz, line 205: I believe we may want a function for this, like `graph_model_load()` to make code lighter.

src/backends/torch/torchlib.cc
```diff
@@ -399,13 +427,33 @@ namespace dd
   void TorchModule::load(TorchModel &model)
   {
     if (!model._traced.empty() && model._proto.empty())
-      _traced = std::make_shared<torch::jit::script::Module>(
-          torch::jit::load(model._traced, _device));
+      {
```
beniz, line 430: Maybe put this block into a function, e.g. `traced_model_load()` ?

src/backends/torch/torchlib.cc
```diff
@@ -884,7 +953,16 @@ namespace dd
...
         this->_logger->info("Restarting optimization from iter {}", it);
-        torch::load(*optimizer, this->_mlmodel._sstate);
+        this->_logger->info("loading " + this->_mlmodel._sstate);
+        try
+          {
+            torch::load(*optimizer, this->_mlmodel._sstate);
+          }
+        catch (std::exception &e)
+          {
+            this->_logger->warn("unable to load " + this->_mlmodel._sstate);
```
beniz, line 963: `_logger->error` ?

src/backends/torch/torchlib.cc
```diff
@@ -1379,7 +1457,7 @@ namespace dd
         unsupo.finalize(ad.getobj("parameters").getobj("output"), out,
                         static_cast<MLModel *>(&this->_mlmodel));
       }
-
+    empty_cuda_cache();
```
beniz, line 1460: Might be unrelated to this PR ?

src/backends/torch/torchlib.h
```diff
@@ -142,6 +142,8 @@ namespace dd
                                 the file where the weights are stored */
     unsigned int _nclasses = 0;
 
+    std::shared_ptr<spdlog::logger> _logger; /**< mllib logger. */
```
beniz, line 145: This is already defined in `MLLib`, I don't think we should have it again here ?

src/caffegraphinput.cc
```diff
@@ -189,7 +189,6 @@ namespace dd
         parse_simple_lstm(net);
         return 0;
       }
-    return 0;
+    return -1;
```
beniz, line 192: This fix is unrelated to this PR.

src/mlservice.h
```diff
@@ -158,7 +158,16 @@ namespace dd
       _init_parameters = ad.getobj("parameters");
       this->_inputc.init(_init_parameters.getobj("input"));
       this->_outputc.init(_init_parameters.getobj("output"));
-      this->init_mllib(_init_parameters.getobj("mllib"));
+      try
```
beniz, line 161: Do you need this ? The main principle has always been to relay lower-level exceptions with their messages, and not obfuscate them at higher levels, when possible.

fantes: I put this here because exception do not seem to be catched at upper level:  dede terminates in case of failure of init_mllib

src/backends/torch/torchgraphbackend.cc
fantes: no, removed

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: ok; removed

src/backends/torch/torchlib.h
fantes: if one want to log directly from torchmodule, I do not see any other way to do

src/caffegraphinput.cc
fantes: this allows to throw smthg if proto do not contains a properly defined LSTM. Will change with smthg proper

src/backends/torch/torchlib.h
beniz: If torchmodule is uniquely associated to an instance of torchlib, I'd suggest that the logger from that torchlib object is used, no ?

src/mlservice.h
beniz: Let's make this another issue, all exceptions should be caught at upper levels.

src/backends/torch/torchlib.h
fantes: torchmodule does not have access to its container , so we give it the pointer to the logger so that it can use the same logger as its torchlib

src/mlservice.h
fantes: removed

src/backends/torch/torchlib.h
beniz: ok

beniz: ok

src/backends/torch/torchlib.cc
```diff
@@ -884,7 +954,16 @@ namespace dd
         size_t end = this->_mlmodel._sstate.rfind(".");
         it = std::stoi(this->_mlmodel._sstate.substr(start, end - start));
         this->_logger->info("Restarting optimization from iter {}", it);
-        torch::load(*optimizer, this->_mlmodel._sstate);
+        this->_logger->info("loading " + this->_mlmodel._sstate);
```
beniz, line 957: Let's make this a function as well

src/backends/torch/torchlib.cc
```diff
@@ -163,6 +163,84 @@ namespace dd
...
+          }
+      }
+  }
+
+  void TorchModule::classif_model_load(const TorchModel &model)
+  {
+    _logger->info("loading " + model._weights);
+    try
+      {
+        torch::load(_classif, model._weights, _device);
```
beniz, line 220: This seems to me a generic loader, from `_model._weight`, I guess this is called `_classif` becauses it started by supporting classification ?

Additionnally, there's no support for the `device` parameter for that type of model ?

src/backends/torch/torchlib.cc
```diff
@@ -163,6 +163,84 @@ namespace dd
...
+      }
+    catch (std::exception &e)
+      {
+        _logger->error("unable to load " + model._weights);
+        throw;
+      }
+  }
+
+  void TorchModule::traced_model_load(TorchModel &model)
```
beniz, line 229: Same remark here regarding the `device` parameter, if applicable.

src/backends/torch/torchlib.cc
BynaryCobweb: _classif is a classification head added on top of traced model for classification tasks. If we want to support others heads, we may want to rename it `_head` and rename `model._weights` to `model._head_weights` ?

src/caffegraphinput.cc
```diff
@@ -181,15 +183,15 @@ namespace dd
   {
     caffe::NetParameter net;
     if (!read_proto(filename, &net))
-      return -1;
+      throw MLLibBadParamException("unable to parse protofile");
```
BynaryCobweb, line 186: In my understanding, this function both returns an error code *and* throws exceptions? Do we need to keep the error code, now it's always 0?

src/backends/torch/torchlib.cc
fantes: ok

src/backends/torch/torchlib.cc
fantes: about device : default behavior is to load to cpu, then `model.to(gpu)` , when i try to load directly to device, i sometimes have errors, so I will fall back to mandatory calls to `.to`

src/backends/torch/torchlib.cc
fantes: see above : I prefer to remove direct loading to device, there is something that does not work with it.

src/caffegraphinput.cc
fantes: ok, removed error code

src/backends/torch/torchlib.cc
fantes: about _model_weight : indeed there will need some cleanup at some time: we can either load a traced model, some weight to put in _classif head, a prototxt, weight to a model that is defined by a prototxt (and that use the same extension as traced nets), and weight for a native model. We should clean up file extensions in order to be able to load several of them in the future.

beniz: > about device : default behavior is to load to cpu, then `model.to(gpu)` , when i try to load directly to device, i sometimes have errors, so I will fall back to mandatory calls to `.to`

Can we double check on that ? Like, are we loading the model twice effectively ? I don't want to slow this PR down, but I fee if the PR is called 'load weights only once', let's check on that :)

fantes: check on what ? do you want to try to load models directly to device ?

beniz (approved):

src/backends/torch/torchlib.cc
```diff
@@ -362,6 +442,7 @@ namespace dd
 
     if (!_classif_layer_file.empty())
       {
+        _logger->info("loading " + _classif_layer_file);
         torch::load(_classif, _classif_layer_file, _device);
```
BynaryCobweb, line 446: Model is loaded here as well (This method is called during training to finalize the classification layer with some example data)
It would be great to replace torch::load with classif_model_load

src/backends/torch/torchlib.cc
fantes: done

src/backends/torch/torchlib.cc
fantes: ok done

PR #867, opened 2020-10-07
fix(torch): Fix conditions to add classification head.

BynaryCobweb:
Classification head was added if there was a traced model, which was not the expected behavior in some cases. It was also added on prediction models in some cases, making them totally useless.

Now the classification layer is added only if we are training with `finetuning:true`.

EDIT: Not sure why it makes nbeats tests fail, I will check that out
EDIT: Seems like the nbeats issue happens randomly, I saved the backtrace for further investigation. This is not related to this PR.

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -187,23 +187,6 @@ namespace dd
         if (!tmodel._traced.empty())
           torch::load(_graph, tmodel._traced, _device);
       }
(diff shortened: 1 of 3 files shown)


beniz:
This is an important step because we expect other head extensions soon, i.e. regression head.

sileht (approved):

sileht (approved):

PR #868, opened 2020-10-07
fix: fix split_data in csvts connector

beniz:
chore: in-code function documentation for csv and csvts connectors

### DIFF
--- a/src/csvinputfileconn.cc
+++ b/src/csvinputfileconn.cc
@@ -221,7 +221,6 @@ namespace dd
           {
             // not a number, skip for now
             if (column_id == col) // if id is string, replace with number /
-                                  // TODO: better scheme
               vals.push_back(c);
             else
               {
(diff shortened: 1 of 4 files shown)


fantes (approved):

sileht:
The commit title doesn't reflex the real change.

Can you change it to: `chore: in-code function documentation for csv and csvts connectors`

beniz:
> The commit title doesn't reflex the real change.
> 
> Can you change it to: `chore: in-code function documentation for csv and csvts connectors` ?

FYI, the important fix is here: https://github.com/jolibrain/deepdetect/blob/285c6322e6df4c9ada948dacb78eb3b780e8386e/src/csvtsinputfileconn.cc#L183

Want 2 commits ?

sileht:
It's always better to have two commits. In case of we need to revert it, we don't revert your cleanup :) 
And also the changelog for the release will be cleaner.

sileht:
Let's say, it's OK for this one.

PR #870, opened 2020-10-07
feat(torch): enable switch on/off of cudnn

fantes:
allows to switch off cudnn in torchlib

parameters.mlllib.cudnn true/false

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -489,6 +489,7 @@ namespace dd
     _best_metric_value = tl._best_metric_value;
     _classification = tl._classification;
     _timeserie = tl._timeserie;
+    _use_cudnn = tl._use_cudnn;
     _loss = tl._loss;
     _template_params = tl._template_params;
   }
@@ -535,6 +536,11 @@ namespace dd
       freeze_traced = lib_ad.get("freeze_traced").get<bool>();
     if (lib_ad.has("loss"))
       _loss = lib_ad.get("loss").get<std::string>();
+
+    if (lib_ad.has("cudnn"))
+      _use_cudnn = lib_ad.get("cudnn").get<bool>();
+
+    at::globalContext().setUserEnabledCuDNN(_use_cudnn);
 
     _device = gpu ? torch::Device(DeviceType::CUDA, gpuid)
                   : torch::Device(DeviceType::CPU);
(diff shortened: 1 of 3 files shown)


src/backends/torch/torchlib.cc
```diff
@@ -536,6 +537,11 @@ namespace dd
     if (lib_ad.has("loss"))
       _loss = lib_ad.get("loss").get<std::string>();
 
+    if (lib_ad.has("cudnn"))
```
beniz, line 540: Let's not duplicate (used 3 times in the PR) this code and have it into a reusable function ?

src/backends/torch/torchlib.cc
```diff
@@ -536,6 +537,11 @@ namespace dd
...
+      _use_cudnn = lib_ad.get("cudnn").get<bool>();
+
+    at::globalContext().setUserEnabledCuDNN(_use_cudnn);
```
beniz, line 543: I've looked at https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/Context.h and my understanding is that `globalContext()` applies to the torch backend library / backend globally. 

If this is true, the usage of `at::globalContext().setUserEnabledCuDNN` is this PR might be wrong in the sense that it modifies a global parameter from within a contextualized single model train/predict function.

I believe we need to make sure the `_use_cudnn` option, as defined per-service in this PR, does effectively apply in a per-service manner, and not globally to the torch backend.

fantes: indeed I am pretty sure that there is no (simple) way to make this per-service, this variable is read:
- at model instantiation
- at every forward

beniz: OK, so this gives us less control than with the Caffe backend at this stage. So obviously what we need is a torch cudnn on/off control at `dede` startup as opposed as a per service control through the API.

Now, maybe there's a way to regain per-layer control with torch much as what Caffe offers, and this may be possible at C++ level, see https://github.com/pytorch/pytorch/blob/7d207363bf10402cf161057f05545d9e96417763/aten/src/ATen/native/Convolution.cpp#L126 for an example.

fantes: I saw that but: 
- it is not clear how to pass such an option at torch level (ie between aten/native and torch there is autograd and registered stubs and I do not understand the how the binding is done) and this option is not exposed in torch conv params (see https://pytorch.org/cppdocs/api/structtorch_1_1nn_1_1_conv_options.html)
- some layers like lstm do not have such 'explicit' option

beniz:
At this stage I believe we may want to close this PR, and return to torch + cudnn elsewhere, typically with a global CUDNN on/off flag for the torch backend instead, since this seems the behavior currently most easily supported from within torch ?

PR #884, opened 2020-10-13
feat(tensorrt): Add support for onnx image classification models

BynaryCobweb:
TensorRT backend can now parse and convert ONNX model to TRT engines, and execute them as well.

The model in the repository must be named `net_tensorRT.onnx` otherwise it will not be detected.

### DIFF
--- a/src/backends/tensorrt/tensorrtlib.cc
+++ b/src/backends/tensorrt/tensorrtlib.cc
@@ -24,6 +24,7 @@
 #include "tensorrtinputconns.h"
 #include "utils/apitools.h"
 #include "NvInferPlugin.h"
+#include "NvOnnxParser.h"
 #include "protoUtils.h"
 #include <cuda_runtime_api.h>
 #include <string>
@@ -39,7 +40,12 @@ namespace dd
     fileops::list_directory(repo, true, false, false, lfiles);
     for (std::string s : lfiles)
       {
-        if (s.find(engineFileName) != std::string::npos)
+        // Ommiting directory name
+        auto fstart = s.find_last_of("/");
+        if (fstart == std::string::npos)
+          fstart = 0;
+
+        if (s.find(engineFileName, fstart) != std::string::npos)
           {
             std::string bs_str;
             for (auto it = s.crbegin(); it != s.crend(); ++it)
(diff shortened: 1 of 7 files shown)


src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -337,60 +352,100 @@ namespace dd
...
 
-            int fixcode
-                = fixProto(this->_mlmodel._repo + "/" + "net_tensorRT.proto",
-                           this->_mlmodel._def);
-            switch (fixcode)
+            nvinfer1::INetworkDefinition *network = nullptr;
+            nvcaffeparser1::ICaffeParser *caffeParser = nullptr;
+            nvonnxparser::IParser *onnxParser = nullptr;
+
+            if (this->_mlmodel._model.find("net_tensorRT.proto")
```
beniz, line 359: Do we want to put that into functions, e.g. `read_trt_caffe_model`, and same for onnx ?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -294,7 +304,12 @@ namespace dd
                 "timeseries not yet implemented over tensorRT backend");
           }
 
-        _nclasses = findNClasses(this->_mlmodel._def, _bbox);
+        if (_nclasses == 0)
```
beniz, line 307: FTR, this works for now for classification / object detection models. For forthcoming OCR and regression models, this will need to be adapted. So we may want to put this into a function, now or later.

src/backends/tensorrt/tensorrtlib.cc
BynaryCobweb: Done

src/backends/tensorrt/tensorrtlib.cc
BynaryCobweb: What exactly do we want to put in a function?

fantes (approved):

src/backends/tensorrt/tensorrtlib.cc
sileht: @beniz ?

beniz (approved):

PR #886, opened 2020-10-14
fix(torch/native): prevent loading weights before instanciating native

fantes:
naive instantiation is deferred to predict/training time, one should not try to load weight during init_mllib (this happens when predicting only, ie w/o training first)

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -196,7 +196,8 @@ namespace dd
 
   void TorchModule::native_model_load(const TorchModel &tmodel)
   {
-    if (!tmodel._native.empty())
+    if (!tmodel._native.empty()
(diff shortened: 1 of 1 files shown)


sileht (approved):

BynaryCobweb (approved):

beniz:
Just curious, why is that ?

BynaryCobweb:
I think it's because the dimensions of some layers in native models are determined by the dimensions of training data

beniz:
But when the model exists we should know the input dim

fantes:
> But when the model exists we should know the input dim

for csv / csvts time series we know only the number of columns (input/oput vars) after (first) transform(). Still torchlib tried to load weights during init_mllib , (using a default value) which lead to some exception throwed.

beniz (approved):

PR #894, opened 2020-10-16
chore: update tensorrt API to remove deprecated calls

beniz:


src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -181,583 +182,593 @@ namespace dd
...
           }
         else
           {
-            _builder->setInt8Mode(false);
-            _builder->setFp16Mode(false);
+            // default is TF32 (aka 10-bit mantissas round-up)
           }
       }
-#pragma GCC diagnostic push
-  }
```
sileht, line 219: This `}` is missing.

sileht (approved):

PR #902, opened 2020-10-19
feat(api): add versions and compile flags to /info

### LINKED ISSUE #897: DeepDetect info call does not return correct version number
When requesting server info, for example with curl
```
curl -X GET "http://localhost:8080/info"
```
The returned version number does not match the current version of dd, e.g
```
{
  "status": {
    "code": 200,
    "msg": "OK"
  },
  "head": {
    "method": "/info",
    "version": "0.1",
    "branch": "master",
    "commit": "28ab168ee9766ea5b765ccfdf5ca619de507fd75",
    "services": []
  }
}
```
The version should be something like 0.9 or 0.10 and not 0.1.

sileht:
Close #897

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 2.8.8)
 project(deepdetect)
 
+option(RELEASE "release mode" OFF)
 option(USE_CAFFE2 "build caffe2 backend")
 option(USE_TF "use TF backend")
 option(USE_NCNN "use NCNN backend")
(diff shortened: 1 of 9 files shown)


fantes (approved):

BynaryCobweb:
I cloned the repo and tried to build with caffe only (`cd build; cmake ../; make -j6`) and caffe build fails with error:
```
cc1plus: error: .build_release/src/caffe/proto/caffe.pb.cc: Input/output error
```

BynaryCobweb (approved):

PR #905, opened 2020-10-19
feat(ml): tensorrt support for regression models

beniz:
This PR adds support for compiling models that end with a regression head. It hasn't been added to the unit tests since we don't have image regression models in our examples set yet.

### DIFF
--- a/src/backends/tensorrt/tensorrtlib.cc
+++ b/src/backends/tensorrt/tensorrtlib.cc
@@ -88,10 +88,11 @@ namespace dd
     _readEngine = tl._readEngine;
     _writeEngine = tl._writeEngine;
     _TRTContextReady = tl._TRTContextReady;
-    _timeserie = tl._timeserie;
     _buffers = tl._buffers;
     _bbox = tl._bbox;
     _ctc = tl._ctc;
+    _timeserie = tl._timeserie;
+    _regression = tl._regression;
     _inputIndex = tl._inputIndex;
     _outputIndex0 = tl._outputIndex0;
     _outputIndex1 = tl._outputIndex1;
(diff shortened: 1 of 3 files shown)


sileht:
api doc ?

BynaryCobweb:
Maybe add a test as well

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -451,7 +457,10 @@ namespace dd
           }
         else // classification
```
BynaryCobweb, line 458: Comment is outdated, should be `classification / regression` ?

Same thing l.677

beniz:
> api doc ?

It should already be as `regression` in the input connector, same as with other backends for regression.

sileht:
> > api doc ?
> 
> It should already be as `regression` in the input connector, same as with other backends for regression.

tensorrt doesn't have the boolean: https://github.com/jolibrain/deepdetect/blame/master/docs/api.md#L357

fantes (approved):

beniz:
> > It should already be as `regression` in the input connector, same as with other backends for regression.
> 
> tensorrt doesn't have the boolean: https://github.com/jolibrain/deepdetect/blame/master/docs/api.md#L357

OK I believe it's actually missing here: https://github.com/jolibrain/deepdetect/blame/master/docs/api.md#L1225 since `regression` should be put into the `output` object (and not the `input` connector as I wrongly said above), and that's what the code does (https://github.com/jolibrain/deepdetect/pull/905/files#diff-0c973a41705db4d5bf4fa9c5640e078702ffbad55e1c63bfd7b384e308313284R269). 

Some other parameters appear to be missing from the `output` object in the API doc actually, so I'll fix the `regression` param doc here and the others in a different PR.

sileht:
:cry:  no test

PR #907, opened 2020-10-20
feat: make python client an install package

sileht:
This will allow to install it directly in jupyter docker image:

https://github.com/jolibrain/dd_platform_docker/pull/39


fantes (approved):

clients/python/dd_client/bench.py
```diff
@@ -0,0 +1,200 @@
+"""
+DeepDetect benchmark tool
+
+Licence:
+Copyright (c) 2017 Emmanuel Benazationa
```
beniz, line 5: I could spell this better :)

clients/python/dd_client/bench.py
```diff
@@ -0,0 +1,200 @@
+"""
```
beniz, line 1: just curious, can we use `git mv` for renaming `dd_bench.py` into `bench.py` ?

beniz (approved):

clients/python/dd_client/bench.py
sileht: It's git that have record my `mv` into an `rm` / `add`. I think this is because the dd_client directory was not existing in the tree before.

PR #909, opened 2020-10-22
feat(torch): Add multigpu support for native models training

BynaryCobweb:
This PR adds multi-GPU support to model training with libtorch, and uses NBeats implementation as a testbed + unit tests.

### DIFF
--- a/src/backends/torch/native/templates/nbeats.h
+++ b/src/backends/torch/native/templates/nbeats.h
@@ -48,7 +48,7 @@
 
 namespace dd
 {
-  class NBeats : public NativeModule
+  class NBeats : public NativeModuleImpl<NBeats>
   {
 
     enum BlockType
@@ -58,7 +58,7 @@ namespace dd
       generic
     };
 
-    class BlockImpl : public torch::nn::Module
+    class BlockImpl : public virtual torch::nn::Module
     {
     public:
       BlockImpl(int units, int thetas_dim, int backcast_length,
@@ -69,10 +69,14 @@ namespace dd
         init_block();
       }
 
-      BlockImpl(BlockImpl &b)
+      BlockImpl(const BlockImpl &b)
           : torch::nn::Module(b), _units(b._units), _thetas_dim(b._thetas_dim),
             _data_size(b._data_size), _backcast_length(b._backcast_length),
             _share_thetas(b._share_thetas)
+      {
+      }
+
+      void reset()
       {
         init_block();
       }
@@ -98,7 +102,8 @@ namespace dd
 
     typedef torch::nn::ModuleHolder<BlockImpl> Block;
 
-    class SeasonalityBlockImpl : public BlockImpl
+    class SeasonalityBlockImpl : public BlockImpl,
+                                 public Cloneable<SeasonalityBlockImpl>
     {
     public:
       SeasonalityBlockImpl(int units, int thetas_dim, int backcast_length,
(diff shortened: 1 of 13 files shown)


src/backends/torch/native/templates/nbeats.h
```diff
@@ -87,9 +92,25 @@ namespace dd
...
       }
 
-      SeasonalityBlockImpl(SeasonalityBlockImpl &b)
-          : BlockImpl(b), _bS(b._bS), _fS(b._fS)
+      SeasonalityBlockImpl(const SeasonalityBlockImpl &b)
+          : torch::nn::Module(b), BlockImpl(b), _bS(b._bS), _fS(b._fS)
+      {
+      }
+
+      SeasonalityBlockImpl &operator=(SeasonalityBlockImpl &&b)
```
beniz, line 100: Just curious, would there be a way to have the cloning behavior + operator at higher level once and for all, heritable ? This would allow writing native modules without considering details of having one ore more GPUs.

src/backends/torch/torchlib.cc
```diff
@@ -66,6 +66,7 @@ namespace dd
     _template = tl._template;
     _nclasses = tl._nclasses;
     _device = tl._device;
+    _multigpu_devices = tl._multigpu_devices;
```
beniz, line 69: Mini detail but I believe it's called `gpuid` everywhere else in DD.

src/backends/torch/torchlib.cc
```diff
@@ -160,8 +164,27 @@ namespace dd
...
+
+        if (gpuids.empty())
+          gpuids.push_back(0);
+
+        for (int gpuid : gpuids)
+          _multigpu_devices.push_back(torch::Device(DeviceType::CUDA, gpuid));
+        _device = _multigpu_devices[0];
+
+        if (_multigpu_devices.size() >= 2)
+          this->_logger->info("Running on multiple devices");
```
beniz, line 186: Maybe indicate the GPU ids.

src/backends/torch/torchlib.cc
```diff
@@ -459,6 +482,10 @@ namespace dd
     if (iter_size <= 0)
       iter_size = 1;
 
+    size_t gpu_count = _multigpu_devices.size();
+    if (gpu_count >= 2)
```
beniz, line 486: `> 1` ?

src/backends/torch/torchmodule.h
```diff
@@ -42,6 +47,8 @@ namespace dd
     TorchModule();
 
     c10::IValue forward(std::vector<c10::IValue> source);
```
beniz, line 49: How do we choose the GPU in this forward method ? Is it possible to have a single forward and pass the id of one or more GPUs ?

src/backends/torch/native/templates/nbeats.h
BynaryCobweb: This is a special case where the default operator does not have the right behavior, as explained in the comment below.
Most of the time default operators are sufficient and torch::nn::Cloneable::clone does the job correctly.

The `reset` method is required by `torch::nn::Cloneable` (it is pure virtual) so a native model is forced to be clonable (and thus parallelizable)

src/backends/torch/torchmodule.h
BynaryCobweb: I put the multi-gpu stuff in a different method since the calls to torch API are not the same (simple `forward` for single gpu vs call to `data_parallel` for multiple gpus). Usually, the module is located on one device, and this device is selected beforehand. With more GPUs, the model must be cloned dynamically during the forward call (or at least this is the way torch does it).

I can make a single forward and passing the ids

src/backends/torch/torchutils.cc
```diff
@@ -94,5 +94,23 @@ namespace dd
                          params);
         }
     }
+
+    std::vector<c10::IValue> unwrap_c10_vector(const c10::IValue &output)
+    {
+      if (output.isTensorList())
+        {
+          auto elems = output.toTensorList();
+          return std::vector<c10::IValue>(elems.begin(), elems.end());
```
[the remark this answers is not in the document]
BynaryCobweb: c10::IValue is a shared pointer, just like Tensors, so the values are not actually copied.

BynaryCobweb:
Should we add multi GPU support for test as well ?

src/backends/torch/torchmodule.h
BynaryCobweb: I changed `forward_multigpu` to `forward_on_devices` that selects the single or multi gpu mode automatically.

fantes (approved):
Nice work!

PR #910, opened 2020-10-22
build: change min cmake version

sileht:


CMakeLists.txt
```diff
@@ -16,6 +15,12 @@ option(USE_SIMSEARCH "build index and search services" OFF)
 option(USE_ANNOY "use annoy as indexer" OFF)
 option(USE_FAISS "use FAISS as indexer" ON)
 
+if (USE_TF)
```
beniz, line 18: My understanding is that additionnally, whenever we use CUDA (10+), cmake 3.14 is needed, maybe we could add this requirement within this PR as well ?

sileht: So maybe we should just require cmake 3.14 no ?

PR #919, opened 2020-10-26
fix: build dd with tensorrt backend

YaYaB:


sileht:
Thank @YaYaB , I also updated the script that generate the file, so we should not forget it anymore.

PR #924, opened 2020-11-01
feat(ml): image regression model training with libtorch

beniz:
This PR adds support regression on image models with the `torch` model.

This includes the reading of datasets in the standard DD form of a list of image path and regression targets.

On the side it slightly refactors the dynamic adding of a linear layer on top of of a pretrained torch model, since this is applicable to both `classification` and `regression` cases.

### DIFF
--- a/src/backends/torch/torchinputconns.cc
+++ b/src/backends/torch/torchinputconns.cc
@@ -393,6 +393,7 @@ namespace dd
       std::unordered_map<std::string, int> &hcorresp_r,
       const std::string &folderPath)
   {
+    _logger->info("Reading image folder {}", folderPath);
 
     // TODO Put file parsing from caffe in common files to use it in other
     // backends
@@ -424,6 +425,35 @@ namespace dd
         ++cl;
         ++uit;
       }
+  }
+
+  void ImgTorchInputFileConn::read_image_list(
+      std::vector<std::pair<std::string, std::vector<double>>> &lfiles,
+      const std::string &listfilePath)
+  {
+    _logger->info("Reading image list file {}", listfilePath);
+    std::ifstream infile(listfilePath);
+    std::string line;
+    int line_num = 1;
+    while (std::getline(infile, line))
+      {
+        std::istringstream iss(line);
+        string filename;
+        string label;
+        iss >> filename >> label;
+
+        std::vector<double> targets;
+        std::istringstream ill(label);
+        std::string targ;
+        while (ill >> targ)
+          targets.push_back(std::stod(targ));
+
+        ++line_num;
+        lfiles.push_back(
+            std::pair<std::string, std::vector<double>>(filename, targets));
+      }
+    _logger->info("Read {} lines in image list file {}", line_num,
+                  listfilePath);
   }
 
   void ImgTorchInputFileConn::transform(const APIData &ad)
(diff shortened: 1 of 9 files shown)


src/backends/torch/torchmodule.cc
```diff
@@ -222,6 +222,26 @@ namespace dd
...
+                               const_cast<TInputConnectorStrategy &>(inputc)
+                                   .get_input_example(device));
+            _linear->to(_device);
+          }
+        catch (std::exception &e)
+          {
+            throw MLLibInternalException(std::string("Libtorch error: ")
+                                         + e.what());
+          }
+      }
```
BynaryCobweb, line 243: If this is both in `post_transform_train` and `post_transform_predict` it can go in `post_transform` instead

src/backends/torch/torchinputconns.cc
```diff
@@ -642,6 +677,55 @@ namespace dd
...
+    return targett;
+  }
+
+  at::Tensor
+  ImgTorchInputFileConn::target_to_tensor(std::vector<double> &target)
+  {
+    int64_t tsize = target.size();
+    at::Tensor targett = torch::zeros(tsize, torch::kFloat32);
+    int n = 0;
+    for (auto i : target) // XXX: from_blob does not seem to work, fills up
```
fantes, line 692: from_blob needs a .clone() afterwards, default behavior is to only use pointer to data , clone copies into a new tensor, maybe this is your problem here ?

beniz: Thanks, I tried, but same spurious data output with `clone()`.

src/backends/torch/torchinputconns.cc
```diff
@@ -508,67 +540,69 @@ namespace dd
...
+                    add_image_file<std::vector<double>>(_dataset, lfile.first,
+                                                        lfile.second);
+                  }
+
+                for (const std::pair<std::string, std::vector<double>> &lfile :
+                     test_lfiles)
+                  {
+                    add_image_file<std::vector<double>>(
+                        _test_dataset, lfile.first, lfile.second);
+                  }
```
BynaryCobweb, line 604: I think it would be great to have dataset file list building and test split in ImgInputConnector, while keeping conversion to tensor and saving to database in the Torch connector. I see two ways of doing that: either we pass a flag to img input connector so that it only reads uris and let the connector read the image files and put them in the database, or ImgInputConnector preprocesses images and call a hook to put the images in db instead of saving them to `ImgInputConnector::_images`. Or something else ?

This can be done in a future PR as well.

src/backends/torch/torchmodule.h
```diff
@@ -50,10 +50,20 @@ namespace dd
...
-    /** Add linear model at the end of module. Automatically detects size of
+    void setup_linear_layer(int nclasses,
+                            std::vector<c10::IValue> input_example);
+
+    /** Add linear layer at the end of module. Automatically detects size of
      * the last layer thanks to the provided example output.*/
-    void setup_classification(int nclasses,
-                              std::vector<c10::IValue> input_example);
+    /*void setup_classification(int nclasses,
+      std::vector<c10::IValue> input_example);*/
```
BynaryCobweb, line 59: Is something planned with these methods, or can they be removed ?

beniz: good catch, done.

src/backends/torch/torchinputconns.cc
beniz: Sure this can be discussed outside this PR. Mainly caffe and torch backends allow training. So refactoring would remove local code which is good.

fantes (approved):

src/backends/torch/torchmodule.h
```diff
@@ -50,10 +50,8 @@ namespace dd
 
     void freeze_traced(bool freeze);
 
-    /** Add linear model at the end of module. Automatically detects size of
-     * the last layer thanks to the provided example output.*/
```
BynaryCobweb, line 54: It would have been great to keep the comment

beniz: tks, done.

sileht (commented):
Maybe some tests ?

beniz:
> Maybe some tests ?

Done.

src/backends/torch/torchlib.cc
```diff
@@ -610,7 +631,16 @@ namespace dd
...
+            else if (_regression)
+              {
+                if (_loss.empty() || _loss == "L1" || _loss == "l1")
+                  loss = torch::l1_loss(y_pred, y);
+                else if (_loss == "L2" || _loss == "l2" || _loss == "eucl")
+                  loss = torch::mse_loss(y_pred, y);
+                else
+                  throw MLLibBadParamException("unknown loss " + _loss);
+              }
+            else // classification
```
sileht, line 643: It's often better to be explicit about this, like:
```
else if (_classification){
    ...
} else {
     throw MLLibBadParamException("unexpected model type");
}
```

beniz: done

PR #929, opened 2020-11-03
fix(timeseries): do not segfault if no valid files in train/test dir

fantes:
throw a nice exception instead


sileht (approved):

sileht:
Can we add a test to avoid any regression for this?

fantes:
> Can we add a test to avoid any regression for this?

ok, done

PR #932, opened 2020-11-03
chore(torch): add doc + minor cleanups

fantes:
this PR adds doc to methods and attributes along with some minor cleanups

### DIFF
--- a/src/backends/torch/torchinputconns.h
+++ b/src/backends/torch/torchinputconns.h
@@ -35,6 +35,7 @@
 #include "backends/torch/db.hpp"
 #include "backends/torch/db_lmdb.hpp"
 #include "csvtsinputfileconn.h"
+#include "torchutils.h"
 
 #define TORCH_TEXT_TRANSACTION_SIZE 100
 #define TORCH_IMG_TRANSACTION_SIZE 10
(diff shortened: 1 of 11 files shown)


sileht (approved):

src/backends/torch/torchinputconns.h
```diff
@@ -80,71 +94,127 @@ namespace dd
...
+     */
     c10::optional<size_t> size() const override
     {
       return cache_size();
     }
 
+    /**
+     * \brief test of emptyness
+     */
     bool empty() const
```
BynaryCobweb, line 186: image

src/backends/torch/torchlib.h
```diff
@@ -71,20 +82,22 @@ namespace dd
...
-    bool _timeserie = false;
-    std::string _loss = "";
-
-    APIData _template_params;
-
-    // models
-    TorchModule _module;
+    unsigned int _nclasses = 0; /**< number of classes*/
+    std::string _template; /**< template identifier (recurrent/bert/gpt2...)*/
+    bool _finetuning = false; /**< add a classif layer after model */
```
BynaryCobweb, line 87: add a custom head? because with regression #924 it's not a classification layer anymore

src/backends/torch/torchmodule.h
```diff
@@ -35,53 +35,102 @@
...
     c10::IValue extract(std::vector<c10::IValue> source,
                         std::string extract_layer);
 
+    /**
+     * \brief checks if extract_layer is extractible
+     */
     bool extractable(std::string extract_layer) const;
+
+    /**
+     * \brief gives all extratible layers
```
BynaryCobweb, line 65: extractible

src/backends/torch/torchmodule.h
```diff
@@ -35,53 +35,102 @@
...
+     */
     std::vector<std::string> extractable_layers() const;
 
+    /**
+     * \brief freeze traced net so tht it is not updated during learning
```
BynaryCobweb, line 70: so that

src/backends/torch/torchmodule.h
```diff
@@ -35,53 +35,102 @@
...
+     */
     void load(TorchModel &model);
 
+    /**
+     * \brief set net to be in eval mode (ie disable dropout ...)
+     */
     void eval();
+
+    /**
+     * \brief set net to be in train mode (ie ensable dropout ...)
```
BynaryCobweb, line 104: enable

src/backends/torch/torchutils.h
```diff
@@ -40,26 +40,50 @@ namespace dd
 
   namespace torch_utils
   {
+
+    /**
+     * \brief empty cuda caching allocator, not clear when useful
```
BynaryCobweb, line 45: This should be called after every job that allocates a model. Pytorch keeps cuda memory allocated event after the deletion of tensors, to avoid reallocating later. Sometimes it means reserving all the GPU ram even if training is over.

src/backends/torch/torchlib.h
fantes: ok done

src/backends/torch/torchmodule.h
fantes: ok

src/backends/torch/torchmodule.h
fantes: ok

src/backends/torch/torchmodule.h
fantes: ok

src/backends/torch/torchutils.h
fantes: ok

beniz (approved):

BynaryCobweb (approved):

PR #933, opened 2020-11-03
fix: ignore JSON conversion throw in partial chains output

beniz:
This PR fixes execution of `chains` that are only partially executed. This is because the inner APIData structure is not cleaned up and remains in a partial state that was throwing on JSON conversion.

This is not optimal but fixes chains and falls back to previous behavior (ignoring unsupported types upon JSON conversion)

### DIFF
--- a/src/apidata.h
+++ b/src/apidata.h
@@ -472,18 +472,15 @@ namespace dd
     void operator()(const std::vector<cv::Mat> &vcv)
     {
       (void)vcv;
-      // Not Implemented
-      throw DataConversionException(
-          "JSON conversion of std::vector<cv::Mat> is not supported");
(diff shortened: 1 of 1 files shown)


sileht (commented):
This just hides another issue, this is a really bad practice to ignore things in API.

When I read this, I just see we put a thing in APIData where its end goal is not to be serialized in JSON. I'm guessing we have an object that should go from one place to another and we used APIData to move the data around instead of writing a correct interface to have the data accessible.

beniz:
> This just hides another issue, this is a really bad practice to ignore things in API.

Agreed, as discussed, added a comment on why we are ignoring this until APIData disappears and API in/out data is automatically validated.

sileht (approved):

PR #949, opened 2020-11-06
fix(caffe): check model destination permissions

sileht:


src/backends/caffe/caffelib.cc
```diff
@@ -181,6 +181,12 @@ namespace dd
     this->_mlmodel._model_template = model_tmpl;
     this->_logger->info("instantiating model template {}", model_tmpl);
 
+    if (!fileops::is_directory_writable(this->_mlmodel._repo))
```
beniz, line 184: we probably could have this check higher up / more generic, maybe in `mlmodel` ?

fantes (approved):

beniz (approved):

PR #950, opened 2020-11-06
fix(torch): clip gradient in rectified adam as stated in annex B of o…

fantes:
as stated in annex B of original paper

new params : solver.clip true | false            (default false, true for ranger)
                      solver.clip_value  (default 5.0)

### DIFF
--- a/src/backends/torch/optim/ranger.cc
+++ b/src/backends/torch/optim/ranger.cc
@@ -47,7 +47,10 @@ namespace dd
            && (std::get<1>(lhs.betas()) == std::get<1>(rhs.betas()))
            && (lhs.eps() == rhs.eps())
            && (lhs.weight_decay() == rhs.weight_decay())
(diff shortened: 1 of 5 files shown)


src/backends/torch/optim/ranger.cc
```diff
@@ -174,6 +179,8 @@ namespace dd
                 grad.add_(-grad.mean(torch::IntArrayRef(dim), true));
               }
 
+            grad.clamp_(-options.clip(), options.clip());
```
beniz, line 182: Just checking: this is outside the `rectified` conditional.

BynaryCobweb (approved):

src/backends/torch/optim/ranger.cc
fantes: clipping is now explicitly controllable (and forced to true in case of rectified)

beniz (approved):

PR #952, opened 2020-11-09
feat(ml): added vision tranformer (ViT) as torch native template

beniz:
This PR adds Vision Transformer (ViT) from https://arxiv.org/abs/2010.11929 as a native C++ model to the DD torch backend. 

The current implementation is adapted from https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/vision_transformer.py

Remaining functionalities to implement:
- [x] passing template parameters to the model constructor
- [x] add unit test

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -274,7 +274,12 @@ namespace dd
       }
     else if (!_template.empty())
       {
-        throw MLLibBadParamException("template");
+        if (NativeFactory::valid_template_def(_template))
+          _module.create_native_template<TInputConnectorStrategy>(
+              _template, lib_ad, this->_inputc, this->_mlmodel, _main_device);
+        else
+          throw MLLibBadParamException("invalid torch model template "
+                                       + _template);
       }
 
     if (_classification)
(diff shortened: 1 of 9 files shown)


src/backends/torch/torchlib.cc
```diff
@@ -231,7 +231,8 @@ namespace dd
       }
     else if (!_template.empty())
       {
-        throw MLLibBadParamException("template");
+        _module.create_native_template<TInputConnectorStrategy>(
+            _template, lib_ad, this->_inputc, this->_mlmodel, _device);
```
BynaryCobweb, line 235: No error is thrown if the template is not valid. Check if template is valid with `NativeFactory::valid_template_def(_template))`?

beniz: The call is made earlier https://github.com/jolibrain/deepdetect/blob/feat_ml_vit_torch_native/src/backends/torch/torchlib.cc#L190
I left it there actually instead of moving it.

BynaryCobweb: Ok I see
The problem is, the template is checked with the presence of traced nets or proto file all at once in a single condition. 
So if the user specifies a wrong template name with a traced net (e.g. "bret" instead of "bert" 
🙂), dede will try to create a native template and fail silently

fantes (approved):

src/backends/torch/torchlib.cc
beniz: You are right, will fix.

beniz: done.

BynaryCobweb (approved):

PR #953, opened 2020-11-09
feat(torch): Add native resnet support

BynaryCobweb:
This PR adds torchvision dependency to include all the native vision models already developed. The version of torchvision must be updated in the cmake to remain compatible with current version of pytorch.

I am not sure yet how to make the models compatible with multigpu pipeline. Running a model on multigpu requires it to be cloneable, and it looks like torchvision models are not (yet to confirm). So the solutions might be:
- patch torchvision
- copy models in dede and change them NativeModule (so the torchvision dependency would not be required anymore?)
- waiting torchvision to implement multi-gpu: https://github.com/pytorch/vision/issues/3052

Additionnal points:
- Should we add other models (resnet%d, vgg, alexnet...) available in torchvision in this PR ?
- It is possible to load weights from python model on native, however the finetuning behavior is not clearly defined for native nets: Do we add the classification layer on top of the native net? or change native net architecture and not initialize the top layer?


CMakeLists.txt
```diff
@@ -730,6 +730,30 @@ if (USE_TORCH)
   set(TORCH_INC_DIR ${TORCH_LOCATION}/include/ ${TORCH_LOCATION}/include/torch/csrc/api/include/ ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/torch/include/torch/csrc/api/include ${TORCH_LOCATION}/.. ${CMAKE_BINARY_DIR}/src)
   set(TORCH_LIB_DIR ${TORCH_LOCATION}/lib/)
 
+  # TORCH VISION
+  message(STATUS "Configuring Pytorch Vision")
+  set(PYTORCH_VISION_COMPLETE ${CMAKE_BINARY_DIR}/CMakeFiles/pytorch_vision-complete)
```
sileht, line 735: We should let cmake handle the update itself and not looks at this file.

CMakeLists.txt
```diff
@@ -730,6 +730,30 @@ if (USE_TORCH)
...
+    PREFIX pytorch_vision
+    INSTALL_DIR ${CMAKE_BINARY_DIR}
+    GIT_REPOSITORY https://github.com/pytorch/vision.git
+    GIT_TAG ${PYTORCH_VISION_COMMIT}
+    GIT_CONFIG advice.detachedHead=false
+    UPDATE_DISCONNECTED 1
+    # CONFIGURE_COMMAND test -f ${PYTORCH_VISION_COMPLETE} && echo Skipping || cmake ../pytorch_vision -DWITH_CUDA=ON
+    CMAKE_ARGS -DWITH_CUDA=${PYTORCH_USE_CUDA} -DCMAKE_PREFIX_PATH=${TORCH_LOCATION} -DCMAKE_INSTALL_PREFIX=${TORCHVISION_LOCATION}
+      "-DCMAKE_CXX_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/" "-DCMAKE_CUDA_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/"
+    BUILD_COMMAND test -f ${PYTORCH_VISION_COMPLETE} && echo Skipping || make -j6
```
sileht, line 750: I think we should keep the default, build command

CMakeLists.txt
```diff
@@ -730,6 +730,30 @@ if (USE_TORCH)
...
+  # set(PYTORCH_VISION_COMMIT b68adcf9a9280aef02fc08daed170d74d0892361) # 0.6 & torch 1.5
+  set(PYTORCH_VISION_COMMIT 78ed10cc51067f1a6bac9352831ef37a3f842784) # 0.7 & torch 1.6
+  set(TORCHVISION_LOCATION "${CMAKE_BINARY_DIR}/pytorch_vision/src/pytorch_vision-install")
+  ExternalProject_Add(
+    pytorch_vision
+    PREFIX pytorch_vision
+    INSTALL_DIR ${CMAKE_BINARY_DIR}
+    GIT_REPOSITORY https://github.com/pytorch/vision.git
+    GIT_TAG ${PYTORCH_VISION_COMMIT}
+    GIT_CONFIG advice.detachedHead=false
```
sileht, line 745: I have no idea why we need that, but I think cmake doesn't like it to detect if it should rebuild the project or not.

CMakeLists.txt
```diff
@@ -730,6 +730,30 @@ if (USE_TORCH)
...
+    UPDATE_DISCONNECTED 1
```
sileht, line 746: Not sure it's needed.

CMakeLists.txt
```diff
@@ -730,6 +730,30 @@ if (USE_TORCH)
...
+    # CONFIGURE_COMMAND test -f ${PYTORCH_VISION_COMPLETE} && echo Skipping || cmake ../pytorch_vision -DWITH_CUDA=ON
+    CMAKE_ARGS -DWITH_CUDA=${PYTORCH_USE_CUDA} -DCMAKE_PREFIX_PATH=${TORCH_LOCATION} -DCMAKE_INSTALL_PREFIX=${TORCHVISION_LOCATION}
+      "-DCMAKE_CXX_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/" "-DCMAKE_CUDA_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/"
+    BUILD_COMMAND test -f ${PYTORCH_VISION_COMPLETE} && echo Skipping || make -j6
+  )
+
+  list(APPEND TORCH_INC_DIR "${TORCHVISION_LOCATION}/include/")
+  list(APPEND TORCH_LIB_DIR "${TORCHVISION_LOCATION}/lib/")
+  list(APPEND TORCH_LIB_DEPS torchvision)
```
sileht, line 755: I think they must be added on each COMMON_XXX list at the file bottom.

Also an add_dependencies maybe need in src/CMakelists.txt

BynaryCobweb: Torchvision stuff is added to TORCH_XXX variables, so it's added in the COMMON_XXX variables

I did it that way because torchvision is always built with pytorch.

CMakeLists.txt
BynaryCobweb: Other libraries use XXX_COMPLETE variables the same way, maybe it should be changed everywhere as well

CMakeLists.txt
BynaryCobweb: It seems like 1 is the default value. You prefer that we don't specify it explicitely?

CMakeLists.txt
BynaryCobweb: Some patches are applied to pytorch, that's why this kind of warning is removed I think.
On torchvision there are no patches, so it does not matter

CMakeLists.txt
BynaryCobweb: ok

CMakeLists.txt
sileht: My bad I misread the variable name.

CMakeLists.txt
```diff
@@ -749,6 +749,25 @@ if (USE_TORCH)
   set(TORCH_INC_DIR ${TORCH_LOCATION}/include/ ${TORCH_LOCATION}/include/torch/csrc/api/include/ ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/torch/include/torch/csrc/api/include ${TORCH_LOCATION}/.. ${CMAKE_BINARY_DIR}/src)
   set(TORCH_LIB_DIR ${TORCH_LOCATION}/lib/)
 
+  # TORCH VISION
+  message(STATUS "Configuring Pytorch Vision")
+  # set(PYTORCH_VISION_COMMIT b68adcf9a9280aef02fc08daed170d74d0892361) # 0.6 & torch 1.5
+  set(PYTORCH_VISION_COMMIT 78ed10cc51067f1a6bac9352831ef37a3f842784) # 0.7 & torch 1.6
```
sileht, line 755: If this is a git tag, I'm in favor of using the tag name instead of  sha1.

CMakeLists.txt
```diff
@@ -749,6 +749,25 @@ if (USE_TORCH)
...
+    PREFIX pytorch_vision
+    INSTALL_DIR ${CMAKE_BINARY_DIR}
+    GIT_REPOSITORY https://github.com/pytorch/vision.git
+    GIT_TAG ${PYTORCH_VISION_COMMIT}
+    CMAKE_ARGS -DWITH_CUDA=${PYTORCH_USE_CUDA} -DCMAKE_PREFIX_PATH=${TORCH_LOCATION} -DCMAKE_INSTALL_PREFIX=${TORCHVISION_LOCATION}
+      "-DCMAKE_CXX_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/" "-DCMAKE_CUDA_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/"
+  )
+
+  list(APPEND TORCH_INC_DIR "${TORCHVISION_LOCATION}/include/")
+  list(APPEND TORCH_LIB_DIR "${TORCHVISION_LOCATION}/lib/")
```
sileht, line 768: Can you add this directory in `docker/get_libs.sh` too.

sileht (changes requested):

CMakeLists.txt
BynaryCobweb: After research this is a warning remover

CMakeLists.txt
BynaryCobweb: done

fantes (approved):

BynaryCobweb:
Added a test

beniz:
Great to see the vision factory.

src/backends/torch/native/native_factory.cc
```diff
@@ -55,22 +58,31 @@ namespace dd
...
-      const std::string tdef, const APIData template_params,
+      const std::string tdef, const APIData &template_params,
       const ImgTorchInputFileConn &inputc)
   {
-    (void)template_params;
     (void)inputc;
+
+    std::vector<std::string> p;
+    if (template_params.has("template_params"))
+      p = template_params.get("template_params")
```
fantes, line 68: what is p for ?

BynaryCobweb: It's no more used, will remove

BynaryCobweb: done

sileht (approved):

PR #955, opened 2020-11-09
fix(torch): Display correct remaining time and iter time

BynaryCobweb:


sileht:
@mergifyio rebase

fantes:
> @Mergifyio rebase

rooh c'est bô !

fantes:
> **Command `rebase`: success**
> 
> > **Branch already up to date**

ah oui  il est sensible dis donc :D

src/backends/torch/torchlib.cc
```diff
@@ -676,14 +681,30 @@ namespace dd
                     throw MLLibInternalException(
                         std::string("Libtorch error: ") + e.what());
                   }
+                tstop = system_clock::now();
```
sileht, line 684: To compute duration you must use `steady_clock::now()` that usually relies on many seconds elapsed since kernel boot. It only moves forward and never readjusted/changed.

`system_clock::now()` relies on system date, initialized with the bios date on boot and then periodically adjusted (with ntp for example). So if the adjustment occurs between two calls of 
`system_clock::now()` you got an unexpected duration, even negative duration sometimes.

fantes (approved):

src/backends/torch/torchlib.cc
BynaryCobweb: ok :+1:

sileht (approved):

PR #958, opened 2020-11-10
chore: follow-up 6d1aa4d

BynaryCobweb:


src/backends/torch/torchlib.cc
```diff
@@ -375,6 +375,7 @@ namespace dd
   int TorchLib<TInputConnectorStrategy, TOutputConnectorStrategy,
                TMLModel>::train(const APIData &ad, APIData &out)
   {
+    using namespace std::chrono;
```
sileht, line 378: left over ?

sileht: Ok you just move it, I think `using namespace` should be forbidden :p

BynaryCobweb: was there before. Its scope is restricted to only this function, to shorten calls to system_clock etc

sileht: I think it's better to use `using std::chrono::system_lock`, then, no?

PR #968, opened 2020-11-13
feat(torch): clip gradient options for all optimizers

fantes:
before this PR, only RANGER was clipping gradient if asked for, we now allow for clipping gradients (both to values and to norm) for all torch optimizers

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -690,8 +690,9 @@ iterations           | int          | yes      | N/A     | Max number of solver'
 snapshot             | int          | yes      | N/A     | Iterations between model snapshots
 snapshot_prefix      | string       | yes      | empty   | Prefix to snapshot file, supports repository
 solver_type          | string       | yes      | SGD     | from "SGD", "ADAGRAD", "NESTEROV", "RMSPROP", "ADADELTA", "ADAM",  "AMSGRAD", "RANGER", "RANGER_PLUS", "ADAMW", "SGDW", "AMSGRADW" (*W version for decoupled weight decay, RANGER_PLUS is ranger + adabelief + centralized_gradient)
-clip                 | bool         | yes      | false (true if RANGER* selected) | clip gradients, implemented only in ranger
-clip_value           | real         | yes      | 5.0     | value for  clipping gradients (used only by RANGER)
+clip                 | bool         | yes      | false (true if RANGER* selected) | gradients with absolute value greater than clip_value will be clipped to below values
+clip_value           | real         | yes      | 5.0     | gradients with absolute value greater than clip_value will be clipped to this value
+clip_norm            | real         | yes      | 100.0   | gradients with euclidean norm greater than clip_norm will be clipped to this value
 rectified            | bool         | yes      | false   | rectified momentum variance ie https://arxiv.org/abs/1908.03265 valid for ADAM[W] and AMSGRAD[W]
 adabelief            | bool         | yes      | false   | adabelief mod for ADAM https://arxiv.org/abs/2010.07468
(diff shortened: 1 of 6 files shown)


docs/api.md
```diff
@@ -690,8 +690,9 @@ iterations           | int          | yes      | N/A     | Max number of solver'
 snapshot             | int          | yes      | N/A     | Iterations between model snapshots
 snapshot_prefix      | string       | yes      | empty   | Prefix to snapshot file, supports repository
 solver_type          | string       | yes      | SGD     | from "SGD", "ADAGRAD", "NESTEROV", "RMSPROP", "ADADELTA", "ADAM",  "AMSGRAD", "RANGER", "RANGER_PLUS", "ADAMW", "SGDW", "AMSGRADW" (*W version for decoupled weight decay, RANGER_PLUS is ranger + adabelief + centralized_gradient)
-clip                 | bool         | yes      | false (true if RANGER* selected) | clip gradients, implemented only in ranger
-clip_value           | real         | yes      | 5.0     | value for  clipping gradients (used only by RANGER)
+clip                 | bool         | yes      | false (true if RANGER* selected) | clip gradients
+clip_value           | real         | yes      | 5.0     | value for  clipping gradients
```
BynaryCobweb, line 694: `gradients with absolute value greater than clip_value will be clipped` is more explicit and better imo

Same for `clip_norm`

tests/ut-torchapi.cc
```diff
@@ -987,3 +987,74 @@ TEST(torchapi, service_train_ranger)
...
+  ASSERT_EQ(created_str, joutstr);
+
+  // Train
+  std::string jtrainstr
+      = "{\"service\":\"imgserv\",\"async\":false,\"parameters\":{"
+        "\"mllib\":{\"solver\":{\"iterations\":"
+        + iterations_resnet50
+        + ",\"base_lr\":1e-5,\"iter_size\":4,\"solver_type\":\"RMSPROP\","
+          "\"clip\":true,\"test_interval\":"
+        + iterations_resnet50
```
BynaryCobweb, line 1013: Maybe we don't need as much iterations to test clipping? Could we reduce the number of iterations and just check that the loss is lower ?

src/backends/torch/torchsolver.cc
```diff
@@ -136,9 +133,38 @@ namespace dd
...
+        if (_clip_value >= 0.0)
+          this->_logger->info("clip_value: {}", _clip_value);
+        if (_clip_norm >= 0.0)
+          this->_logger->info("clip_norm: {}", _clip_value);
+        if (_clip_norm < 0.0 || _clip_value < 0.0)
+          {
+            this->_logger->warn(
+                "Gradient clipping selected, but no value given, "
+                "will clip gradient norm to {}",
+                DEFAULT_CLIP_NORM);
```
BynaryCobweb, line 151: Same thing for DEFAULT_CLIP_VALUE?

docs/api.md
fantes: ok done

src/backends/torch/torchsolver.cc
fantes: if nothing is given, currently code only clips norm of gradient, not value

BynaryCobweb: ok
but DEFAULT_CLIP_VALUE is never used in the code

tests/ut-torchapi.cc
fantes: done somthing

src/backends/torch/torchsolver.cc
fantes: if not defined, we now clip both absolute values and norm

BynaryCobweb (approved):

sileht (approved):

PR #969, opened 2020-11-13
chore: torch dataset refactored into its own C++ files

beniz:
This is a placeholder PR to be merged after #924 so that it eases rebasing before a larger set of changes take place, i.e. data augmentation, reading images from lists, ...


src/backends/torch/torchinputconns.h
```diff
@@ -32,9 +32,10 @@
 
 #include "imginputfileconn.h"
 #include "txtinputfileconn.h"
-#include "backends/torch/db.hpp"
-#include "backends/torch/db_lmdb.hpp"
+/*#include "backends/torch/db.hpp"
```
BynaryCobweb, line 35: Remove if not needed anymore

BynaryCobweb (approved):

src/backends/torch/torchinputconns.h
sileht: :+1:

beniz: done

PR #975, opened 2020-11-13
feat: introduce aotpp interface for deepdetect

sileht:


fantes (approved):

src/http/app_component.hpp
```diff
@@ -0,0 +1,131 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2020 Jolibrain SASU
+ * Author: Mehdi Abaakouk <mabaakouk@jolibrain.com>
```
beniz, line 4: Email to fix I guess, that's all :)

sileht: fixed

beniz (approved):

PR #976, opened 2020-11-15
fix: tensorrt refinedet CI and linking to nvcaffeparsers with tensorrt OSS

beniz:


beniz:
@sileht this seems to not work from within the TensorRT docker image ( cc @YaYaB ), any idea ?

YaYaB:
https://github.com/jolibrain/deepdetect/blob/master/docker/gpu_tensorrt.Dockerfile#L110
All the dockerfiles seem to use multi stage building. Thus we need to copy the built library that are linked to DD using the script https://github.com/jolibrain/deepdetect/blob/master/docker/gpu_tensorrt.Dockerfile#L107.

However the libs associated to tensorrt's build do not seem to be inserted in here.
This i why it might be necessary to add the following in the file https://github.com/jolibrain/deepdetect/blob/master/docker/get_libs.sh#L14 
```
/opt/deepdetect/build/tensorrt-oss/bin/*
```

However note that you might encounter an issue with the order of the linked library discovery (https://unix.stackexchange.com/questions/367600/what-is-the-order-that-linuxs-dynamic-linker-searches-paths-in)
Updating the `LD_LIBRARY_PATH` with `/usr/lib` should prevent ld to search the tensorrt-oss libraries in `/usr/lib/x86_64-linux-gnu/`

sileht:
@mergifyio rebase

sileht:
We currently build the tensorrt docker image based on the nvidia tensorrt docker image.
This image contains a already build version of tensorrt. But with build deepdetect with tensorrt_oss  (`-USE_TENSORRT_OSS=ON`).

The tensorrt libs look not compatible with each other.

That explains why the @YaYaB workaround work, as it forces the linker to load tensort_oss before tensort.

So I see two solutions:
* We build deepdetect with the tensorrt version provided in the nvidia docker image (eg `USE_TENSORRT_OSS=OFF`)
* We build deepdetect based on the nvidia cuda image and copy tensorrt_oss libs with other libs (eg get_libs.sh)
   (I have to check first all USE_TENSORRT_OSS deps are available)

sileht:
tensorrt-oss need some bits from tensorrt-ga, so we have to build them together.

I found another difference we use tensorrt-oss 20.09 with tensorrt 20.03, I wonder that can lead to some issue.

beniz:
Thanks @sileht. I believe we should use TENSORRT_OSS as default, and deprecated the other. Historically NVidia did OSS while we already had integrated the fully poprietary version. The OSS allows us to patch the Caffe parser and make RefineDet work with TensorRT.

beniz:
Also, maybe we could make the TensoRT / OSS + RefineDet docker problem into its own issue and merge this current fix to the regular build.

sileht:
+1 to this, it's required and has tests I need :D

sileht:
I have some patches to cleanup the tensorrt build and fix the refinedet issue.

PR #977, opened 2020-11-15
Support for torch training by reading large image datasets directly from lists of files and targets

beniz:
~~This is placeholder PR that depends on #924  and #969 and will be rebased after these PRs have been merged.~~

It implements reading data from files. This is complicated by the dependance on input connector `DDImg` on once side and the virtual function `get_batch` from `torch::Dataset` on the other side.

EDIT: this is part of an update to `TorchDataset::get_batch` and images db support with the torch backend.

### DIFF
--- a/src/backends/torch/torchdataset.cc
+++ b/src/backends/torch/torchdataset.cc
@@ -142,10 +142,21 @@ namespace dd
     if (!_db)
       {
         _indices.clear();
-
-        for (unsigned int i = 0; i < _batches.size(); ++i)
+        if (!_lfiles.empty()) // list of files
           {
-            _indices.push_back(i);
+            _indices = std::vector<int64_t>(_lfiles.size());
+            std::iota(std::begin(_indices), std::end(_indices), 0);
+          }
+        else if (!_batches.empty())
+          {
+            for (unsigned int i = 0; i < _batches.size(); ++i)
+              {
+                _indices.push_back(i);
+              }
+          }
+        else
+          {
+            //
           }
       }
     else // below db case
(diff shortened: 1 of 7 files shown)


src/backends/torch/torchdataset.cc
```diff
@@ -311,4 +372,131 @@ namespace dd
...
+        add_batch({ imgt }, { targett });
+        return 0;
+      }
+    else
+      {
+        return -1;
+      }
+  }
+
+  at::Tensor TorchDataset::image_to_tensor(const cv::Mat &bgr,
```
fantes, line 447: `image_to_tensor` is a classical method because it uses attributes of `_inputc` (mean, scale std...),
=>  so you need to add `_inputc` as an attribute of dataset so that `image_to_tensor` can use all these. 

I wonder of it would not be simpler to make `image_to_tensor` function a generic util function ie `torch_utils::image_to_tensor`, and passing all required parameters explicitly to it would allow not to  put `inputc` as an attribute of dataset and would make more explicit what is needed to transform an image to a tensor

(or if you do not want to add all parameters, just pass a reference to inputc; ideally it should be something in between : a struct in the inputc that contains the transformation parameters needed )

src/backends/torch/torchdataset.h
```diff
@@ -63,10 +68,15 @@ namespace dd
   public:
     std::shared_ptr<db::DB> _dbData; /**< db data */
     std::vector<int64_t> _indices;   /**< id/key  of data points */
+    std::vector<std::pair<std::string, std::vector<double>>>
+        _lfiles; /**< list of files */
 
     std::vector<TorchBatch> _batches; /**< Vector containing the whole dataset
                                          (the "cached data") */
     std::string _dbFullName;          /**< db filename */
+    InputConnectorStrategy *_inputc
```
fantes, line 77: linked to comments above

src/backends/torch/torchinputconns.h
```diff
@@ -190,6 +190,8 @@ namespace dd
      */
     ImgTorchInputFileConn() : ImgInputFileConn()
     {
+      _dataset._inputc = this;
```
fantes, line 193: again, same as above

src/backends/torch/torchinputconns.h
```diff
@@ -199,6 +201,8 @@ namespace dd
     ImgTorchInputFileConn(const ImgTorchInputFileConn &i)
         : ImgInputFileConn(i), TorchInputInterface(i)
     {
+      _dataset._inputc = this;
```
fantes, line 204: again again, same as above above

src/backends/torch/torchinputconns.cc
```diff
@@ -306,18 +308,26 @@ namespace dd
...
-                for (const std::pair<std::string, std::vector<double>> &lfile :
-                     lfiles)
+                if (_db)
                   {
-                    add_image_file<std::vector<double>>(_dataset, lfile.first,
-                                                        lfile.second);
+                    for (const std::pair<std::string, std::vector<double>>
+                             &lfile : lfiles)
+                      {
+                        _dataset.add_image_file(lfile.first, lfile.second,
```
fantes, line 316: linked to same comment : here you pass `_height` and `_width` from this (inputc) , but adding image needs other params from `inputc` that will be read through  `_dataset._inputc`

src/backends/torch/torchdataset.cc
```diff
@@ -311,4 +372,131 @@ namespace dd
...
+        add_batch({ imgt }, { targett });
+        return 0;
+      }
+    else
+      {
+        return -1;
+      }
+  }
+
+  int TorchDataset::add_image_file(const std::string &fname,
```
fantes, line 411: linked to comment below: here you pass _height and _width from upper inputc , but adding image needs other params from inputc that will be read through _dataset._inputc

src/backends/torch/torchdataset.cc
beniz: Unfortunately `inputc` is needed in forthcoming PRs when encoded images are stored in dbs (instead of tensors). This is because `DDimg` and `inputc->copy_parameters_to` are then needed. Agreed that this cannot be seen from this PR alone.

There are other forthcoming issues of this kind coming up.

Unless you and others think otherwise, I do favor to get it working then rationalize / simplify as best as we can once all features are in. 

Basically have dataset on the side of the input connector does run us into unexpected issues.

fantes (approved):

sileht (approved):

PR #979, opened 2020-11-16
fix(torch/native) : do not fail if trying to load weights before (lat…

fantes:
…e) allocation of native model


sileht (approved):

fantes:
nbeats/ prediction simply does not work without this patch

csv/csvts API will be cleanup up later so that model is really allocated in init_mllib and not at beginning of train

beniz:
FTR, as discussed elsewhere this is a required fix at the moment, though we may want to enforce full creation within `init_mllib` at service creation, with slight adjustements to the API.

beniz (approved):

PR #988, opened 2020-11-17
feat(torch): implement resume mllib option for torchlib: if true, reu…

fantes:
…se previous solver state

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -772,6 +772,7 @@ gradient_centralization | bool | yes | false for RANGER, true for RANGER_PLUS| f
 test_interval | int    | yes      | N/A     | Number of iterations between testing phases
 base_lr       | real   | yes      | N/A     | Initial learning rate
 iter_size     | int    | yes      | 1       | Number of passes (iter_size * batch_size) at every iteration
+resume        | bool   | yes      | false   | Whether to resume training from solver state 
 
 Net:
 
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -346,7 +346,7 @@ namespace dd
                    TMLModel>::save_if_best(APIData &meas_out,
                                            int64_t elapsed_it,
                                            TorchSolver &tsolver,
-                                           int64_t best_to_remove)
+                                           int64_t best_iteration_number)
   {
     double cur_meas = std::numeric_limits<double>::infinity();
     std::string meas;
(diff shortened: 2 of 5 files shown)


sileht (approved):

docs/api.md
```diff
@@ -772,6 +772,7 @@ gradient_centralization | bool | yes | false for RANGER, true for RANGER_PLUS| f
 test_interval | int    | yes      | N/A     | Number of iterations between testing phases
 base_lr       | real   | yes      | N/A     | Initial learning rate
 iter_size     | int    | yes      | 1       | Number of passes (iter_size * batch_size) at every iteration
+resume        | bool   | yes      | false   | Whether to resume training from solverstate
```
BynaryCobweb, line 775: With torch backend it's not called "solverstate" like in caffe

src/backends/torch/torchsolver.h
```diff
@@ -70,6 +70,14 @@ namespace dd
      */
     void save(std::string sfile);
 
+    /**
+     * \brief restore solver state, checks solverstate presecnce  and returns
```
BynaryCobweb, line 74: presence
\+ reload best metrics

src/backends/torch/torchlib.cc
```diff
@@ -531,9 +531,8 @@ namespace dd
     // create solver
     tsolver.create(_module);
 
-    int it = 0;
-    // reload solver and set it value accordingly
-    it = tsolver.load(this->_mlmodel._sstate, _main_device);
+    int it = tsolver.resume(ad_mllib, this->_mlmodel, _main_device,
```
BynaryCobweb, line 534: Need to reload `best_to_remove` as well ?

BynaryCobweb: Looking at the code it seems that `best_to_remove` is always equal to -1

docs/api.md
fantes: replaced with "solver state" ;)

src/backends/torch/torchsolver.h
fantes: ok

src/backends/torch/torchlib.cc
fantes: > Need to reload `best_to_remove` as well ?

nice one, I did not remember the existence of this one

fantes: > Looking at the code it seems that `best_to_remove` is always equal to -1

It should change, it is updated by save_if_best (changed if best)

BynaryCobweb: ok I see

BynaryCobweb (approved):

PR #992, opened 2020-11-18
chore: make sure model files don't remain in torch training unit tests

beniz:
This fixes the broken cleanup of directories in torch unit tests.

### DIFF
--- a/tests/ut-torchapi.cc
+++ b/tests/ut-torchapi.cc
@@ -200,12 +200,18 @@ TEST(torchapi, service_train_images_split)
       << "accuracy good";
   ASSERT_TRUE(jd["body"]["measure"]["f1"].GetDouble() <= 1) << "f1";
 
-  fileops::remove_file(resnet50_train_repo,
-                       "checkpoint-" + iterations_resnet50 + ".ptw");
-  fileops::remove_file(resnet50_train_repo,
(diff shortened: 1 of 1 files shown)


tests/ut-torchapi.cc
```diff
@@ -200,12 +200,18 @@ TEST(torchapi, service_train_images_split)
...
+  for (std::string ff : lfiles)
+    {
+      if (ff.find("checkpoint") != std::string::npos
+          || ff.find("solver") != std::string::npos)
+        remove(ff.c_str());
+    }
+  ASSERT_TRUE(!fileops::file_exists(resnet50_train_repo + "checkpoint-"
+                                    + iterations_resnet50 + ".ptw"));
+  ASSERT_TRUE(!fileops::file_exists(resnet50_train_repo + "checkpoint-"
+                                    + iterations_resnet50 + ".pt"));
```
BynaryCobweb, line 214: Any reason for checking only checkpoint and not solver ?

beniz: no, I did assume this was enough to check that the remove loop was fine...

BynaryCobweb: ok

BynaryCobweb (approved):

PR #993, opened 2020-11-19
feat: print stacktrace on segfault

sileht:


CMakeLists.txt
```diff
@@ -97,7 +97,6 @@ if(RELEASE)
     set(BUILD_TYPE release)
 else()
     set(BUILD_TYPE dev)
-    string(APPEND CMAKE_CXX_FLAGS " -g")
```
BynaryCobweb, line 100: Removed because not needed? (just asking)

BynaryCobweb (approved):

fantes (approved):

CMakeLists.txt
sileht: It was passed twice.

PR #995, opened 2020-11-20
feat(torch/timeseries): add (p)ReLU layers in recurrent template,

fantes:
allowing to compute mlp-like embeddings before LSTM layers

syntax `layers : ["AR5","AR5", "A5", "L50", "L50"]`  
`AR` for affine + relu, `A` for affine only
in torch, it will be implemented as PReLU, ie leaky relu with learning of leak param


BynaryCobweb (approved):

beniz:
image

beniz (approved):

PR #999, opened 2020-11-23
build: don't git clone manually

sileht:
cmake was confused by external project directory that was not generated
by itself. If the external project need to be recreated it was raising
an error instead of overriding the directory.

This fixes that. But also makes the CI more robust against temporary
network issue, because cmake automatically retries if download fail,


fantes (approved):

BynaryCobweb (approved):

sileht:
@mergifyio update

PR #1000, opened 2020-11-23
feat: gives also per target error and not only global eucl when selec…

fantes:
…ting measure:eucll


src/supervisedoutputconnector.h
```diff
@@ -1053,11 +1053,29 @@ namespace dd
             }
           if (beucll)
             {
-              double meucll = eucll(ad_res, -1);
+              std::tuple<double, std::vector<double>> tmeucll
+                  = eucll(ad_res, -1);
+              double meucll = std::get<0>(tmeucll);
```
BynaryCobweb, line 1058: FYI possible syntax
```cpp
double meucll;
std::vector<double> all_meucll;
std::tie(meucll, all_meucll) = eucll(ad_res, -1);
```
C++ tuple unwrapping :) https://en.cppreference.com/w/cpp/utility/tuple/tie

BynaryCobweb (approved):

src/supervisedoutputconnector.h
fantes: ok, it is much nicer :)

beniz: image

sileht:
Commit title is a bit long

sileht (approved):
1000eme pr! 🥇

PR #1002, opened 2020-11-23
feat: add some more error messages to log

fantes:
some of these messages were not output anywhere, for instance in jupyter


src/mlmodel.h
```diff
@@ -219,15 +219,19 @@ namespace dd
       bool isDir;
       bool exists = fileops::file_exists(_repo, isDir);
       if (exists && !isDir)
-        throw MLLibBadParamException(
-            "file exists with same name as repository");
+        {
+          std::string errmsg = "file exists with same name as repository";
+          logger->error(errmsg);
```
sileht, line 224: Not sure this should be an `error`. I think it should be `debug` or `info`.

The (discussable) rule of thumb:
* `error` for unexpected behavior
* `warning` for strange/unexpected behavior that we deal with or ignore
* `info` track important step of the ongoing request/process to understand where we are
* `debug` anything else help to debug the software

Another view that helps to decide where to put a message:
* `error` is for the administrator of the platform if they see them, they will open an issue
* `warning` is for the administrator of the platform if they see them, they may open an issue
* `info`  should be useful for administrator and developer
* `debug` should be for the developer only

logs are not for end-users or API-users, they already have errors reported via the API itself (eg 4XX, 5XX HTTP errors)

sileht (commented):
Maybe what we really need is to log all 4XX errors as info in logs, no?

fantes:
> Maybe what we really need is to log all 4XX errors as info in logs, no?

that would be good, but I do not know how/where to to it

src/mlmodel.h
beniz: Understood, very useful @sileht. It may be interesting to decide for the above and apply it throughout the code as another PR since until now we haven't used these semantics correctly. Typically we've used `error` for most user bad inputs etc...

As you suggest elsewhere we may need to log API related exceptions/errors differently if we decide to not use the `error` log typically.

So maybe we could move the change of semantics to a different PR and schedule the task (I'm actually interested in doing it).

sileht: I created a task.

sileht:
> > Maybe what we really need is to log all 4XX errors as info in logs, no?
> 
> that would be good, but I do not know how/where to to it

I think we can create an oatpp hook to do that, I created a card to do that later

beniz (approved):

BynaryCobweb (approved):

PR #1019, opened 2020-11-25
refactor(graph): much saner namespaces / class names

fantes:
this PR cleans up namespace around graph, prepares to some deeper refactor around operators for easier code extendability


beniz (approved):

BynaryCobweb:
To me it looks strange to have a class named "Base". Even in the graph namespace, "Base" could refer to multiple things, in my opinion. In the graph code, ie inside of the graph namespace, the name "Base" is not very self explanatory

fantes:
> To me it looks strange to have a class named "Base". Even in the graph namespace, "Base" could refer to multiple things, in my opinion. In the graph code, ie inside of the graph namespace, the name "Base" is not very self explanatory

I agree, the idea was that graph::Base is clear, but Base used alone when in namespace is weird. Do you have any idea ?

BynaryCobweb:
> > To me it looks strange to have a class named "Base". Even in the graph namespace, "Base" could refer to multiple things, in my opinion. In the graph code, ie inside of the graph namespace, the name "Base" is not very self explanatory
> 
> I agree, the idea was that graph::Base is clear, but Base used alone when in namespace is weird. Do you have any idea ? graph::BaseGraph is not really completely satisfying either

Actually I prefer to keep graph::BaseGraph. I don't think the redundancy is a problem here.

Another option would be to call it `BaseBackend`, since we have `TorchGraphBackend` (that could be renamed into TorchBackend`?). Actually anything that is not just "Base" is fine (graph::BaseModule? ...)

sileht:
What about `dd::graph::core` ?

BynaryCobweb:
> What about `dd::graph::core` ?
> 
> mmmm same pb here : core alone is not clear enough... @BynaryCobweb what do you think ?

dd::graph::Base is a class, so it would be dd::graph::Core?

sileht:
I was seeing a class hierarchy like:
```
dd::graph::core:Graph
dd::graph::core:Vertex
dd::graph::core:VertexProperty
dd::graph::core:Edge


class TorchGraphBackend : public virtual core::Graph, public torch::nn::Module
```

BynaryCobweb (approved):

PR #1026, opened 2020-11-27
build(tensorrt): use same protobuf version as other parts

sileht:


sileht:
tensorrt like only protobuf 3.0 ...

PR #1035, opened 2020-11-30
fix(torch): when resume, override solver parameters after solver reload

fantes:
this PR allows correct overriding of previous solver params when resuming. Previously, saved solver parameters were used, not the new ones from the  API .


BynaryCobweb (approved):

sileht (approved):

sileht:
tests?

PR #1044, opened 2020-12-01
fix(csvts): clarify error msg on seq lengths

fantes:


src/backends/torch/torchinputconns.cc
```diff
@@ -807,12 +807,19 @@ namespace dd
...
+      errmsg = "data does not contains enough timesteps, "
+               "discarding (seq_size: "
+               + std::to_string(seq_size)
+               + "  timesteps: " + std::to_string(_timesteps) + " )";
+    else
+      errmsg = "data does not contains enough timesteps, "
+               "discarding (seq_size: "
+               + std::to_string(seq_size)
+               + "  backcast_timesteps: " + std::to_string(_backcast_timesteps)
+               + +"   forecast_timesteps: "
```
BynaryCobweb, line 821: I think one of the "+" should be removed

fantes: thx !

beniz (approved):

BynaryCobweb (approved):

PR #1045, opened 2020-12-01
feat: custom & improved torch multi-gpu

beniz:
This PR modifies libtorch default C++ data parallel multi-gpu distribution mode with the following speed-ups:
~- do not deep clone modules if source and target devices are the same~
- only deep clone modules on the first multi-gpu iteration
- do not recopy weights across gpus when using multiple forward passes over multiple batches (using `iter_size`).

The C++ libtorch data parallel implementation is notoriously slow, see https://github.com/pytorch/pytorch/issues/40581 and this PR speeds things up a bit. 

Next step would be a pure C++ DDP implementation.

Note: this PR patches pytorch 1.6 to make `clone_` and other module's inner members accessible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -713,6 +713,7 @@ if (USE_TORCH)
     ${PYTORCH_PATCHES_PATH}/pytorch_16_logging.patch
     ${PYTORCH_PATCHES_PATH}/pytorch_14_namespace.patch
     ${PYTORCH_PATCHES_PATH}/pytorch_16_fatbin.patch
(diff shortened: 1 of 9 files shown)


linjiapro:
Thanks Emmanuel!

beniz:
I am not happy with this PR as it is hacky and the final performances are not satisfactory. It would nee to be reimplemented from scratch, as suggested by https://github.com/pytorch/pytorch/issues/40581#issuecomment-739600368 or we shall wait for https://github.com/pytorch/pytorch/issues/48959

So I'd like to close this PR.

PR #1047, opened 2020-12-01
fix(nbeats): fix loss for backcast part

fantes:
last backcast output is residual error on backcast reconstruction, and thus should be encouraged to be zero


sileht:
@mergifyio update

sileht (approved):

sileht:
tests ?

beniz (approved):

fantes:
> tests ?

same as before, if they worked they will work even better :D

PR #1063, opened 2020-12-04
refactor(torch): change parameter initialization order for clarity

BynaryCobweb:
Prior to this refactor, module initialization code was mixed up with
mllib configuration, which makes the code very unclear I think.
    
Here is an attempt to make things better, by separating initialization
of configuration variables like _classification, _regression, _timeserie
etc., initialization of mltype, and model configuration (_module). This
has the benefit of making the API behavior clearer, since we can
understand the API from the "configuration variables" part only.


fantes (approved):

src/backends/torch/torchlib.cc
```diff
@@ -244,91 +289,59 @@ namespace dd
     if (unsupported_model_configuration)
       throw MLLibInternalException("Use of libtorch backend needs either: "
                                    "traced net, protofile or native template");
-    // Create the model
+
+    // FIXME(louis): out of if(bert) because we allow not to specify template
+    // at predict. Should we change this?
     this->_inputc._input_format = "bert";
```
beniz, line 295: default input format is `bert` ? This is weird indeed, though I do understand it's been so "historically".

beniz (approved):

src/backends/torch/torchlib.cc
BynaryCobweb: @sileht made a PR about this some time ago #893, I don't know why it stayed like this until now

PR #1064, opened 2020-12-04
fix(torch/timeseries): restore previous behavior wrt timesteps API pa…

fantes:
…ram : timestep not used at predict time


src/backends/torch/torchinputconns.cc
```diff
@@ -905,31 +935,15 @@ namespace dd
...
-        std::vector<double> labelvec;
 
-        for (unsigned int li = 0; li < label_size; ++li)
-          labelvec.push_back(seq[ti]._v[_label_pos[li]]);
-        for (int di = 0; di < this->_datadim; ++di)
-          if (std::find(_label_pos.begin(), _label_pos.end(), di)
-              == _label_pos.end())
-            datavec.push_back(seq[ti]._v[di]);
+    if (_train)
+      for (unsigned long int ti = tstart;
```
BynaryCobweb, line 940: use size_t instead

fantes: ok, done

src/backends/torch/torchinputconns.cc
```diff
@@ -879,6 +879,37 @@ namespace dd
...
+  void
+  CSVTSTorchInputFileConn::add_seq(const unsigned long int ti,
+                                   const std::vector<CSVline> &seq,
+                                   std::vector<at::Tensor> &data_sequence,
+                                   std::vector<at::Tensor> &label_sequence)
+  {
+    std::vector<double> datavec;
+    std::vector<double> labelvec;
+    unsigned int label_size = _label_pos.size();
+    unsigned int data_size = _datadim - label_size;
```
BynaryCobweb, line 891: use `size_t` instead of `unsigned int`

BynaryCobweb: ti can also be a size_t

src/backends/torch/torchinputconns.cc
```diff
@@ -942,22 +956,31 @@ namespace dd
...
+              {
+                discard_warn(vecindex, seq.size(), test);
+                continue;
+              }
+            for (; tstart + _timesteps < static_cast<long int>(seq.size());
+                 tstart += _offset)
+              add_data_instance_labels(tstart, vecindex, dataset, seq);
+            if (tstart < static_cast<long int>(seq.size()) - 1)
+              add_data_instance_labels(seq.size() - _timesteps, vecindex,
+                                       dataset, seq);
```
BynaryCobweb, line 975: So if sequence length is 500, timestep is 200 and offset is 200, the training sequences will be [0, 200], [200, 400] and [300, 500] ? The last sequence seems arbitrary, maybe this behavior should be referenced somewhere?

src/backends/torch/torchinputconns.cc
```diff
@@ -905,31 +935,15 @@ namespace dd
...
-        std::vector<double> datavec;
-        std::vector<double> labelvec;
 
-        for (unsigned int li = 0; li < label_size; ++li)
-          labelvec.push_back(seq[ti]._v[_label_pos[li]]);
-        for (int di = 0; di < this->_datadim; ++di)
-          if (std::find(_label_pos.begin(), _label_pos.end(), di)
-              == _label_pos.end())
-            datavec.push_back(seq[ti]._v[di]);
+    if (_train)
```
BynaryCobweb, line 939: `if (_train)` could be removed if we pass the length of the sequence as an argument to `add_data_instance_labels`

src/backends/torch/torchinputconns.cc
fantes: this is the best way I could think of for taking into account all data (ie do not discard [400, 500] in your example). I added a sentence in api.md

docs/api.md
```diff
@@ -680,7 +680,7 @@ resume        | bool           | yes      | false          | Whether to resume t
 class_weights | array of float | yes      | 1.0 everywhere | Whether to weight some classes more / less than others, e.g. [1.0,0.1,1.0]
 ignore_label  | int            | yes      | N/A            | A single label to be ignored by the loss (i.e. no gradients)
 timesteps     | int            | yes      | N/A            | Number of timesteps for recurrence ('csvts', `ctc` OCR) models
```
BynaryCobweb, line 682: Timesteps are used only at training time

BynaryCobweb:
I wonder what happens if we give a negative "timesteps" (overflow?) or no timesteps at training ?

fantes:
clarified doc

BynaryCobweb (approved):

beniz (approved):

PR #1069, opened 2020-12-08
feat(tools): trace_torchvision can trace models for loading weights w…

BynaryCobweb:
…ith dd native models


fantes (approved):

tools/torch/trace_torchvision.py
```diff
@@ -96,6 +110,11 @@
...
     logging.info("Exporting model %s %s", mname, "(pretrained)" if args.pretrained else "")
     model = model_classes[mname](pretrained=args.pretrained, progress=args.verbose)
+
+    if args.to_dd_native:
+        # Make model NativeModuleWrapper compliant
+        model = Wrapper(model)
+
     model.eval()
 
     example = torch.rand(1, 3, 224, 224)
```
beniz, line 120: We may want these values as default to input parameters.

beniz (approved):

PR #1070, opened 2020-12-08
feat(torch): native models can load weights from any jit file

BynaryCobweb:
Added a function "copy_weights" to copy common weights from a
script module to a nn::Module. Weights that do not match are
ignored, and printed to the logs as warning.

### DIFF
--- a/src/backends/torch/torchmodule.cc
+++ b/src/backends/torch/torchmodule.cc
@@ -122,7 +122,8 @@ namespace dd
         _logger->info("loading " + tmodel._native);
         try
           {
-            torch::load(_native, tmodel._native, _device);
+            torch_utils::load_weights(*_native, tmodel._native, _device,
+                                      _logger);
           }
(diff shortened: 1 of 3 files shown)


fantes (approved):

beniz:
Great, though I'd have weights that do not match generate an error instead, what do you think ?

fantes:
> Great, though I'd have weights that do not match generate an error instead, what do you think ?

In case of finetuning, maybe that should not be an error ?

BynaryCobweb:
> > Great, though I'd have weights that do not match generate an error instead, what do you think ?
> 
> In case of finetuning, maybe that should not be an error ?

The point of this change is that we can load only some of the weights when finetuning.
I can add a parameter so that we throw when not finetuning, and we don't throw otherwise

sileht (approved):

BynaryCobweb:
@Mergifyio update

PR #1074, opened 2020-12-09
fix: simsearch build with annoy

beniz:


fantes (approved):

beniz:
> Since annoy is an external libs, we can set SYSTEM flag in cmake
> 
> here: https://github.com/jolibrain/deepdetect/blob/master/CMakeLists.txt#L295
> 
> ```
> include_directories(SYSTEM "${ANNOY_INCLUDE_DIR}")
> ```

Thanks, kind of black magic to me :) Added it and removed the GCC pragmas.

BynaryCobweb (approved):

PR #1083, opened 2020-12-15
refactor: make logger torch-friendly

fantes:
This PR
-  changes implementation of CaffeLogger so that it can be used directly from classes that implement `std::ostream& operator<<(std::ostream& stream, some_class&);`  (see https://github.com/jolibrain/caffe/pull/80 for implementation of the logger only)
- simplifies _a lot_ pytorch patching, hopefully leading to much easier libtorch update to upstream (see https://github.com/jolibrain/deepdetect/pull/1086)
- should not  change its API except for accepting above usage (ie `LOG(INFO) << "message" << some_class_that_implements_above_operator_like_many_in_libtorch;`)
- should not change its behavior elsewhere in dd

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,7 +710,8 @@ if (USE_TORCH)
 
   set(PYTORCH_PATCHES
     ${PYTORCH_PATCHES_PATH}/pytorch_15_compile.patch
-    ${PYTORCH_PATCHES_PATH}/pytorch_16_logging.patch
+    ${PYTORCH_PATCHES_PATH}/pytorch_16_new_logger.patch
+    ${PYTORCH_PATCHES_PATH}/pytorch_16_use_new_logger.patch
     ${PYTORCH_PATCHES_PATH}/pytorch_14_namespace.patch
     ${PYTORCH_PATCHES_PATH}/pytorch_16_fatbin.patch
   )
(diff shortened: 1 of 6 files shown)


src/backends/torch/llogging.h
```diff
@@ -191,108 +191,57 @@ class CaffeLogger
...
-  (void)severity;
-  return CaffeLogger("none");
-}
-inline CaffeLogger DLOG(const std::string &severity)
-{
-  (void)severity;
-  return CaffeLogger("none");
-}
+
+#define DFATAL(severity) CaffeLogger("none").sstream()
```
BynaryCobweb, line 229: I was fearing that Caffelogger would be destroyed before logging anything, but I tested on a toy example and it seems like it is not the case.

Standard says
```
The destructor is called whenever an object's lifetime ends, which includes 
...
end of the *full* expression, for nameless temporaries
```
So it's fine, and I learnt something :slightly_smiling_face:

src/backends/torch/torchutils.h
```diff
@@ -35,6 +35,8 @@
 #pragma GCC diagnostic pop
 #include <torch/script.h>
 
+#include <google/protobuf/message.h>
```
BynaryCobweb, line 38: why this?

CMakeLists.txt
```diff
@@ -383,8 +383,8 @@ if (USE_DLIB)
     message(STATUS "Dlib will be built")
 endif()
 
-set(CAFFE_DD_USER jolibrain)
-set(CAFFE_DD_BRANCH master)
+set(CAFFE_DD_USER fantes)
```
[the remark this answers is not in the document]
fantes: will do as soon as https://github.com/jolibrain/caffe/pull/80 is merged

src/backends/torch/llogging.h
fantes: indeed it is exaclty what is done in torch logger when not using glog, so I hypothesized it works like this :)

src/backends/torch/torchutils.h
fantes: do not know why it triggered here:  in case of separate compilation I got sometimes a "google::protobuf::Message" not declared ...

CMakeLists.txt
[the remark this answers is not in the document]
BynaryCobweb: ok

src/backends/torch/torchutils.h
BynaryCobweb: Strange... Where is the error located? It may be in a file that includes a file that itself includes torchutils.h or something like that

fantes: yes, from something-json-something if I remember correctly

CMakeLists.txt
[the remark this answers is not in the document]
beniz: Merged https://github.com/jolibrain/caffe/pull/80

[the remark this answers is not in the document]
beniz: If we can stay with `jolibrain` user that might be better though.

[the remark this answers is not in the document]
fantes: done

src/backends/torch/torchutils.h
fantes: but anywat as google::protobuf::message is used just below, I think it is a good idea to include it here :)

BynaryCobweb (approved):

beniz (approved):

PR #1095, opened 2020-12-17
Multiple test sets

fantes:
this PR adds support for multiple test sets for torch backend.
example call:
```
data: ["path/to/train/set", "path/to/first/test/set", "path/to/second/test/set", ....]
```
support for csvts, img, txt (ie everything supported by torch mllib atm)

### DIFF
--- a/src/backends/caffe/caffeinputconns.cc
+++ b/src/backends/caffe/caffeinputconns.cc
@@ -611,6 +611,13 @@ namespace dd
         write_images_to_hdf5(img_lists.at(1), test_dbfullname, test_list,
                              alphabet, max_ocr_length, false);
       }
+    if (img_lists.size() > 2)
+      {
+        _logger->error(
+            "multiple test sets not supported by caffe backend yet");
+        throw InputConnectorBadParamException(
+            "multiple test sets not supported by caffe backend yet");
+      }
 
     // save the alphabet as corresp file
     std::ofstream correspf(_model_repo + "/" + _correspname, std::ios::binary);
@@ -1341,8 +1348,9 @@ namespace dd
   }
 
   /*- DDCCsv -*/
-  int DDCCsv::read_file(const std::string &fname)
+  int DDCCsv::read_file(const std::string &fname, int test_id)
   {
+    (void)test_id;
     if (_cifc)
       {
         _cifc->read_csv(fname);
@@ -1483,7 +1491,8 @@ namespace dd
   {
     if (!_db)
       {
-        CSVInputFileConn::add_test_csvline(id, vals);
+        // MULTIPLE TEST SETS : we consider here only 1 test set
+        CSVInputFileConn::add_test_csvline(0, id, vals);
         return;
       }
 
@@ -1535,8 +1544,9 @@ namespace dd
     if (!fileops::file_exists(_csv_fname))
       throw InputConnectorBadParamException("training CSV file " + _csv_fname
                                             + " does not exist");
+    // MULTIPLE TEST SETS : we consider here only 1 test set
     if (_uris.size() > 1)
-      _csv_test_fname = _uris.at(1);
+      _csv_test_fnames.push_back(_uris.at(1));
     /*if (ad_input.has("label"))
       _label = ad_input.get("label").get<std::string>();
     else if (_train && _label.empty()) throw
@@ -1644,13 +1654,27 @@ namespace dd
         // XXX: remove in-memory data, which pre-processing is useless and
         // should be avoided
         destroy_txt_entries(_txt);
-        destroy_txt_entries(_test_txt);
+        // MULTIPLE TEST SETS : we consider here only 1 test set
+        //        destroy_txt_entries(_test_txt);
+        for (auto tt : _tests_txt)
+          destroy_txt_entries(tt);
+        _tests_txt.clear();
 
         return 0;
       }
 
     _db_batchsize = _txt.size();
-    _db_testbatchsize = _test_txt.size();
+    // MULTIPLE TEST SETS : we consider here only 1 test set
+
+    if (_tests_txt.size() > 1)
+      {
+        _logger->error(
+            "multiple test sets not supported by caffe backend yet");
+        throw InputConnectorBadParamException(
+            "multiple test sets not supported by caffe backend yet");
+      }
+
+    _db_testbatchsize = _tests_txt[0].size();
 
     _logger->info("db_batchsize={} / db_testbatchsize={}", _db_batchsize,
                   _db_testbatchsize);
@@ -1661,13 +1685,19 @@ namespace dd
     else
       write_sparse_txt_to_db(dbfullname, _txt);
     destroy_txt_entries(_txt);
-    if (!_test_txt.empty())
+    // MULTIPLE TEST SETS : we consider here only 1 test set
+    if (!_tests_txt.empty() && !_tests_txt[0].empty())
       {
         if (!_sparse)
-          write_txt_to_db(testdbfullname, _test_txt);
+          // MULTIPLE TEST SETS : we consider here only 1 test set
+          write_txt_to_db(testdbfullname, _tests_txt[0]);
         else
-          write_sparse_txt_to_db(testdbfullname, _test_txt);
-        destroy_txt_entries(_test_txt);
+          // MULTIPLE TEST SETS : we consider here only 1 test set
+          write_sparse_txt_to_db(testdbfullname, _tests_txt[0]);
+        // MULTIPLE TEST SETS : we consider here only 1 test set
+        for (auto tt : _tests_txt)
+          destroy_txt_entries(tt);
+        _tests_txt.clear();
       }
 
     return 0;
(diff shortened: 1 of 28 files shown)


tests/ut-torchapi.cc
```diff
@@ -308,9 +310,216 @@ TEST(torchapi, service_train_images)
...
 
   fileops::clear_directory(resnet50_train_repo + "train.lmdb");
-  fileops::clear_directory(resnet50_train_repo + "test.lmdb");
+  fileops::clear_directory(resnet50_train_repo + "test_0.lmdb");
   fileops::remove_dir(resnet50_train_repo + "train.lmdb");
-  fileops::remove_dir(resnet50_train_repo + "test.lmdb");
+  fileops::remove_dir(resnet50_train_repo + "test_0.lmdb");
+}
+
+TEST(torchapi, service_train_images_multiple_testsets)
```
BynaryCobweb, line 318: I don't think it is useful to add other full training tests (with 200 iterations or so) just to check the "measure" field. Some assertions can be removed, and more specific assertions can be added. The training time can be reduced to 1 or 2 iterations, unless we want to check that the loss shrinks (but it is already done in other tests so it would be redundant here I think)

fantes: you are very right, will do !

fantes: done !

src/backends/caffe/caffeinputconns.cc
```diff
@@ -1653,13 +1677,20 @@ namespace dd
...
-          write_txt_to_db(testdbfullname, _test_txt);
+          // MULTIPLE TEST SETS : we consider here only 1 test set
+          write_txt_to_db(testdbfullname, _tests_txt[0]);
         else
-          write_sparse_txt_to_db(testdbfullname, _test_txt);
-        destroy_txt_entries(_test_txt);
+          // MULTIPLE TEST SETS : we consider here only 1 test set
+          write_sparse_txt_to_db(testdbfullname, _tests_txt[0]);
+        // MULTIPLE TEST SETS : we consider here only 1 test set
+        // destroy_txt_entries(_test_txt);
```
beniz, line 1690: remove?

src/backends/torch/torchdataset.h
```diff
@@ -302,6 +304,218 @@ namespace dd
                             torch::Tensor &targett, const bool &bw);
   };
 
+  /**
+   * \brief this class holds several datasets, for implementeing multiple
```
beniz, line 308: Can remove the sentence after the `,`

src/backends/torch/torchsolver.cc
```diff
@@ -201,29 +201,41 @@ namespace dd
...
+                std::string test_name;
+                bestfile.open(bestfilename, std::ios::in);
+                // first three fields are thrown away
+                bestfile >> tmp >> bin >> tmp >> tmp >> tmp >> test_name;
+                if (test_name != set_names[test_id])
+                  _logger->warn(
+                      "test names not matching: {} (API) vs {} (file)",
+                      set_names[test_id], test_name);
+                bestfile.close();
+                best_iteration_numbers[test_id] = std::atof(bin.c_str());
```
beniz, line 232: How do we choose checkpoint to load from eventually when we have multiple metrics ?

src/utils/fileops.hpp
```diff
@@ -428,6 +428,34 @@ namespace dd
 
       return 0;
     }
+
```
beniz, line 431: Add a comment to describe the function

src/utils/utils.hpp
```diff
@@ -54,6 +54,17 @@ namespace dd
       return true;
     }
 
+    static bool unique(int64_t val, std::vector<int64_t> vec)
```
beniz, line 57: Maybe this is more compact: 
```
auto it = std::unique( a.begin(), a.end() );
bool wasUnique = (it == a.end() );
```
though this is not exactly what you are doing here.

beniz (approved):

src/backends/torch/torchdataset.cc
```diff
@@ -540,7 +540,7 @@ namespace dd
       {
         /*at::Tensor imgt = image_to_tensor(dimg._imgs[0], height, width);
         at::Tensor targett = target_to_tensor(target);
-        add_batch({ imgt }, { targett });*/
+        add_batch({ imgt }, { targe<tt });*/
```
BynaryCobweb, line 543: Added typo?

src/backends/torch/torchinputconns.cc
```diff
@@ -544,6 +639,13 @@ namespace dd
     return output;
   }
 
+  void TxtTorchInputFileConn::fill_dataset(
```
BynaryCobweb, line 642: I find this function redundant, `fill_dataset(testsets[test_id], entries)` is actually the same length as `fill_dataset(testsets, entries, test_id)` and is clearer imho

src/backends/torch/torchinputconns.cc
```diff
@@ -965,17 +1065,34 @@ namespace dd
         }
   }
 
-  void CSVTSTorchInputFileConn::fill_dataset(TorchDataset &dataset,
-                                             bool use_csvtsdata_test)
+  void CSVTSTorchInputFileConn::fill_dataset()
```
BynaryCobweb, line 1068: The id of passing TorchDataset to functions like this is to have a generic way of managing datasets, and making the code simpler.

I think this function could have been kept the way it was, and iteration on test set could be performed in ::transform.

src/backends/torch/torchinputconns.h
```diff
@@ -79,6 +79,20 @@ namespace dd
     {
     }
 
+    void check_tests_sizes(size_t tests_size, size_t names_size)
+    {
+      if (tests_size != names_size)
+        {
+          std::string msg
+              = "something wrong happened with datasets definitions: "
```
BynaryCobweb, line 87: When is this supposed to happen? Is it an internal error or bad parameters from the user?

src/backends/torch/torchsolver.cc
BynaryCobweb: API parameter? It could be interesting to keep all checkpoint and compare them with eachother

BynaryCobweb: Currently I think the most recent checkpoint is used

src/backends/caffe/caffeinputconns.cc
fantes: ok

src/backends/torch/torchdataset.h
fantes: ok

src/backends/torch/torchsolver.cc
fantes: indeed it is the most recent file that is used, this is the behavior defined here https://github.com/jolibrain/deepdetect/blob/dbafe3a315a3f8f5920e9d7b8f77da76e6014251/src/backends/torch/torchmodel.cc#L54

ATM, several weights files remains in the repo: the best for every metric, and the regular snapshots. 

for now, user has to copy the proper file into  new repo (or equivalently remove unwanted weights). 
upgrading this behavior (for instance changing resume from true to best for metric n) may be out of the scope of this PR

src/utils/fileops.hpp
fantes: ok

src/utils/utils.hpp
fantes: It seems to me that std::unique removes duplicates, which is not what I wanted to do, for some reason I do not remember

src/csvtsinputfileconn.cc
```diff
@@ -78,8 +83,9 @@ namespace dd
       return -1;
   }
 
-  int DDCsvTS::read_dir(const std::string &dir)
+  int DDCsvTS::read_dir(const std::string &dir, int test_id)
   {
+    (void)test_id;
```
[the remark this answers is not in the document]
fantes: you are very right, this function read the dir passed in parameter _and_ the `_csv_test_fnames` _attribute_ , which is at very least extremely questionable :) 
this is inherited/copyied from csv connector behavior which does the same (attribute used is the one from the csv connector, as csvts connector inherits from csv connector). This could be refactored for clarity, or left as is because it works ...

src/backends/torch/torchdataset.cc
fantes: yup

src/backends/torch/torchinputconns.cc
fantes: ok, removed

src/backends/torch/torchinputconns.h
fantes: It is not supposed to happen, this smells fear

src/backends/torch/torchinputconns.cc
fantes: ok, refactored

fantes:
@BynaryCobweb I am not sure to get your last point, all checkpoints thare are best for any metric are kept (or should be) . What do you want to put in API ?

BynaryCobweb:
> @BynaryCobweb I am not sure to get your last point, all checkpoints thare are best for any metric are kept (or should be) . What do you want to put in API ?

src/backends/torch/torchinputconns.h
BynaryCobweb: If this is not supposed to happen it should be a `InputConnectorInternalException`

fantes: ok, done

tests/ut-torchapi.cc
```diff
@@ -853,6 +1014,79 @@ TEST(torchapi, service_train_csvts_nbeats)
...
+      = "{\"service\":\"" + sname
+        + "\",\"async\":false,\"parameters\":{\"input\":{\"shuffle\":true,"
+          "\"separator\":\",\",\"scale\":true,\"backcast_timesteps\":50,"
+          "\"forecast_timesteps\":50,\"ignore\":["
+          "\"output\"]},\"mllib\":{\"gpu\":false,\"solver\":{\"iterations\":10"
+        + ",\"test_interval\":3,\"base_lr\":0.1,\"snapshot\":500,\"test_"
+          "initialization\":false,\"solver_type\":\"ADAM\"},\"net\":{\"batch_"
+          "size\":2,\"test_batch_"
+          "size\":10}},\"output\":{\"measure\":[\"L1\",\"L2\"]}},\"data\":[\""
+        + csvts_data + "\",\"" + csvts_test + "\",\"" + csvts_test + "\"]}";
```
BynaryCobweb, line 1057: csvts_test2?

fantes: nice catch !

fantes: fixed

BynaryCobweb (approved):

PR #1099, opened 2020-12-21
fix: scale image input with NCNN

beniz:


src/backends/ncnn/ncnninputconns.h
```diff
@@ -113,6 +113,20 @@ namespace dd
...
+            {
+              std::vector<float> vscale = { norm };
+              _in.substract_mean_normalize(0, vscale.data());
+            }
+          else
+            {
+              std::vector<float> vscale = { norm, norm, norm };
+              _in.substract_mean_normalize(0, vscale.data());
+            }
+        }
```
BynaryCobweb, line 129: In torch backend, scale is applied before std and mean. Should this be the preferred behavior ? I feel like if the user provides both scale and mean / std at the same time, and they are not supposed to, an exception should be thrown.

tests/ut-ncnnapi.cc
```diff
@@ -85,4 +85,20 @@ TEST(ncnnapi, service_predict)
...
+        + incept_repo + "face.jpg\"]}";
+  joutstr = japi.jrender(japi.service_predict(jpredictstr));
+  std::cout << "joutstr=" << joutstr << std::endl;
+  jd.Parse<rapidjson::kParseNanAndInfFlag>(joutstr.c_str());
+  ASSERT_TRUE(!jd.HasParseError());
+  ASSERT_EQ(200, jd["status"]["code"]);
+  ASSERT_TRUE(jd["body"]["predictions"].IsArray());
+  cl1 = jd["body"]["predictions"][0]["classes"][0]["cat"].GetString();
+  ASSERT_TRUE(jd["body"]["predictions"][0]["classes"][0]["prob"].GetDouble()
+              > 0.4);
```
BynaryCobweb, line 103: This test checks that there are no errors when providing scale, but offer no guarantee that the scale is correctly applied

src/backends/ncnn/ncnninputconns.h
beniz: Yes, raising the exception should go into another PR I believe.

tests/ut-ncnnapi.cc
beniz: Yes, we don't have models like this yet, at least it goes through the code, that uses the same NCNN mat function as `mean` and `std`.

fantes (approved):

tests/ut-ncnnapi.cc
BynaryCobweb: I can try to export a pytorch model to ncnn... Maybe in the future :)

BynaryCobweb (approved):

PR #1106, opened 2020-12-30
fix: wrong template_params with torch from DD API

beniz:


beniz:
@merfifyio update

fantes:
I think this change is not compatible with https://github.com/jolibrain/deepdetect/blob/eeb783a9929bb8c4282ab079b3448429b6c2b009/src/backends/torch/native/native_factory.cc#L50

beniz:
> I think this change is not compatible with
> 
> https://github.com/jolibrain/deepdetect/blob/eeb783a9929bb8c4282ab079b3448429b6c2b009/src/backends/torch/native/native_factory.cc#L50

Yes, good catch, though the tests are passing :(  We need a clear semantic then. Personnally I don't like the `template_params` object too much since we've never used it before to configure the templates, but I recognize it's a fine addition, so I'm ready to keep it. 

However, I am willing to keep the fix that is in this PR, i.e. having the object `_template_params` storing exactly what's in the API object. If that is fine I can try to fix the native factory code.

fantes:
> > I think this change is not compatible with
> > https://github.com/jolibrain/deepdetect/blob/eeb783a9929bb8c4282ab079b3448429b6c2b009/src/backends/torch/native/native_factory.cc#L50
> 
> Yes, good catch, though the tests are passing :( We need a clear semantic then. Personnally I don't like the `template_params` object too much since we've never used it before to configure the templates, but I recognize it's a fine addition, so I'm ready to keep it.
> 
> However, I am willing to keep the fix that is in this PR, i.e. having the object `_template_params` storing exactly what's in the API object. If that is fine I can try to fix the native factory code.

the idea here was to be able to pass a template_param in a generic way, for now we use / parse simple string, but in the future we may want to pass more complex structures. The native_factory part was written very quickly , feel free to fix it the way you want :)

BynaryCobweb:
The NativeFactory need the number of classes for classification models, and it is not contained in `_template_params`.
I agree that having `_template_params` used for all mllib parameters is not a good idea though

beniz:
I've added the two following modifications:
- `nbeats` template now requires `"template_params":{"stackdef":["t2","s4","g3","b3"]}"`, this is a breaking change @louisj @fantes 
- `template_params` automatically gets `nclasses` added to it if available from the main `mllib` object. This is a temporary hack for vision native models typically, since DTO should take over APIData and then we'll have to review and cleanup all nested structures.

Let me know if this requires further improvements, and beware of the change with `nbeats` template.

fantes (approved):

BynaryCobweb (approved):

PR #1107, opened 2020-12-30
ml: torch ViT small and tiny models as in https://arxiv.org/abs/201212877

beniz:
Weights are available from https://github.com/facebookresearch/deit


beniz:
@mergifyio update

beniz:
The semantic check fails because `ml` is not in the standard, so I'll merge by hand once it's been reviewed.

BynaryCobweb (approved):

fantes (approved):

PR #1111, opened 2021-01-04
build: remove use of pytorch-complete file

BynaryCobweb:
Let CMake handle rebuild of pytorch.


fantes:
did you double check this does not patch/recompile every time ? the very same change on ncnn backend does not work : it tries to apply patches every time. It also seem to me that I tried to do almost the same change a few month ago, and it kept recompiling pytorch again and again...  I do not understant this cmake dark magic

sileht:
I have looked at the ncnn rebuild two weeks ago and I come to the conclusion that PATCH_COMMAND does not play well with GIT_TAG/GIT_REPOSITORY.

Git Checkout can fail if patches are already applied and if it succeed, cmake retry to apply them, but they are already there.

Some folks on cmake lists recommended using a special or fork to manage patches on top of upstream repo/branch to avoid this.

PR #1112, opened 2021-01-05
test(torch): Make some tests deterministic

BynaryCobweb:


BynaryCobweb:
CI failed once and then passed when I restarted it.
It means that added `at::set_deterministic(true)` and other stuff still does not make the test deterministic

sileht (approved):

fantes (approved):

PR #1114, opened 2021-01-05
Fix NCNN bbox loop + Add datatype fp32/fp16 control to NCNN backend inference

### LINKED ISSUE #1113: NCNN backend outputs spurious bboxes in some cases.
This issue can be directly observed by running NCNN backend unit tests with `./ut_ncnnapi`, some boxes are spurious, e.g.:

1- with negative coordinates
```
{"bbox":{"ymax":202.06507873535157,"xmax":230.23036193847657,"xmin":71.65133666992188,"ymin":-6.1509599685668949},"cat":"15","prob":0.036612819880247119}
```
This problem seems to only affect very low confidence results. Using `confidence_threshold` to cut results lower than a relevant threshold value (e.g. 0.25) appears to remove the issue. 

Current understanding is that numerical approximations may have bbox coordinates drift a bit as regression outputs of the neural net. Since bboxes are computed within [0,1], small initial numerical drifts outside of these bounds are amplified, yielding truely negative coordinates. This is expected to be witnessed only for boxes that may be close to a image's edges.

2- with NaN values (only observed with Vulkan on Apple M1 GPU at the moment):
```
{"last":true,"bbox":{"ymax":NaN,"xmax":NaN,"ymin":NaN,"xmin":NaN},"prob":0.9482421875,"cat":"15"}]
```

This issue appears to disappear on the affected GPU by removing `fp16` arithmetics in inference. The previous issue above does not appear to be much affected by floating point precision. 

As a fix we'll provide a `datatype` selector for the `ncnn` backend, similar to the one option already availlable with the `tensorrt` backend.

beniz:
Should fix #1113

### DIFF
--- a/src/backends/ncnn/ncnnlib.cc
+++ b/src/backends/ncnn/ncnnlib.cc
@@ -55,7 +55,6 @@ namespace dd
   {
     this->_libname = "ncnn";
     _net = new ncnn::Net();
-    _net->opt.lightmode = true;
     _net->opt.num_threads = _threads;
     _net->opt.blob_allocator = &_blob_pool_allocator;
     _net->opt.workspace_allocator = &_workspace_pool_allocator;
@@ -94,6 +93,13 @@ namespace dd
   void NCNNLib<TInputConnectorStrategy, TOutputConnectorStrategy,
                TMLModel>::init_mllib(const APIData &ad)
   {
+    bool use_fp32 = (ad.has("datatype")
+                     && ad.get("datatype").get<std::string>()
+                            == "fp32"); // default is fp16
+    _net->opt.use_fp16_packed = !use_fp32;
+    _net->opt.use_fp16_storage = !use_fp32;
+    _net->opt.use_fp16_arithmetic = !use_fp32;
+
     int res = _net->load_param(this->_mlmodel._params.c_str());
     if (res != 0)
       {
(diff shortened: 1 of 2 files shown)


beniz:
The `fp32` support is tested via #1105 with dedicated unit test.

src/backends/ncnn/ncnnlib.cc
```diff
@@ -94,6 +93,13 @@ namespace dd
   void NCNNLib<TInputConnectorStrategy, TOutputConnectorStrategy,
                TMLModel>::init_mllib(const APIData &ad)
   {
+    bool use_fp32 = (ad.has("datatype")
+                     && ad.get("datatype").get<std::string>()
+                            == "fp32"); // default is fp16
+    _net->opt.use_fp16_packed = !use_fp32;
+    _net->opt.use_fp16_storage = !use_fp32;
```
fantes, line 100: when loading weights from caffe, I am pretty sure they are stored as fp32, so in this case if you do not provide explicitely datatype= fp32 in API, does not everything fail ?

fantes: ok it seems that there is an explicit conversion from fp32 to fp16 after load time and before inference if  use_weight_fp16_storage is true, as least in lstm_x86

fantes (approved):

BynaryCobweb (approved):

src/backends/ncnn/ncnnlib.cc
beniz: Thanks, this is useful. I did put/leave `fp16` as the implicit default for two reasons: when using NCNN it's usually we are running on compute-limited platform, and it seems to have been the default at least on the branch I was on, but it is the default setting in their example code snippets and benchmarks... We can revert it in case some issues do surface...

PR #1117, opened 2021-01-07
fix: wrong NCNN bbox output scaling

beniz:
We could have `_img_sizes` in the main image data connector instead.

### DIFF
--- a/src/backends/ncnn/ncnnlib.cc
+++ b/src/backends/ncnn/ncnnlib.cc
@@ -281,6 +281,21 @@ namespace dd
 
     if (bbox == true)
       {
+        std::string uri = inputc._ids.at(0);
+        auto bit = inputc._imgs_size.find(uri);
+        int rows = 1;
+        int cols = 1;
+        if (bit != inputc._imgs_size.end())
+          {
+            // original image size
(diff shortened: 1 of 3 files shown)


src/backends/ncnn/ncnnlib.cc
```diff
@@ -281,6 +281,21 @@ namespace dd
...
+        if (bit != inputc._imgs_size.end())
+          {
+            // original image size
+            rows = (*bit).second.first;
+            cols = (*bit).second.second;
+          }
+        else
+          {
+            this->_logger->error("couldn't find original image size for {}",
+                                 uri);
```
BynaryCobweb, line 297: Throw an error maybe? If rows = cols = 1, then we divide by 0 later in the code, leading to probably NaN bbox, which is not easy to debug.

fantes (approved):

src/backends/ncnn/ncnnlib.cc
beniz: Good catch, I've updated the PR with an internal exception, and added another commit for having the same exact behavior with the caffe backend.

BynaryCobweb (approved):

PR #1120, opened 2021-01-08
fix: disable OpenMP by default

sileht:
Since we have oat++ by default, we got some segfault from time to time
in libtbb.

libtbb is used by opencv for resizing images for example.

On our side, we also parallelize the image load and resizing with
OpenMP.

Disabling OpenMP doesn't trigger the libtbb issue.

Since reading files in parallel doesn't really improve throughput, we
disable it by default.


sileht:
I finally got the crash even without openmp :(

PR #1128, opened 2021-01-12
feat(ml): retain_graph control via API for torch autograd

beniz:
Equivalent to the `retain_graph=True` in Pytorch.


fantes (approved):
I wonder if this should be automatically set iff graph has multiple outputs

BynaryCobweb:
> I wonder if this should be automatically set iff graph has multiple outputs

Do we have a use case of this in dede ?

BynaryCobweb (approved):

PR #1141, opened 2021-01-19
feat: running custom methods from jit models

beniz:
Example usage:

- From https://github.com/openai/CLIP, get OpenAI CLIP model from https://openaipublic.azureedge.net/clip/models/40d365715913c9da98579312b702a82c18be219cc2a73407c4526f58eba950af/ViT-B-32.pt  and put it into a directory

- Create service:

```bash
curl -X PUT http://localhost:8080/services/testclip -d '                                             
{"mllib":"torch","description":"image","type":"supervised","model":{"repository":"/path/to/model\
s/clip/"},"parameters":{"input":{"connector":"image","width":224,"height":224},"mllib":{"gpu":true,"gpuid":0}}}                                                                                       
'       
```

- Inference on an image:

```bash
curl -X POST http://localhost:8080/predict -d '                                                      
{                                                                                                    
"service":"testclip",                                                                                
"parameters": {                                                                                      
    "output": {                                                                                                                                                               
    },                                                                                               
    "mllib": {                                                                                       
      "gpu": true,                                                                                   
      "forward_method":"encode_image",                                                                                                                                        
      "extract_layer":"last" ,
     "input": {                                                                                       
         "rgb": true,                                                                                    
         "mean":[122,115,125],                                                                           
         "std":[66,66,70]                                                                                
      }                                                                                
    }                                                                                                
  },                                                                                                 
  "data": [                                                                                          
    "/path/to/img.jpg"                                                  
  ]                                                                                                  
}                                                                                                    
'           
```

This should return the encoded tensor values for the image.

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -945,6 +945,8 @@ namespace dd
     APIData params = ad.getobj("parameters");
     APIData output_params = params.getobj("output");
     std::string extract_layer;
+    bool extract_last = false;
+    std::string forward_method;
 
     if (params.has("mllib"))
       {
@@ -956,7 +958,13 @@ namespace dd
               predict_batch_size = ad_net.get("test_batch_size").get<int>();
           }
         if (ad_mllib.has("extract_layer"))
-          extract_layer = ad_mllib.get("extract_layer").get<std::string>();
+          {
+            extract_layer = ad_mllib.get("extract_layer").get<std::string>();
+            if (extract_layer == "last")
+              extract_last = true;
+          }
+        if (ad_mllib.has("forward_method"))
+          forward_method = ad_mllib.get("forward_method").get<std::string>();
       }
 
     bool lstm_continuation = false;
@@ -987,7 +995,8 @@ namespace dd
     torch::Device cpu("cpu");
     _module.eval();
 
-    if (!extract_layer.empty() && !_module.extractable(extract_layer))
+    if (!extract_last && !extract_layer.empty()
+        && !_module.extractable(extract_layer))
       {
         std::string els;
         for (const auto &el : _module.extractable_layers())
@@ -1033,8 +1042,9 @@ namespace dd
         Tensor output;
         try
           {
-            if (extract_layer.empty())
-              output = torch_utils::to_tensor_safe(_module.forward(in_vals));
+            if (extract_layer.empty() || extract_last)
+              output = torch_utils::to_tensor_safe(
+                  _module.forward(in_vals, forward_method));
             else
               output = torch_utils::to_tensor_safe(
                   _module.extract(in_vals, extract_layer));
@@ -1114,8 +1124,8 @@ namespace dd
                   best_count = output_params.get("best").get<int>();
                 std::tuple<Tensor, Tensor> sorted_output
                     = output.sort(1, true);
-                auto probs_acc
-                    = std::get<0>(sorted_output).accessor<float, 2>();
+                Tensor probsf = std::get<0>(sorted_output).to(torch::kFloat);
+                auto probs_acc = probsf.accessor<float, 2>();
                 auto indices_acc
                     = std::get<1>(sorted_output).accessor<int64_t, 2>();
 
--- a/src/backends/torch/torchmodule.cc
+++ b/src/backends/torch/torchmodule.cc
@@ -266,7 +266,8 @@ namespace dd
       }
   }
 
-  c10::IValue TorchModule::forward(std::vector<c10::IValue> source)
+  c10::IValue TorchModule::forward(std::vector<c10::IValue> source,
+                                   const std::string &forward_method)
   {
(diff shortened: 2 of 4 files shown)


fantes (approved):

src/backends/torch/torchlib.cc
```diff
@@ -888,7 +890,13 @@ namespace dd
               predict_batch_size = ad_net.get("test_batch_size").get<int>();
           }
         if (ad_mllib.has("extract_layer"))
-          extract_layer = ad_mllib.get("extract_layer").get<std::string>();
+          {
+            extract_layer = ad_mllib.get("extract_layer").get<std::string>();
+            if (extract_layer == "last")
```
BynaryCobweb, line 895: So `extract_last` is true iff extract_layer == "last"? Seems redundant

src/backends/torch/torchmodule.cc
```diff
@@ -275,8 +276,24 @@ namespace dd
...
+          {
+            for (const auto &method : _traced->get_methods())
+              {
+                if (method.name() == forward_method)
+                  {
+                    _logger->info("found forward method {}", method.name());
+                    auto output = _traced->get_method(forward_method)(
+                        std::move(source));
+                    source = torch_utils::unwrap_c10_vector(output);
+                  }
```
BynaryCobweb, line 289: Error if not found

BynaryCobweb: Actually code for get_method is following:
```c++
  Method get_method(const std::string& name) const {
    if (auto method = find_method(name)) {
      return *method;
    }
    AT_ERROR("Method '", name, "' is not defined.");
  }
```
So we can do
```c++
if (auto method = find_method(name)) {
      _logger->info("found forward method {}", method.name());
      auto output = (*method)(std::move(source));
      source = torch_utils::unwrap_c10_vector(output);
}
else throw MLLibBadParamException("Method " + forward_method + " not found in traced model");
```

src/backends/torch/torchmodule.h
```diff
@@ -52,7 +52,8 @@ namespace dd
     /**
      * \brief forward (inference) pass over the network
      */
-    c10::IValue forward(std::vector<c10::IValue> source);
+    c10::IValue forward(std::vector<c10::IValue> source,
```
BynaryCobweb, line 55: I'd prefer forward_method as a class parameter, to keep forward signature simple

src/backends/torch/torchlib.cc
```diff
@@ -965,8 +974,9 @@ namespace dd
         Tensor output;
         try
           {
-            if (extract_layer.empty())
-              output = torch_utils::to_tensor_safe(_module.forward(in_vals));
+            if (extract_layer.empty() || extract_last)
```
BynaryCobweb, line 977: so extract_last does exactly the same as extract_layer.empty() ?

src/backends/torch/torchmodule.h
beniz: Not really because the use-case is to be able to change the forward method per call. E.g. the CLIP model has `encode_text` and  `encode_image` within the same JIT model.

src/backends/torch/torchlib.cc
beniz: No, `extract_layer` set to `"last"` bypasses the call to `extractable()`. This is a bit hacky, but I did it for two reasons:
- `extractable()` is not implemented on our side for traced models yet
- it unlocks the useful and generic use-case of running a network an returning the flatten tensors from the last layer. GAN generator and even segmentation models fall within this category. Though some elements should be refined in the future (e.g. non-flatten tensors).

sileht:
@mergifyio update

src/backends/torch/torchlib.cc
BynaryCobweb: Ok, I missed the fact that if extract_layer != "" we already return a tensor

src/backends/torch/torchmodule.h
BynaryCobweb: ok

src/backends/torch/torchmodule.cc
beniz: Done.

BynaryCobweb:
Maybe add a test as well?

beniz:
> Maybe add a test as well?

Not at this stage, I'll wait til the integration of GAN generator support, coming up soon.

beniz:
> Maybe add a test as well?

The test on the `extract_layer: last` output will come from another PR that adds support for GAN generators. 

We don't have models for testing `forward_method` unless we either use the one from CLIP or craft one.

BynaryCobweb (approved):

sileht:
@Mergifyio rebase

beniz:
@mergifyio update

sileht:
@Mergifyio rebase

sileht:
@Mergifyio rebase

PR #1146, opened 2021-01-21
feat: use oatpp::DTO to parse img-input-connector APIData

sileht:


src/imginputfileconn.h
```diff
@@ -453,81 +455,92 @@ namespace dd
...
     {
+      auto params = ad.createSharedDTO<dd::DTO::ImgInputConnectorParameters>();
+
       // optional parameters.
-      if (ad.has("width"))
-        _width = ad.get("width").get<int>();
-      if (ad.has("height"))
-        _height = ad.get("height").get<int>();
-      if (ad.has("crop_width"))
+      if (params->width != nullptr)
```
beniz, line 461: Maybe we could write a short templated function that does the check against `nullptr` implicitely to simplify the code ?

src/imginputfileconn.h
```diff
@@ -43,6 +43,8 @@
 #include "utils/apitools.h"
 #include <random>
 
+#include "http/dto/img_connector.hpp"
```
beniz, line 46: CI says you are missing this file !

src/imginputfileconn.h
lganzzzo: you can do just:

```cpp
if (params->width) ... //<-- check for null
```

src/imginputfileconn.h
```diff
@@ -453,81 +455,92 @@ namespace dd
...
-      if (ad.has("shuffle"))
-        _shuffle = ad.get("shuffle").get<bool>();
-      if (ad.has("seed"))
-        _seed = ad.get("seed").get<int>();
-      if (ad.has("test_split"))
-        _test_split = ad.get("test_split").get<double>();
-      if (ad.has("mean"))
+
+      if (params->bw != nullptr)
+        _bw = params->bw;
```
lganzzzo, line 488: `oatpp::Boolean` has the following implicit conversions:

```cpp
  {
    oatpp::Boolean ob = nullptr;
    bool b = ob;
    OATPP_ASSERT(b == false);
  }

  {
    oatpp::Boolean ob = false;
    bool b = ob;
    OATPP_ASSERT(b == false);
  }

  {
    oatpp::Boolean ob = true;
    bool b = ob;
    OATPP_ASSERT(b == true);
  }
```

beniz (approved):

fantes (approved):

PR #1147, opened 2021-01-21
Pull request for dto-ncnn

sileht:
## feat: use DTO for NCNN init parameters


beniz:
Thanks, sounds simple. What do think about putting the backend DTOs headers into `src/backends/ncnn/dto` for instance for ncnn instead of within the higher level `http/` dir ?

tests/ut-ncnnapi.cc
```diff
@@ -312,7 +320,18 @@ TEST(ncnnapi, ocr)
         + ocr_repo
         + "\"},\"parameters\":{\"input\":{\"connector\":\"image\",\"ctc\":"
           "true, \"height\":136,\"width\":220},\"mllib\":{\"nclasses\":69}}}";
-  std::string joutstr = japi.jrender(japi.service_create(sname, jstr));
+
+  auto service_create = dd::DTO::ServiceCreate::createShared();
```
beniz, line 324: Is there a way of creating the DTO object directly from the JSON string ?

sileht: Yes, and that's a good idea to do it in tests, it will ensure the current json string still work as-is.

Issue #1151, opened 2021-01-22
multiword_ocr fails to convert to NCNN


cchadowitz-pf:
#### Configuration

- Version of DeepDetect:
    - [x] Docker CPU - built from https://github.com/jolibrain/deepdetect/archive/v0.13.0.tar.gz using `docker build --build-arg DEEPDETECT_DEFAULT_MODELS=false -t jolibrain/deepdetect_cpu_pfi:0.13.0 --no-cache -f docker/cpu.Dockerfile .`
- Commit (shown by the server when starting):
This build still doesn't show the commit ref when starting, but I used the release archive listed above:
```
DeepDetect  (dev)
GIT REF: :
COMPILE_FLAGS: USE_CAFFE2=OFF USE_TF=OFF USE_NCNN=ON USE_TORCH=OFF USE_HDF5=ON USE_CAFFE=ON USE_TENSORRT=OFF USE_TENSORRT_OSS=OFF USE_DLIB=OFF USE_CUDA_CV=OFF USE_SIMSEARCH=ON USE_ANNOY=OFF USE_FAISS=ON USE_COMMAND_LINE=ON USE_JSON_API=ON USE_HTTP_SERVER=ON
DEPS_VERSION: OPENCV_VERSION=3.2.0 CUDA_VERSION= CUDNN_VERSION= TENSORRT_VERSION=
```
#### Your question / the problem you're facing:
I'm trying to compare the `word_detect_v2` and `multiword_ocr` Caffe models (as a chain) vs the same models converted to the NCNN backend. The `word_detect_v2` model converted without an error, but the `multiword_ocr` model threw an error and killed the server (see the bottom for the server logs).

#### Error message (if any) / steps to reproduce the problem:

- [x] list of API calls:
For `word_detect_v2`, I used:
```
curl -X PUT "http://localhost:8080/services/word_detect_v2" -d '
{
  "description": "word_detect_v2",
  "mllib": "caffe",
  "type": "supervised",
  "parameters": {
    "input": {
      "connector": "image",
      "width": 512,
      "height": 512
    },
    "mllib": {
      "nclasses": 2,
      "gpu": true
    },
    "output": {}
  },
  "model":{
   "repository":"/opt/models/word_detect_v2/"
  }
}' | jq
```
```
curl -X PUT "http://localhost:8080/services/word_detect_v2_ncnn" -d '
{
  "description": "word_detect_v2",
  "mllib": "ncnn",
  "type": "supervised",
[...]

PR #1152, opened 2021-01-22
fix: missing continuation indicator layer patch for NCNN

beniz:
This is what causes #1151 but does not solve the OCR model conversion issue.


fantes:
I guess I missed this one

fantes (changes requested):
superseeded by https://github.com/jolibrain/deepdetect/pull/1153

PR #1153, opened 2021-01-26
fix(ncnn): update innerproduct so that it does not pack data

fantes:
this PR superseeds https://github.com/jolibrain/deepdetect/pull/1152
(ie it incorporates this PR1152 + fixes another issue)
this should fix https://github.com/jolibrain/deepdetect/issues/1151
issue fixed :  ncnn now uses a lot of architecture dependent optimizations , and our patch on innerproducts for keeping h dimension is implemented only for generic code thus we force generic code execution ; before this patch, data was packed for some architecture optim but generic code was called. This fix by disabling data packing for innerproduct layer

now with UT ! :)

### DIFF
--- a/tests/ut-ncnnapi.cc
+++ b/tests/ut-ncnnapi.cc
@@ -39,6 +39,7 @@ static std::string squeezenet_ssd_repo
     = "../examples/ncnn/squeezenet_ssd_ncnn/";
 static std::string squeezenet_repo = "../examples/ncnn/squeezenet_ncnn/";
 static std::string incept_repo = "../examples/ncnn/squeezenet_ssd_ncnn/";
+static std::string ocr_repo = "../examples/ncnn/ocr/";
 static std::string sinus = "../examples/all/sinus/";
 static std::string model_templates_repo = "../templates/caffe/";
 static std::string gpuid = "0"; // change as needed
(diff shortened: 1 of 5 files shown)


beniz:
So this subsumes #1152 right ?

tests/ut-ncnnapi.cc
```diff
@@ -299,3 +300,33 @@ TEST(ncnnapi, service_lstm)
...
+                            + ocr_repo + "word_ocr.jpg\"]}";
+  joutstr = japi.jrender(japi.service_predict(jpredictstr));
+  JDoc jd;
+  std::cout << "joutstr=" << joutstr << std::endl;
+  jd.Parse<rapidjson::kParseNanAndInfFlag>(joutstr.c_str());
+  ASSERT_TRUE(!jd.HasParseError());
+  ASSERT_EQ(200, jd["status"]["code"]);
+  ASSERT_TRUE(jd["body"]["predictions"].IsArray());
+  ASSERT_TRUE(jd["body"]["predictions"].Size() == 1);
+  ASSERT_TRUE(jd["body"]["predictions"][0]["classes"][0]["cat"] == "beleved");
```
beniz, line 331: beleved ? :)

beniz (approved):

tests/ut-ncnnapi.cc
fantes: yup, using previous ncnn (ie our old fork) I also get "beleved", while using caffe I get "believed"

beniz: Let's try with fp32 activated (fp16 is default), would be surprised it corrects the output though...

BynaryCobweb (approved):

sileht:
@Mergifyio rebase

fantes:
@Mergifyio rebase

PR #1176, opened 2021-02-10
fix(torch/txt): correctly handle test sets in case of no splitting

fantes:
this PR fixes a bug in case of training with txt _and_ torch _and_ using explicit test set. This bug was introduced in https://github.com/jolibrain/deepdetect/pull/1095.
Also add a test case.


fantes:
@Mergifyio rebase

PR #1182, opened 2021-02-12
feat: use DTO for SVM input connector

beniz:


fantes (commented):
It seems that we are going toward a generic inputconnector_dto (as we cannot pass a SvmInputConnector object to predict call)

PR #1192, opened 2021-02-23
feat(torch): add inference for torch detection models

BynaryCobweb:
Supports detection models that are available in torchvision: Faster
RCNN, RetinaNet

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -75,6 +75,7 @@ namespace dd
     _classification = tl._classification;
     _regression = tl._regression;
     _timeserie = tl._timeserie;
+    _bbox = tl._bbox;
     _loss = tl._loss;
     _template_params = tl._template_params;
   }
@@ -248,12 +249,16 @@ namespace dd
       {
         _seq_training = true;
       }
+    else if (_template == "fasterrcnn" || _template == "retinanet")
+      {
+        _bbox = true;
+      }
     else if (_template.find("recurrent") != std::string::npos
              || NativeFactory::is_timeserie(_template))
       {
         _timeserie = true;
       }
-    if (!_regression && !_timeserie && self_supervised.empty())
+    if (!_regression && !_timeserie && !_bbox && self_supervised.empty())
       _classification = true; // classification is default
 
     // Set mltype
@@ -261,6 +266,8 @@ namespace dd
       this->_mltype = "classification";
     else if (_regression)
       this->_mltype = "regression";
+    else if (_bbox)
+      this->_mltype = "detection";
 
     // Create the model
     _module._device = _main_device;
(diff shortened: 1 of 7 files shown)


fantes (approved):

src/backends/torch/torchlib.cc
```diff
@@ -248,6 +248,10 @@ namespace dd
       {
         _seq_training = true;
       }
+    else if (_template == "fasterrcnn" || _template == "retinanet")
+      {
+        // bbox = true; // ?
```
beniz, line 253: We could force it automatically yes.

src/backends/torch/torchlib.cc
```diff
@@ -967,6 +980,25 @@ namespace dd
           forward_method = ad_mllib.get("forward_method").get<std::string>();
       }
 
+    if (output_params.has("bbox") && output_params.get("bbox").get<bool>())
+      {
+        bbox = true;
+      }
+    if (output_params.has("confidence_threshold"))
```
beniz, line 987: Interesting, we were not using it for classification models then! Then we should/could fix it here or in another tiny PR for classification models ?

src/backends/torch/torchlib.cc
```diff
@@ -1117,7 +1149,88 @@ namespace dd
             if (_module._native != nullptr)
               output = _module._native->cleanup_output(output);
 
-            if (_classification)
+            if (bbox)
+              {
+                // Supporting only Faster RCNN format at the moment.
```
beniz, line 1154: Since RetinaNet is mentioned in the PR, this only applies to Faster-RCNN, correct ?

BynaryCobweb: I forgot to change the comment, this works for RetinaNet as well.
There are two pure detection models in torchvision: Faster RCNN and RetinaNet, and they have the same output format.

src/backends/torch/torchlib.cc
BynaryCobweb: Do we need to pass it to the output parameters at predict time then?

src/backends/torch/torchlib.cc
BynaryCobweb: Will fix in another PR

src/backends/torch/torchlib.cc
beniz: Normally yes, but if we hardcode it, it'll provide some slack. There's no use not setting it up actually.

src/backends/torch/torchlib.cc
beniz: Let's not forget!

beniz (approved):

beniz:
@BynaryCobweb could you add to the PR description the steps to get a model and execute an inference ? It's always useful.

PR #1194, opened 2021-02-24
fix(torch/timeseries): in case of db, correctly finalize db

fantes:
this PR allows to use csvts + db
(not completely usefull combination: all csvts are read before being put into db, some more work is needed to put them into db one after the other. ATM only interesting property is to release CPU memory after putting them into db)
also add ut


BynaryCobweb (approved):

beniz:
Does the API need to be udpated ?

beniz (approved):

PR #1201, opened 2021-02-26
feat(torch/csvts): create db incrementally

fantes:
this PR allows to create db for timeseries on the fly (torch backend only), ie series are added to the db just after being read, then freed. Previously they were all read before being put into db, leading to very high temporary memory usage. This PR also allows not to rebuild db if already present.


src/backends/torch/torchdataset.cc
```diff
@@ -587,6 +587,25 @@ namespace dd
       }
   }
 
+  void TorchMultipleDataset::add_test_name_if_necessary(std::string longname,
+                                                        int test_id)
+  {
+    std::string name = fileops::shortname(longname);
+    if (_datasets.size() == static_cast<size_t>(test_id + 1)
```
BynaryCobweb, line 594: what happens if `_dataset.size() > static_cast<size_t>(test_id + 1)`?

src/backends/torch/torchdataset.cc
```diff
@@ -587,6 +604,25 @@ namespace dd
...
   }
 
+  void TorchMultipleDataset::add_test_name_if_necessary(std::string longname,
+                                                        int test_id)
+  {
+    std::string name = fileops::shortname(longname);
+    if (_datasets.size() >= static_cast<size_t>(test_id + 1)
+        && _datasets_names[test_id] == name)
+      return;
+    if (_datasets.size() == static_cast<size_t>(test_id + 1)
```
BynaryCobweb, line 614: shouldn't it be >= here too?

fantes: fixed

BynaryCobweb (approved):

beniz (approved):

PR #1203, opened 2021-03-01
feat(torch): torch detection training

Bycob:
Support of Faster RCNN & RetinaNet training on CPU at batch size 1
only.

Training on GPU leads to a segfault: https://github.com/pytorch/vision/issues/3525
EDIT: Fixed by upgrading pytorch to 1.8, see #1222

Models must be traced with a modified torchvision until this bug is resolved: 
https://github.com/pytorch/vision/issues/3469
`trace_torchvision` script uses the same torchvision used to build dede, which has the necessary modification.

TODO
- [x] working Faster RCNN training
- [ ] working RetinaNet training
- [x] Check that the MAP returned by dede is correct

```bash
# trace model with 2 classes
python3 tools/torch/trace_torchvision.py -v fasterrcnn_resnet50_fpn -o "/opt/platform/models/training/fasterrcnn_test_coco/" --num_classes 2

# service creation call
curl -X PUT 'http://localhost:8080/services/fasterrcnn_train' -d '{
    "description": "Torch traced fasterrcnn on cars",
    "mllib": "torch",
    "model": {
        "repository": "/opt/platform/models/training/fasterrcnn_train_cars/"
    },
    "parameters": {
        "input": {
            "connector": "image",
            "width":224,
            "height":224,
            "rgb":true,
            "scale":0.0039,
            "bbox":true
        },
        "mllib":{
            "template":"fasterrcnn",
            "nclasses":2,
            "gpu":true,
            "gpuid":0
        },
        "output": {
            "bbox": true,
            "store_config": true
        }
    },
    "mltype": "detection",
    "type": "supervised"
}'

# train call
curl -X POST 'http://localhost:8080/train' -d '{
   "service":"fasterrcnn_train",
   "parameters":{
      "input":{
         "db":true
      },
      "output":{
         "measure":[
            "map"
         ]
      },
      "mllib":{
         "solver":{
            "iterations":5000,
            "base_lr":1e-5,
            "finetune":true,
            "iter_size":16,
            "solver_type":"ADAM",
            "test_interval":250
         },
         "net":{
            "batch_size":1
         },
         "nclasses":1,
         "resume":false
      }
   },
   "data":[
      "/opt/platform/examples/cars/train.txt",
      "/opt/platform/examples/cars/test.txt"
   ]
}'
```

### DIFF
--- a/src/backends/torch/torchdataset.cc
+++ b/src/backends/torch/torchdataset.cc
@@ -525,6 +525,75 @@ namespace dd
     return add_image_file(fname, { target_to_tensor(target) }, height, width);
   }
 
+  int TorchDataset::add_image_bbox_file(const std::string &fname,
+                                        const std::string &bboxfname,
+                                        const int &height, const int &width)
+  {
+    // read image before reading bboxes to get the size of the image
+    ImgTorchInputFileConn *inputc
+        = reinterpret_cast<ImgTorchInputFileConn *>(_inputc);
+
+    DDImg dimg;
+    inputc->copy_parameters_to(dimg);
+
+    try
+      {
+        if (dimg.read_file(fname, -1))
+          {
+            this->_logger->error("Uri failed: {}", fname);
+          }
+      }
+    catch (std::exception &e)
+      {
+        this->_logger->error("Uri failed: {}", fname);
+      }
+
+    if (dimg._imgs.size() == 0)
+      {
+        return -1;
+      }
+    int orig_height = dimg._imgs_size[0].first;
+    int orig_width = dimg._imgs_size[0].second;
+
+    // read bbox file
+    std::vector<at::Tensor> bboxes;
+    std::vector<at::Tensor> classes;
+
+    std::ifstream infile(bboxfname);
+    std::string line;
+    double wfactor
+        = static_cast<double>(width) / static_cast<double>(orig_width);
+    double hfactor
+        = static_cast<double>(height) / static_cast<double>(orig_height);
+
+    while (std::getline(infile, line))
+      {
+        std::istringstream iss(line);
+        std::string val;
+        iss >> val;
+        int cls = std::stoi(val);
+        classes.push_back(target_to_tensor(cls));
+
+        std::vector<double> bbox(4);
+
+        iss >> val; // xmin
+        bbox[0] = std::stod(val) * wfactor;
+        iss >> val; // ymin
+        bbox[1] = std::stod(val) * hfactor;
+        iss >> val; // xmax
+        bbox[2] = std::stod(val) * wfactor;
+        iss >> val; // ymax
+        bbox[3] = std::stod(val) * hfactor;
+        // using target_to_tensor<std::vector<double>> (used by regression)
+        bboxes.push_back(target_to_tensor(bbox));
+      }
+
+    // add image
+    add_image_batch(dimg._imgs[0], height, width,
+                    { torch::stack(bboxes), torch::cat(classes) });
+    return 0;
+  }
+
   at::Tensor TorchDataset::image_to_tensor(const cv::Mat &bgr,
                                            const int &height, const int &width)
   {
--- a/src/backends/torch/torchinputconns.cc
+++ b/src/backends/torch/torchinputconns.cc
@@ -29,6 +29,29 @@ namespace dd
 
   using namespace torch;
 
+  void
+  TorchInputInterface::set_test_names(size_t test_dataset_count,
+                                      const std::vector<std::string> &uris)
+  {
+    std::vector<std::string> test_names;
+    if (uris.size() > 1)
+      for (size_t i = 1; i < uris.size(); ++i)
+        test_names.push_back(fileops::shortname(uris[i]));
+    else
+      {
+        if (test_dataset_count == 1)
+          {
+            _tilogger->info("test set will be named \"split\"");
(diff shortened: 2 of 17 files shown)


src/backends/torch/torchinputconns.cc
```diff
@@ -351,6 +396,56 @@ namespace dd
...
+                tests_lfiles.resize(_uris.size() - 1);
+                for (size_t i = 1; i < _uris.size(); ++i)
+                  {
+                    read_image_bbox(tests_lfiles[i - 1], _uris.at(i));
+                  }
+
+                if (_dataset._shuffle)
+                  shuffle_dataset<std::string>(lfiles);
+
+                // bool has_test_data = test_lfiles.size() != 0;
```
fantes, line 416: remove this ?

src/backends/torch/torchlib.cc
```diff
@@ -1530,13 +1578,165 @@ namespace dd
...
+      }
     ad_res.add("batch_size",
                entry_id); // here batch_size = tested entries count
     SupervisedOutput::measure(ad_res, ad_out, out, test_id, test_name);
     _module.train();
     return 0;
   }
 
+  double area(const std::vector<double> &bbox)
+  {
```
fantes, line 1595: maybe move below functions out of torchlib.cc?

fantes (approved):
nice work !

src/backends/torch/torchlib.cc
Bycob: Do you have a suggestion for where it could be moved? I think these function can apply to all backends, so maybe in `src/utils`?

fantes: ok for src/utils :)

beniz:
@Bycob can we specify the `loss` automatically instead of having to set it to `model` ?

src/backends/torch/torchdataset.cc
```diff
@@ -525,6 +525,75 @@ namespace dd
...
+
+    while (std::getline(infile, line))
+      {
+        std::istringstream iss(line);
+        std::string val;
+        iss >> val;
+        int cls = std::stoi(val);
+        classes.push_back(target_to_tensor(cls));
+
+        std::vector<double> bbox;
```
beniz, line 577: you may want to reserve the size of the vector since it is known in advance.

src/backends/torch/torchlib.cc
```diff
@@ -725,7 +731,13 @@ namespace dd
                                              + e.what());
               }
 
-            Tensor loss = tloss.loss(y_pred, y, in_vals);
+            Tensor loss;
+            if (_loss == "model")
```
beniz, line 735: Since the template is known, and if this is the only correct loss option for these object detection models, I believe we should set it automatically.

src/backends/torch/torchlib.cc
```diff
@@ -1150,7 +1162,7 @@ namespace dd
 
             if (bbox)
               {
-                // Supporting only Faster RCNN format at the moment.
+                // Supporting only torchvision format at the moment.
```
beniz, line 1165: you mean `torchvision output format` ?

src/backends/torch/torchlib.cc
beniz: you can also create `src/bbox_utils` files, it is a common practice as bboxes require some specific ops.

src/backends/torch/torchmodule.cc
```diff
@@ -294,7 +294,16 @@ namespace dd
             source = torch_utils::unwrap_c10_vector(output);
           }
       }
-    c10::IValue out_val = source.at(_linear_in);
+    c10::IValue out_val;
+    if (_training && _loss_id >= 0)
```
beniz, line 298: Maybe add a comment that describes what this block does and how loss_id works.

src/backends/torch/torchmodule.h
```diff
@@ -195,7 +195,10 @@ namespace dd
     torch::nn::Linear _linear = nullptr;
 
     torch::Device _device;
-    int _linear_in = 0; /**<id of the input of the final linear layer */
+    bool _training = false; /**<true if model is in training mode */
+    int _linear_in = 0;     /**<id of the input of the final linear layer */
```
beniz, line 199: does this only apply to the object detection models from torchvision ?

src/backends/torch/torchlib.cc
Bycob: So `src/utils/bbox.hpp`?

src/backends/torch/torchlib.cc
Bycob: yes

src/backends/torch/torchmodule.h
Bycob: `_training` field is set when we call the method train(). Currently it's used only for models that have different output on training and eval mode, ie detection models.

src/utils/bbox.hpp
```diff
@@ -0,0 +1,66 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2014-2015 Emmanuel Benazera
```
beniz, line 3: I don't remember writing this :) I let you update the copyright and email!

Bycob: Sure :)

beniz (approved):

Issue #1190, opened 2021-02-23
Inconsistent predictons using refinedet model


YaYaB:
#### Configuration

- Version of DeepDetect:
    - [ ] Locally compiled on:
        - [ ] Ubuntu 18.04 LTS
        - [ ] Other:
    - [ ] Docker CPU
    - [X] Docker GPU
    - [ ] Amazon AMI
- Commit (shown by the server when starting):
GIT REF: heads/v0.13.0:1d85f0063f44f574cd1ca3fcfc9339d8c055f220
#### Your question / the problem you're facing:
I've been analysing a bit the predictions made by detection models and more particularly refinedet model and I found at some point that the predictions with a batch_size at least equal to 2 are not consistent.
To illustrate this I took a model available in deepdetect's models: **faces-512**.
The following script will prepare all we need to setup to launch the test:
```
# choose a base path
BASE_PATH=TODO
mkdir "$BATH_PATH/faces_512"
cd "$BATH_PATH/faces_512"

# Download and extract the model
wget https://deepdetect.com/models/init/desktop/images/detection/faces_512.tar.gz && tar -xvf faces_512.tar.gz

# Download corresponding DD's image
docker pull jolibrain/deepdetect_gpu

# Launch container based on this image
docker run --name dd --gpus 1 -u $(id -u ${USER}):$(id -g ${USER}) -v $BATH_PATH/faces_512:/opt/models -p 8083:8080 -it jolibrain/deepdetect_gpu:latest
```
#### Error message (if any) / steps to reproduce the problem:

- Service creation
```
curl -X PUT "http://localhost:8083/services/raw_face_detection" -d '{
"mllib":"caffe",
"description":"image classification service",
"type":"supervised",
"parameters":{
    "input":{
    "connector":"image",
    "width":512,
    "height":512
    },
    "mllib": {
       "nclasses": 3,
       "best":-1,
       "gpuid":0,
       "net": {"test_batch_size":2}
    }
},
"model":{
    "repository":"$BATH_PATH/faces_512"
}
}'
```

- Prediction
If you launch this prediction request several time you'll observe that the results obtained are not the same
```
curl -X POST "http://localhost:8083/predict" -d '{
      "service":"face_512",
      "parameters":{
        "input":{
          "width":512,
          "height":512
        },
      "mllib":{"net":{"test_batch_size":2}},
        "output":{
          "bbox": true
        }
      },
[...]

PR #1210, opened 2021-03-03
fix(caffe/detection): fix rare spurious detection decoding, see bug 1190

fantes:
fix for https://github.com/jolibrain/deepdetect/issues/1190
this bug was certainly also affecting ssd models, but triggers very rarely (last image of batch may have  reported some wrong bboxes in addition to correct ones. in my understanding it happens only when some values read out of table still make sense, ie the num of image in batch)


src/backends/caffe/caffelib.cc
```diff
@@ -3181,9 +3181,9 @@ namespace dd
                 int results_height = results[0]->height();
                 const int det_size = 7;
                 const float *outr = results[0]->cpu_data();
+                int k = 0;
```
BynaryCobweb, line 3184: what does that "k" do? not sure to understand by reading the code, it seems like it prevents to go beyond `result_heights`, which is the total number of prediction in the batch? maybe add a comment or renaming the variable?

fantes: exactly, so I guess no doc is needed because you had it w/o any help :D

beniz (approved):

BynaryCobweb (approved):

PR #1213, opened 2021-03-04
feat(timeseries): add MAE and MSE metrics

fantes:
MAE and MSE metrics for timeseries (both torch and caffe) 
param.output.measure:["mae","mse"]


src/backends/caffe/caffelib.cc
```diff
@@ -2641,21 +2641,51 @@ namespace dd
...
+                            float res = dv_float_data.at(j).at(k);
+                            target.push_back(res);
+                            if (!cic->_dont_scale_labels)
+                              {
+                                double max
+                                    = cic->_max_vals[cic->_label_pos[k]];
+                                double min
+                                    = cic->_min_vals[cic->_label_pos[k]];
+                                if (cic->_scale_between_minus1_and_1)
+                                  res += 0.5;
```
BynaryCobweb, line 2659: Is this correct? if values are scaled between -1 and 1, then this should be `res = (res+1) * 0.5`, I think?

fantes: indeed we are doing the exact opposite of https://github.com/jolibrain/deepdetect/blob/9597bc0aa36b8d5a397d0b98937cb54347035e50/src/csvinputfileconn.h#L338
and i think you are right, this flag should be _scale_between_minus0.5_and_0.5 ...

fantes:
@Mergofyio rebase

fantes:
@Mergifyio rebase

src/backends/caffe/caffelib.cc
fantes: changed the name of this flag => done

BynaryCobweb (approved):

fantes:
@Mergifyio rebase

tests/ut-torchapi.cc
```diff
@@ -1167,6 +1168,12 @@ TEST(torchapi, service_train_csvts_nbeats)
   ASSERT_TRUE(fabs(jd["body"]["measure"]["train_loss"].GetDouble()) > 0);
   ASSERT_TRUE(jd["body"]["measure"].HasMember("L1_mean_error"));
   ASSERT_TRUE(jd["body"]["measure"]["L1_max_error_0"].GetDouble() > 0.0);
+  ASSERT_TRUE(jd["body"]["measure"].HasMember("MAE_0"));
+  // below mae should betwice normlized error because signal values are between
```
BynaryCobweb, line 1172: typos

beniz (approved):

beniz:
@fantes great! Do we want all the TS metrics to be activated by default on TS training job ?

fantes:
> @fantes great! Do we want all the TS metrics to be activated by default on TS training job ?

no, that is definitely too much on multivariate datasets. It is very easy to change, the most difficult part is to think of a simple API. Any suggestions ?

fantes:
see https://github.com/jolibrain/deepdetect/pull/1231 for a try

PR #1227, opened 2021-03-11
SWA

fantes:
enable swa https://arxiv.org/abs/1803.05407
parameters.mlllib.solver.swa = true   honored only in case of RANGER or RANGER_PLUS
sam_swa_metric
sam_swa_legend

### DIFF
--- a/src/backends/torch/optim/ranger.cc
+++ b/src/backends/torch/optim/ranger.cc
@@ -52,7 +52,8 @@ namespace dd
            && (lhs.lookahead() == rhs.lookahead())
            && (lhs.adabelief() == rhs.adabelief())
            && (lhs.gradient_centralization() == rhs.gradient_centralization())
-           && (lhs.lsteps() == rhs.lsteps()) && (lhs.lalpha() == rhs.lalpha());
+           && (lhs.lsteps() == rhs.lsteps()) && (lhs.lalpha() == rhs.lalpha())
+           && (lhs.swa() == rhs.swa());
   }
 
   void RangerOptions::serialize(torch::serialize::OutputArchive &archive) const
@@ -68,6 +69,7 @@ namespace dd
     _TORCH_OPTIM_SERIALIZE_TORCH_ARG(gradient_centralization);
     _TORCH_OPTIM_SERIALIZE_TORCH_ARG(lsteps);
     _TORCH_OPTIM_SERIALIZE_TORCH_ARG(lalpha);
+    _TORCH_OPTIM_SERIALIZE_TORCH_ARG(swa);
   }
 
   void RangerOptions::serialize(torch::serialize::InputArchive &archive)
@@ -83,6 +85,7 @@ namespace dd
     _TORCH_OPTIM_DESERIALIZE_TORCH_ARG(bool, gradient_centralization);
     _TORCH_OPTIM_DESERIALIZE_TORCH_ARG(int, lsteps);
     _TORCH_OPTIM_DESERIALIZE_TORCH_ARG(double, lalpha);
+    _TORCH_OPTIM_DESERIALIZE_TORCH_ARG(bool, swa);
   }
 
   bool operator==(const RangerParamState &lhs, const RangerParamState &rhs)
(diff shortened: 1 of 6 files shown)


beniz:
Can it work with ADAM ?

src/backends/torch/optim/ranger.cc
```diff
@@ -227,11 +237,40 @@ namespace dd
...
 
+  void Ranger::swap_swa_sgd()
+  {
+    for (auto &group : param_groups_)
+      {
+        for (auto &p : group.params())
+          {
+            auto &options = static_cast<RangerOptions &>(group.options());
+            if (!options.swa())
+              continue;
```
Bycob, line 261: Move in outer loop, since the whole group can be skipped?

src/backends/torch/torchlib.cc
```diff
@@ -485,10 +485,12 @@ namespace dd
                 TMLModel>::snapshot(int64_t elapsed_it, TorchSolver &tsolver)
   {
     this->_logger->info("Saving checkpoint after {} iterations", elapsed_it);
+    tsolver.swap_swa_sgd();
     this->_module.save_checkpoint(this->_mlmodel, std::to_string(elapsed_it));
-    // Save optimizer
+    // Save optimizer, swapped version
     tsolver.save(this->_mlmodel._repo + "/solver-" + std::to_string(elapsed_it)
                  + ".pt");
+    tsolver.swap_swa_sgd();
```
Bycob, line 493: This looks a bit like a hack

fantes: agreed, but I could find any other way to do , do you have any idea ? 
(swa simply makes 

> Can it work with ADAM ?

theroetically yes, problem is on implementation side, the average is more naturally implemented inside the solver itself so that it can be saved/snapshoted along with other solver params, so we need to rewrite adam (or simply inherit) in order to add an attribute  and override save/load. 
Instead of inheriting a new solver from adam, I preferred to add this in our ranger (which is basically adam + some options). So if you want to try swa on adam, you can select ranger and deactivate all option (lookaeahd, gradient centralization and so on) in order to get vanilla adam

fantes:
> Can it work with ADAM ?

theroetically yes, problem is on implementation side, the average is more naturally implemented inside the solver itself so that it can be saved/snapshoted along with other solver params, so we need to rewrite adam (or simply inherit) in order to add an attribute and override save/load.
Instead of inheriting a new solver from adam, I preferred to add this in our ranger (which is basically adam + some options). So if you want to try swa on adam, you can select ranger and deactivate all option (lookaeahd, gradient centralization and so on) in order to get vanilla adam

src/backends/torch/torchlib.cc
Bycob: I would suggest moving "swap_swa_sgd" call directly in ranger.
If we need to save the solver both with average and with gradients, create a new method "save_predict_checkpoint"? If the solver is always saved with averaged weights, no need for a new method.

src/backends/torch/torchsolver.cc
```diff
@@ -251,6 +257,7 @@ namespace dd
         try
           {
             torch::load(*_optimizer, sstate, device);
+            swap_swa_sgd();
```
Bycob, line 260: swap_swa_sgd directly in Ranger::load?

src/backends/torch/torchsolver.h
```diff
@@ -105,6 +106,12 @@ namespace dd
       return _base_lr;
     }
 
+    void swap_swa_sgd()
+    {
+      if (_swa)
+        (reinterpret_cast<Ranger *>(_optimizer.get()))->swap_swa_sgd();
+    }
```
Bycob, line 113: With all modifications above we should not require this method anymore.

src/backends/torch/optim/ranger.cc
fantes: ok, done

src/backends/torch/torchsolver.h
fantes: mmmm I do not see how to remove it entirely

src/backends/torch/torchlib.cc
fantes: i cleanud up things a bit, now there is a solver.eval() and solver.train() that mimics module.train() and module.eval() , makeing things a bit more readable

src/backends/torch/torchsolver.cc
fantes: ok

fantes: well, no in fact, see below : there is an explicit solver.train() that hides the swap

fantes:
@Bycob let me know if this version seems better to your eyes

Bycob (approved):

beniz (approved):

PR #1230, opened 2021-03-12
fix(nbeats): make seasonality block work

fantes:
this PR fixes seasonality block thanks to @mathilde-schveitzer 's work
basically, the number of function in basis was not large enough, we now put a max value as a default
we also get back to simpler linspaces definitions (ie we also fix the conversion from timesteps to values to give to functions from this basis)
exemple : 
parameters.mllib.template = "nbeats"
parameters.mllib.template_params= {"stackdef" : ["s","g8","g8","b3","h8"] }

s for 1 seasonality stack
g8 for 1 generic stack of output width 8 * _datadim (number of signals)
b3 for 3 blocks per stack
h8 : internal width of 8 * _datadim everywhere (every block contains starts with a  3 layer MLP of width 8 * _datadim in this case)

### DIFF
--- a/src/backends/torch/native/native_factory.cc
+++ b/src/backends/torch/native/native_factory.cc
@@ -48,8 +48,15 @@ namespace dd
   {
     if (tdef.find("nbeats") != std::string::npos)
       {
-        std::vector<std::string> p;
         double bc_loss_coef = NBEATS_DEFAULT_BACKCAST_LOSS_COEFF;
+        // t2 means 1 trend stack  with two functions in basis
+        // s means 1 seasonalitystack with max number of functions in basis
+        // g3 means 1 generic stack of 3 * datadim width after backbone
+        // b3 means 3 blocksr stack
+        // h10 means a width of 10* datadim for the backbone (common to all
+        // blocks)
+        std::vector<std::string> p{ "t2", "s", "g3", "b3", "h10" };
+
         if (template_params.has("stackdef"))
           {
             p = template_params.get("stackdef")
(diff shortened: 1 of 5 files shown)


src/backends/torch/native/templates/nbeats.h
```diff
@@ -364,8 +364,15 @@ namespace dd
...
               _stack_types.push_back(seasonality);
-              _thetas_dims.push_back(
-                  std::stoi(s.substr(pos + _season_str.size())));
+              int val = -1;
+              try
+                {
+                  val = std::stoi(s.substr(pos + _season_str.size()));
+                }
+              catch (std::invalid_argument &e)
+                {
```
Bycob, line 373: Why error is ignored here? can we still put a number after "s" e.g "s512" ? if so, I think we should be reading this number only if the string is not "s", and keep the error

fantes: ok, should be bettre now

beniz: image


The only usage of the word I could find !

fantes: https://en.wiktionary.org/wiki/bettre

Bycob (approved):

src/backends/torch/native/native_factory.cc
```diff
@@ -48,8 +48,9 @@ namespace dd
   {
     if (tdef.find("nbeats") != std::string::npos)
       {
-        std::vector<std::string> p;
         double bc_loss_coef = NBEATS_DEFAULT_BACKCAST_LOSS_COEFF;
+        std::vector<std::string> p{ "t2", "s", "g3", "b3", "h10" };
```
beniz, line 52: Can you add a comment on what this means, FTR ?

fantes: ok, done

beniz (approved):

PR #1231, opened 2021-03-15
feat(timeseries): do not output per serie metrics as a default, add p…

fantes:
…refix _all for displaying all metrics


Bycob (approved):

Bycob:
Maybe it would be great to use L1 instead of L1_all in one test and check that the metrics for all signals are not there.
This would improve test coverage.

fantes:
> Maybe it would be great to use L1 instead of L1_all in one test and check that the metrics for all signals are not there.
> This would improve test coverage.

ok, done

beniz (approved):

PR #1236, opened 2021-03-16
fix(torch): display msg if resume fails, also fails if not best_model…

fantes:
….txt file

### DIFF
--- a/src/backends/torch/torchsolver.cc
+++ b/src/backends/torch/torchsolver.cc
@@ -277,40 +277,12 @@ namespace dd
     // reload solver if asked for and set it value accordingly
     if (ad_mllib.has("resume") && ad_mllib.get("resume").get<bool>())
       {
-        for (size_t test_id = 0; test_id < best_iteration_numbers.size();
-             ++test_id)
-          {
-            std::string bestfilename
-                = mlmodel._repo
-                  + fileops::insert_suffix("_test_" + std::to_string(test_id),
-                                           mlmodel._best_model_filename);
-            std::ifstream bestfile;
-            try
-              {
-                std::string tmp;
-                std::string bin;
-                std::string test_name;
-                bestfile.open(bestfilename, std::ios::in);
-                // first three fields are thrown away
-                bestfile >> tmp >> bin >> tmp >> tmp >> tmp >> test_name;
-                if (test_name != set_names[test_id])
-                  _logger->warn(
-                      "test names not matching: {} (API) vs {} (file)",
-                      set_names[test_id], test_name);
-                bestfile.close();
-                best_iteration_numbers[test_id] = std::atof(bin.c_str());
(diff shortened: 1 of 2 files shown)


src/backends/torch/torchsolver.cc
```diff
@@ -277,6 +277,25 @@ namespace dd
     // reload solver if asked for and set it value accordingly
     if (ad_mllib.has("resume") && ad_mllib.get("resume").get<bool>())
       {
+        if (mlmodel._sstate.empty())
+          {
+            std::string msg
+                = "resuming a model requires a solverstate file in model "
```
beniz, line 283: `solverstate` was the file extension for caffe, I suggest we change it here so it is relevant for torch.

src/backends/torch/torchsolver.cc
```diff
@@ -277,6 +277,25 @@ namespace dd
...
+            throw MLLibBadParamException(msg);
+          }
+        else
+          try
+            {
+              return load(mlmodel._sstate, main_device);
+            }
+          catch (std::exception &e)
+            {
+              this->_logger->error("Failed to load solver state");
```
beniz, line 295: maybe log the name of the directory in which the file should be found ?

src/backends/torch/torchsolver.cc
```diff
@@ -291,6 +310,14 @@ namespace dd
                 std::string bin;
                 std::string test_name;
                 bestfile.open(bestfilename, std::ios::in);
+                if (!bestfile.is_open())
```
beniz, line 313: Does this mean resuming fails if there's a file but no identified best model ?

src/backends/torch/torchsolver.cc
```diff
@@ -306,30 +333,15 @@ namespace dd
...
-              this->_logger->error("Failed to load solver state");
-              throw;
-            }
       }
     else if (!mlmodel._sstate.empty())
       {
         this->_logger->error("not resuming while a solverstate file remains "
                              "in model repository");
         throw MLLibBadParamException(
             "a solverstate file requires a resume argument for training, "
```
beniz, line 342: Same remark as above, we should swap `solverstate` with another formulation that is better grounded to torch filenames.

src/backends/torch/torchsolver.cc
fantes: ok, done

src/backends/torch/torchsolver.cc
fantes: ok

src/backends/torch/torchsolver.cc
fantes: yes, and it is corrected now: resume can be done even if not test has been done yet

Bycob (approved):

beniz (approved):

PR #1240, opened 2021-03-22
feat(torch): model publishing with the platform

Bycob:


fantes:
maybe add a simple test ?

fantes (approved):
maybe add a test ?

beniz (approved):

PR #1243, opened 2021-03-24
fix(torch): allow partial or mismatching weights loading only if fine…

Bycob:
…tuning

### DIFF
--- a/src/backends/torch/torchmodule.cc
+++ b/src/backends/torch/torchmodule.cc
@@ -124,8 +124,9 @@ namespace dd
         _logger->info("loading " + tmodel._native);
         try
           {
+            // finetuning relaxes strictness when loading weights
             torch_utils::load_weights(*_native, tmodel._native, _device,
-                                      _logger);
+                                      _logger, !_finetuning);
           }
         catch (std::exception &e)
           {
--- a/src/backends/torch/torchutils.cc
+++ b/src/backends/torch/torchutils.cc
@@ -128,7 +128,7 @@ namespace dd
 
     void copy_weights(const torch::jit::script::Module &from,
                       torch::nn::Module &to, const torch::Device &device,
-                      std::shared_ptr<spdlog::logger> logger)
+                      std::shared_ptr<spdlog::logger> logger, bool strict)
     {
       auto from_params = from.named_parameters();
       auto to_params = to.named_parameters();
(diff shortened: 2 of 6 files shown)


src/backends/torch/torchutils.cc
```diff
@@ -174,17 +186,27 @@ namespace dd
...
+                      param_name + " was not found in source module.");
+                }
             }
         }
     }
 
     void load_weights(torch::nn::Module &module, const std::string &filename,
                       const torch::Device &device,
-                      std::shared_ptr<spdlog::logger> logger)
+                      std::shared_ptr<spdlog::logger> logger, bool strict)
```
beniz, line 206: `const bool &strict` ?

beniz:
My main question is where is the `strict` value set to true, when this is useful / required ?

fantes (approved):

Bycob:
> My main question is where is the `strict` value set to true, when this is useful / required ?

"strict" forces all the weights to correspond when you load them. This is the behavior we want most of the time, but if we want to finetune a resnet50 model for example, we have to skip the last layer. So in case of finetuning the "strict" behavior is not enforced.

src/backends/torch/torchutils.cc
Bycob: references are pointers, and pointers are bigger than booleans, so unless there is a reason I don't see here, it is faster to pass booleans by value

beniz:
> This is the behavior we want most of the time

but where is `strict` set to true ?

src/backends/torch/torchmodule.cc
```diff
@@ -122,8 +122,9 @@ namespace dd
         _logger->info("loading " + tmodel._native);
         try
           {
+            // partial weights loading is allowed only if finetuning
             torch_utils::load_weights(*_native, tmodel._native, _device,
-                                      _logger);
+                                      _logger, !_finetuning);
```
Bycob, line 127: here

beniz: OK, let's add a comment, e.g. `// finetuning relaxes strictness when loading weights
Otherwise I can see the case where it's hard to track where `strict` is set.

Bycob: Done

PR #1245, opened 2021-03-29
refactor(dto): ncnn init and generic dto objects

Bycob:


src/backends/ncnn/ncnnlib.cc
```diff
@@ -31,6 +31,8 @@
 #include "net.h"
 #include <iostream>
 
+#include "dto/mllib.hpp"
```
beniz, line 34: This file could be included in `mllibstrategy.h` since it is needed for all backends ?

beniz: Also, it's already included into `ncnnlib.h`.

src/dto/mllib.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include "oatpp/core/Types.hpp"
+#include "oatpp/core/macro/codegen.hpp"
+
+namespace dd
+{
+  namespace DTO
+  {
+#include OATPP_CODEGEN_BEGIN(DTO) ///< Begin DTO codegen section
+
+    class MLLib : public oatpp::DTO
```
beniz, line 36: We already have an `MLLib` class in `mllibstrategy.h`, this could be `DTO_MLLib` ?

src/dto/model.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "oatpp/core/Types.hpp"
+#include "oatpp/core/macro/codegen.hpp"
+
+namespace dd
+{
+  namespace DTO
+  {
+#include OATPP_CODEGEN_BEGIN(DTO) ///< Begin DTO codegen section
+
+    class Model : public oatpp::DTO
```
beniz, line 34: `DTO_DD_Model` instead, or something else ? I understand the namespace should be enough, but I find it confusing. Maybe we just need a naming convention for the API objects.

src/dto/mllib.hpp
beniz: Or let's define a naming convention of the API objects.

src/supervisedoutputconnector.h
```diff
@@ -161,10 +163,11 @@ namespace dd
...
     {
       APIData ad_out = ad.getobj("parameters").getobj("output");
-      if (ad_out.has("best"))
-        _best = ad_out.get("best").get<int>();
+      auto output_params
+          = ad_out.createSharedDTO<PredictOutputParametersDto>();
+      _best = output_params->best;
       if (_best == -1)
-        _best = ad_out.get("nclasses").get<int>();
+        _best = output_params->nclasses;
```
Bycob, line 170: Is there a case where `nclasses` is defined by the user in the `parameters > output` block?

src/supervisedoutputconnector.h
```diff
@@ -242,13 +245,11 @@ namespace dd
...
      */
-    void best_cats(const APIData &ad_out, SupervisedOutput &bcats,
+    void best_cats(SupervisedOutput &bcats, const int &output_param_best,
                    const int &nclasses, const bool &has_bbox,
                    const bool &has_roi, const bool &has_mask) const
     {
-      int best = _best;
-      if (ad_out.has("best"))
-        best = ad_out.get("best").get<int>();
+      int best = output_param_best;
```
Bycob, line 252: Here the behavior changes. Before DTOs the parameter could be set during `init`, and can be modified in predict call if specified. Now that we added a default value, we cannot know if the user specified this parameter or not during predict, so we modify it every time.

Is it fine like this? Or should we stick to the old behavior? I'm not sure how to do with DTOs but it must be possible to know if the user specified a value or not (probably by removing the default value at predict time)

src/supervisedoutputconnector.h
beniz: Not defined by the user, but added from https://github.com/jolibrain/deepdetect/blob/master/src/backends/caffe/caffelib.cc#L3656

src/supervisedoutputconnector.h
beniz: A solution might be to have a `default_best` and `user_best` and to use `default_best` when `user_best` hasn't been set ? or any similar mechanism.
Otherwise there's the risk that the last set value does stick, whereas `predict` calls should not permanently modify the resource that is the service.

src/supervisedoutputconnector.h
Bycob: Ok but we are in the `init()` method of the supervised connector here, and I don't think the mllib ever has the opportunity to modify this parameter. So unless `SupervisedOutputConnector::init()` is called from somewhere else at prediction time, nclasses may always be default / not initialized.

Bycob:
## Updates

- `best` now has exactly the same behavior as before
-  Predict Output DTO (predict.hpp) moved from `http/dto/ ` to `dto/`
- I kept the Output DTO (used when creating a service) and the Predict Output DTO (used for predict). Should they be the same DTO? or two separate objects as it is right now? I prefer the second option since output parameters seem to be very different between service creation and predict call. Maybe we should create a Train Output DTO as well?

beniz (approved):

fantes (approved):

PR #1248, opened 2021-03-31
feat(torch): allow FP16 inference on GPU

fantes:
at predict time, allow to change internal data representation
simply add: 
`parameters.mllib.datatype : "fp16"`
(default is fp32)
also supports fp64 (double) on cpu and gpu

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -79,6 +79,7 @@ namespace dd
     _bbox = tl._bbox;
     _loss = tl._loss;
     _template_params = tl._template_params;
+    _dtype = tl._dtype;
   }
 
   template <class TInputConnectorStrategy, class TOutputConnectorStrategy,
@@ -201,6 +202,27 @@ namespace dd
         _template_params = lib_ad.getobj("template_params");
         if (lib_ad.has("nclasses"))
           _template_params.add("nclasses", lib_ad.get("nclasses").get<int>());
+      }
+    if (lib_ad.has("datatype"))
+      {
+        std::string dt = lib_ad.get("datatype").get<std::string>();
+        if (dt == "fp32")
+          {
+            _dtype = torch::kFloat32;
+            this->_logger->info("will predict in FP32");
+          }
+        else if (dt == "fp16")
+          {
+            _dtype = torch::kFloat16;
+            this->_logger->info("will predict in FP16");
+          }
+        else if (dt == "fp64")
+          {
+            _dtype = torch::kFloat64;
+            this->_logger->info("will predict in FP64");
+          }
+        else
+          throw MLLibBadParamException("unknown datatype " + dt);
       }
 
     // Find GPU id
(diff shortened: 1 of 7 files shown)


src/backends/torch/torchutils.cc
```diff
@@ -57,7 +57,12 @@ namespace dd
       if (!value.isTensor())
         throw MLLibInternalException("Expected Tensor, found "
                                      + value.tagKind());
-      return value.toTensor();
+      torch::Tensor t = value.toTensor();
+      if (t.scalar_type() == torch::kFloat16
+          || t.scalar_type() == torch::kFloat64)
+        return t.to(torch::kFloat32);
+      else
+        return t;
```
Bycob, line 65: I think checking type is redundant here, since it may be done internally in `.to(...)`. So this could be simplified into `value.toTensor().to(torch::kFloat32)`

tests/ut-torchapi.cc
```diff
@@ -1453,8 +1508,6 @@ TEST(torchapi, service_train_csvts_nbeats_db)
   std::string uri = jd["body"]["predictions"][0]["uri"].GetString();
   ASSERT_EQ("../examples/all/sinus/predict/seq_2.csv #0_99", uri);
   ASSERT_TRUE(jd["body"]["predictions"][0]["series"].IsArray());
-  ASSERT_TRUE(jd["body"]["predictions"][0]["series"][0]["out"][0].GetDouble()
-              >= -2.0);
```
Bycob, line 1457: Finally removed ? :)
Can't we just have at least a lower bound, e.g. -10 or -100? so if the value is NaN or -1e300 it still fails

Bycob (approved):

src/backends/torch/torchutils.cc
fantes: indeed this check is mandatory: if type is int[something] we do not want to convert (some models take int tensor as inputs, especially traced models)

beniz:
@fantes great! Can you update the API documentation accordingly ?

src/backends/torch/torchlib.cc
```diff
@@ -1033,7 +1064,19 @@ namespace dd
         throw;
       }
     this->_stats.transform_end();
-
+    if (_dtype == torch::kFloat16)
+      {
+        this->_logger->info("performing inference in FP16");
```
beniz, line 1069: I don't believe we want to log this for every inference. I suggest we either remove the logging, or log it once at service creation / mllib init only.

tests/ut-torchapi.cc
fantes: ok, done

fantes:
> @fantes great! Can you update the API documentation accordingly ?

ok, done

src/backends/torch/torchlib.cc
fantes: ok, done, it wasa debugging  log ;)

beniz: image
 ?

fantes: rooooh !

src/backends/torch/torchlib.cc
```diff
@@ -1033,7 +1073,7 @@ namespace dd
         throw;
       }
     this->_stats.transform_end();
-
+    _module.to(_dtype);
```
beniz, line 1076: Can we do this only if the type has changed ? Or this call has no direct performance impact ?

fantes: check is done internally in torch (according to pytorch doc)

beniz (approved):

src/backends/torch/torchutils.cc
Bycob: oh you're right :+1:

fantes: I did not think about this until some ut fail ;)

PR #1251, opened 2021-04-01
fix(torch): Fix underflow in CSVTS::serialize_bounds

Bycob:


fantes (approved):

src/csvtsinputfileconn.cc
```diff
@@ -515,18 +515,30 @@ namespace dd
...
+    if (_min_vals.size() > 0)
+      {
+        out << "min_vals: ";
+        for (unsigned int i = 0; i < _min_vals.size() - 1; ++i)
+          out << " " << std::setprecision(boundsprecision) << _min_vals[i]
+              << " " << delim;
+        out << " " << std::setprecision(boundsprecision)
+            << _min_vals[_min_vals.size() - 1] << std::endl;
+      }
+    else
```
beniz, line 527: `min_vals` and `max_vals` only apply when `scale` is set to true. Does this code raise an exception if `scale` is set to false ?

Bycob: `serialize_bounds` is never called when `scale` is set to true ([here](https://github.com/jolibrain/deepdetect/blob/665b0ba2a921375bdef24ff53e7a194d88b8b84b/src/csvtsinputfileconn.cc#L164))

beniz: OK good, I guess you mean when scale is set to false.

beniz (approved):

PR #1252, opened 2021-04-01
madgrad

fantes:
implementation of madgrad https://arxiv.org/abs/2101.11075 , based on https://github.com/facebookresearch/madgrad
simply choose parameters.mllib.solver.solver_type = MADGRAD
also support swa and lookahead
parameters.mllib.solver.lookahead = true
parameters.mllib.solver.swa = true

### DIFF
--- a/src/backends/torch/optim/madgrad.h
+++ b/src/backends/torch/optim/madgrad.h
@@ -1,0 +1,132 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2019-2020 Jolibrain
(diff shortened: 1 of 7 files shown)


Bycob (approved):

src/backends/torch/optim/madgrad.h
```diff
@@ -0,0 +1,132 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2019-2020 Jolibrain
```
beniz, line 3: 2021 already :)

beniz (approved):

PR #1267, opened 2021-04-26
feat(torch): add batch size > 1 for detection models

Bycob:


src/backends/torch/torchinputconns.h
```diff
@@ -315,6 +309,16 @@ namespace dd
 
   private:
     bool _bbox = false;
+
+    void update_dataset_parameters()
+    {
+      _dataset._inputc = this;
+      _dataset._image = true;
+      _dataset._bbox = true;
```
beniz, line 317: `bbox` may not always be true ?

Bycob: you're right!

fantes (approved):

beniz (approved):

PR #1269, opened 2021-04-27
refactor(torch): torch init to DTO

Bycob:


lganzzzo (approved):
@Bycob ,

Looks good!

And the part with `GpuIds` custom type is correctly implemented.
There is no serializer method for `GpuIds`, but since you don't need to serialize it back to JSON, it's totally fine.

src/backends/torch/torchlib.cc
```diff
@@ -149,84 +150,76 @@ namespace dd
...
+    int embedding_size = mllib_dto->embedding_size;
+    bool freeze_traced = mllib_dto->freeze_traced;
+    _finetuning = mllib_dto->finetuning;
+    _loss = mllib_dto->loss->std_str();
+
+    auto template_params_dto = mllib_dto->template_params;
+    if (template_params_dto != nullptr)
       {
-        _nclasses = lib_ad.get("ntargets").get<int>();
+        /*
```
beniz, line 189: remove commented code ?

src/dto/mllib.hpp
```diff
@@ -45,6 +46,97 @@ namespace dd
...
+      {
+        info->description
+            = "GPU id, use single int for single GPU, -1 for using all GPUs, "
+              "and array e.g. [1,3] for selecting among multiple GPUs";
+      };
+      DTO_FIELD(GpuIds, gpuid) = { VGpuIds{} };
+
+      DTO_FIELD_INFO(finetuning)
+      {
+        // TODO torch does not require weights since the weights are in the
```
beniz, line 86: replace `TODO` with `XXX` for a point of warning/check ?

beniz (approved):

PR #1276, opened 2021-05-04
feat(ml): data augmentation for object detection with torch backend

beniz:
This PR adds data augmentation on both images and bounding boxes for training robus object detectors with the torch backend.
There's support for mirroring, rotations and random affine transforms. Cutout is not supported at this stage, it is easy to add it. Multiple transforms can be applied sequentially.

Example augmented inputs:

test_aug_36
test_aug_9
test_aug_0

It also adds the additional fix/features below:- 
- fixes a bug that had test db reader apply data augmentation at test time
- allow on-the-fly training image resizing when reading from a db created with a different image size
- fix rotation 90 vs 270 degrees,

### DIFF
--- a/src/backends/torch/torchdataaug.cc
+++ b/src/backends/torch/torchdataaug.cc
@@ -24,50 +24,136 @@
 namespace dd
 {
 
+  void write_image_with_bboxes(const cv::Mat &src,
+                               const std::vector<std::vector<float>> &bboxes,
+                               const std::string fpath, int &ii)
+  {
+    cv::Mat src_bb = src.clone();
+    for (size_t bb = 0; bb < bboxes.size(); ++bb)
+      {
+        cv::Rect r(bboxes[bb][0], bboxes[bb][1], bboxes[bb][2] - bboxes[bb][0],
+                   bboxes[bb][3] - bboxes[bb][1]);
+        cv::rectangle(src_bb, r, cv::Scalar(255, 0, 0), 1, 8, 0);
+      }
+    cv::imwrite(fpath + "/test_aug_" + std::to_string(ii) + ".png", src_bb);
+    ++ii;
+  }
+
   void TorchImgRandAugCV::augment(cv::Mat &src)
   {
     // apply augmentation
-    if (_mirror)
-      applyMirror(src);
-    if (_rotate)
-      applyRotate(src);
-    if (_geometry)
-      applyGeometry(src);
+    applyGeometry(src, _geometry_params);
+    applyCutout(src, _cutout_params);
 
-    // should be last, in this order
-    if (_cutout > 0.0)
-      applyCutout(src);
-    if (_crop_size > 0)
-      applyCrop(src);
+    // these transforms do affect dimensions
+    applyCrop(src, _crop_params);
+    applyMirror(src);
+    applyRotate(src);
   }
 
-  void TorchImgRandAugCV::applyMirror(cv::Mat &src)
+  void
+  TorchImgRandAugCV::augment_with_bbox(cv::Mat &src,
+                                       std::vector<torch::Tensor> &targets)
   {
+    torch::Tensor t = targets[0];
+    int nbbox = t.size(0);
+    std::vector<std::vector<float>> bboxes;
+    for (int bb = 0; bb < nbbox; ++bb)
+      {
+        std::vector<float> bbox;
+        for (int d = 0; d < 4; ++d)
+          {
+            bbox.push_back(t[bb][d].item<float>());
+          }
+        bboxes.push_back(bbox); // add (xmin, ymin, xmax, ymax)
+      }
+
+    bool mirror = applyMirror(src);
+    if (mirror)
+      {
+        applyMirrorBBox(bboxes, static_cast<float>(src.cols));
+      }
+    int rot = applyRotate(src);
+    if (rot > 0)
+      {
+        applyRotateBBox(bboxes, static_cast<float>(src.cols),
+                        static_cast<float>(src.rows), rot);
+      }
+    // XXX: no cutout with bboxes (yet)
+    GeometryParams geoparams = _geometry_params;
+    cv::Mat src_c = src.clone();
+    applyGeometry(src_c, geoparams, true);
+    if (!geoparams._lambda.empty())
+      {
+        // geometry on bboxes
+        std::vector<std::vector<float>> bboxes_c = bboxes;
+        applyGeometryBBox(bboxes_c, geoparams, src_c.cols, src_c.rows);
+        if (!bboxes_c.empty()) // some bboxes remain
+          {
+            src = src_c;
+            bboxes = bboxes_c;
+          }
+      }
+
+    // replacing the initial bboxes with the transformed ones.
+    nbbox = bboxes.size();
+    for (int bb = 0; bb < nbbox; ++bb)
+      {
+        for (int d = 0; d < 4; ++d)
+          {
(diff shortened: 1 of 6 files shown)


src/backends/torch/torchdataaug.h
```diff
@@ -23,32 +23,44 @@
...
-          _geometry_zoom_out(geometry_zoom_out),
-          _geometry_zoom_in(geometry_zoom_in),
-          _geometry_pad_mode(geometry_pad_mode), _uniform_real_1(0.0, 1.0),
-          _bernouilli(0.5), _uniform_int_rotate(0, 3)
+    ImgAugParams(const int &img_width, const int &img_height)
+        : _img_width(img_width), _img_height(img_height)
+    {
+    }
+
+    ~ImgAugParams()
```
Bycob, line 46: If the destructor is empty, it's good practice to omit it

src/backends/torch/torchdataaug.h
```diff
@@ -23,32 +23,44 @@
...
+    }
+
+    int _img_width;
+    int _img_height;
+  };
+
+  class CropParams : public ImgAugParams
+  {
+  public:
+    CropParams() : ImgAugParams()
```
Bycob, line 57: You can write `CropParams() = default;`

src/backends/torch/torchlib.cc
```diff
@@ -606,46 +606,48 @@ namespace dd
...
+                = CropParams(crop_size, inputc.width(), inputc.height());
             this->_logger->info("crop_size : {}", crop_size);
           }
         float cutout = 0.0;
+        CutoutParams cutout_params;
         if (ad_mllib.has("cutout"))
           {
             cutout = ad_mllib.get("cutout").get<double>();
+            cutout_params
+                = CutoutParams(cutout, inputc.width(), inputc.height());
```
Bycob, line 623:
```cpp
            cutout_params
                = CutoutParams(ad_mllib.get("cutout").get<double>(), inputc.width(), inputc.height());
            this->_logger->info("cutout: {}", cutout_params.cutout);
```
And remove `float cutout = 0.0`? Same  for `crop_size`.

src/backends/torch/torchdataaug.cc
```diff
@@ -27,47 +27,127 @@ namespace dd
...
+        // debug
+        /*cv::Mat src_bb = src.clone();
+        cv::Rect r(bboxes[bb][0], bboxes[bb][1], bboxes[bb][2] - bboxes[bb][0],
+                   bboxes[bb][3] - bboxes[bb][1]);
+        cv::rectangle(src_bb, r, cv::Scalar(255, 0, 0), 1, 8, 0);
+        cv::imwrite("/tmp/test_aug_" + std::to_string(ii)
+                        + ".png",
+                    src_bb);
+                    ++ii;*/
+        // debug
```
Bycob, line 99: It could be useful to put this in a function somewhere, in case we need it to debug data augmentation later

src/backends/torch/torchdataaug.cc
```diff
@@ -79,69 +159,133 @@ namespace dd
...
   {
+    std::vector<std::vector<float>> nbboxes;
+    for (size_t i = 0; i < bboxes.size(); ++i)
+      {
+        std::vector<float> bbox = bboxes.at(i);
+        std::vector<float> nbox;
+        if (rot == 1) // 90
+          {
+            nbox.push_back(bbox[1]);              // xmin <- ymin
+            nbox.push_back(img_height - bbox[2]); // ymin <- height-xmax
```
Bycob, line 180: `width-xmax` rather than `height-xmax`?

src/backends/torch/torchdataaug.cc
```diff
@@ -279,11 +428,115 @@ namespace dd
...
+            && bbox[1] <= img_height)
+          {
+            std::vector<float> nbbox;
+            nbbox.push_back(std::max(0.0f, bbox[0])); // xmin
+            nbbox.push_back(std::max(0.0f, bbox[1])); // ymin
+            nbbox.push_back(
+                std::min(static_cast<float>(img_width), bbox[2])); // xmax
+            nbbox.push_back(
+                std::min(static_cast<float>(img_height), bbox[3])); // ymax
+            float surfbb = (bbox[2] - bbox[0]) * (bbox[3] - bbox[1]);
```
Bycob, line 510: There is a function `area` in [utils/bbox.hpp](https://github.com/jolibrain/deepdetect/blob/master/src/utils/bbox.hpp)

src/backends/torch/torchdataaug.cc
```diff
@@ -279,11 +428,115 @@ namespace dd
...
+        bboxes[i][1] += img_height;
+        bboxes[i][3] += img_height;
+      }
+
+    // use cp lambda on bboxes
+    warpBBoxes(bboxes, cp._lambda);
+
+    // filter bboxes
+    filterBBoxes(bboxes, img_width, img_height);
+    // TODO: reject if 0 bbox
```
Bycob, line 540: what happens at the moment? models are fed images with 0 bbox during training?

src/backends/torch/torchdataaug.cc
beniz: I don't think so, because at this stage, the input image has already been rotated.

src/backends/torch/torchdataaug.cc
beniz: My understanding is that the sample should be discarded as object detection models cannot be trained from 0 bbox.

src/backends/torch/torchdataaug.cc
beniz: OK, not necessary at this stage I believe, as it should be rewritten/templated for float. Overkill, no ?

src/backends/torch/torchlib.cc
beniz: done

src/backends/torch/torchdataaug.cc
beniz: done

src/backends/torch/torchdataaug.cc
Bycob: If img_height is the image height after rotation, then the `rot == 3` block is wrong, since it uses `height` as image height before rotation.

beniz: The code is correct, what has changed is https://github.com/jolibrain/deepdetect/pull/1276/files#diff-d9fef5075d6397cff9fb5b52d7a57b88f9c388c96f51907d8bde5a5d0f1a1c10R161 and the comment should be that's 90, and the comment that says 90 should say 170, as flipping with 0 means counterclockwise.

Since OpenCV 3.2, there's a `cv::rotate` function, maybe I could just use it, would make `applyRotate` simpler.

beniz: https://stackoverflow.com/questions/2259678/easiest-way-to-rotate-by-90-degrees-an-image-using-opencv

src/backends/torch/torchdataaug.cc
beniz: Two choices if the transformation yields 0 bboxes:
- reject the image
- reject the transformation

I'll implement the second choice by cancelling the effect of the geometric augmentation on that particular image.

src/backends/torch/torchdataaug.cc
beniz: @Bycob you are right, there was a bug visible when using non-squared images, this is now fixed!

fantes (approved):

PR #1277, opened 2021-05-12
fix(ncnn): fix ncnnapi deserialization error

Bycob:
NCNN ci was not executed

### DIFF
--- a/src/apidata.cc
+++ b/src/apidata.cc
@@ -123,8 +123,7 @@ namespace dd
           {
             APIData ad;
             ad.fromRapidJson(jval[cit->name.GetString()]);
-            std::vector<APIData> vad = { ad };
-            add(cit->name.GetString(), vad);
+            add(cit->name.GetString(), ad);
           }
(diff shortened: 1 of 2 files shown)


fantes (approved):

beniz (approved):

src/apidata.cc
```diff
@@ -123,8 +123,7 @@ namespace dd
           {
             APIData ad;
             ad.fromRapidJson(jval[cit->name.GetString()]);
-            std::vector<APIData> vad = { ad };
-            add(cit->name.GetString(), vad);
+            add(cit->name.GetString(), ad);
```
beniz, line 126: This is part of #1269 already right ? do we need it here as well ?

beniz:
This PR appears to be subsumed by #1280

src/apidata.cc
Bycob: This is needed for the tests to pass. This is the same modification everywhere, so we can just let git handle it and there should not be any problem.

PR #1278, opened 2021-05-12
refactor(tensorrt) : predict to dto

fantes:


beniz:
Current error on gpu trt docker build is : 
```
/opt/deepdetect/src/backends/tensorrt/tensorrtlib.cc:412:46: error: 'class dd::DTO::MLLib' has no member named 'gpuid'
#24 329.1   412 |     _gpuid = predict_dto->parameters->mllib->gpuid->_ids[0];
```

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -406,32 +407,28 @@ namespace dd
         _net_mutex); // no concurrent calls since the net is not
                      // re-instantiated
 
-    if (ad.getobj("parameters").getobj("mllib").has("gpuid"))
-      _gpuid = ad.getobj("parameters").getobj("mllib").get("gpuid").get<int>();
+    auto predict_dto = ad.createSharedDTO<DTO::ServicePredict>();
+
+    _gpuid = predict_dto->parameters->mllib->gpuid->_ids[0];
```
Bycob, line 412: _ids size can be 0 in the case where the user provides an empty array as gpuid, so we should at least prevent the segfault by using `at()` instead of `[]`

src/dto/service_predict.hpp
```diff
@@ -40,6 +40,7 @@ namespace dd
       DTO_FIELD(String, service);
       DTO_FIELD(Object<Parameters>, parameters) = Parameters::createShared();
       DTO_FIELD(List<String>, data);
+      DTO_FIELD(Boolean, chain) = false;
```
Bycob, line 43: "chain" should not be in the DTO since it is added to the APIData internally. We need to find a better way to pass it to the predict call (maybe directly as parameter?)

src/backends/tensorrt/tensorrtlib.cc
fantes: as using tensorrt w/o any gpu does not make sense, I will put gpuid to zero if _ids.size() == 0 ; does it seem ok for you ?

src/dto/service_predict.hpp
fantes: removed, will be changed in the future

fantes:
@Mergifyio rebase

beniz (approved):

Bycob (approved):

PR #1283, opened 2021-05-15
chore(ci): tensorrt on by default

beniz:
Get tensorRT unit tests on by default, and this requires TRT to work with Ubuntu 20.04, and this seems to imply TRT8.

### DIFF
--- a/patches/pytorch/pytorch_18_string_view.patch
+++ b/patches/pytorch/pytorch_18_string_view.patch
@@ -1,0 +1,142 @@
+diff --git a/c10/util/string_view.h b/c10/util/string_view.h
+index 2958887afc..a331022aab 100644
+--- a/c10/util/string_view.h
++++ b/c10/util/string_view.h
+@@ -112,7 +112,8 @@ class basic_string_view final {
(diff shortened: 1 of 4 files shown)


beniz:
This now fails due to this error building torch vision: https://github.com/pytorch/vision/issues/3313

beniz:
This might be a fix, https://github.com/pytorch/vision/issues/3695#issuecomment-829363992

patches/pytorch/pytorch_18_string_view.patch
```diff
@@ -0,0 +1,142 @@
+diff --git a/c10/util/string_view.h b/c10/util/string_view.h
```
Bycob, line 1: Maybe deactivate formatting for this patch?

beniz:
This PR now needs our tensorrt patches to support trt8, will be another PR.

patches/pytorch/pytorch_18_string_view.patch
beniz: Yes, and I will resubmit the patch as another PR since it is mandatory for building torch 1.8 with cuda 11.3

PR #1285, opened 2021-05-17
Torch multigpu V2

Bycob:


src/backends/torch/native/native_wrapper.h
```diff
@@ -29,6 +29,35 @@
 
 #include "native_net.h"
 
+namespace std
```
beniz, line 32: do we want to put this block somewhere in `utils/` ?

src/backends/torch/torchlib.cc
```diff
@@ -648,17 +648,67 @@ namespace dd
...
-                    _regression, _classification, _module, this->_logger);
-    TorchSolver tsolver(_module, tloss, _devices, this->_logger);
-    Tensor class_weights = {};
+
+    typedef struct
+    {
+      std::shared_ptr<TorchModule> module;
+      std::shared_ptr<TorchLoss> loss;
+      std::shared_ptr<TorchSolver> solver;
+    } Rank;
```
beniz, line 688: curious: why is this called Rank ? :)

src/backends/torch/torchutils.cc
```diff
@@ -201,6 +201,32 @@ namespace dd
...
+    void copy_native_weights(const torch::nn::Module &from,
+                             torch::nn::Module &to,
+                             const torch::Device &device)
+    {
+      torch::NoGradGuard guard;
+
+      auto from_params = from.parameters();
+      auto to_params = to.parameters();
+
+      for (size_t i = 0; i < from_params.size(); ++i)
```
beniz, line 215: Does it make any sense to use `omp` on the for loop here ?

src/backends/torch/torchlib.cc
Bycob: With ProcessGroups a "rank" is given to each process so that they know on which GPU they must operate. I kept the same name.

src/backends/torch/torchlib.cc
```diff
@@ -794,190 +853,213 @@ namespace dd
...
+        for (size_t j = 1; j < _devices.size(); ++j)
+          {
+            auto params_j = ranks[j].module->parameters();
+
+            for (size_t pi = 0; pi < params.size(); ++pi)
               {
-                tstart = steady_clock::now();
+                Tensor &grad = params[pi].mutable_grad();
+                Tensor gradj = params_j[pi].grad();
+                grad.add_(gradj.to(_main_device));
```
beniz, line 885: Do we need to normalize the gradients by the number of GPUs ?

Bycob: Yes, I can do that

beniz (approved):

fantes (approved):
nice work !

PR #1294, opened 2021-05-26
chore: allow any ONNX model name with tensorrt

beniz:
Allows any ONNX model name for conversion to tensorRT. 

Note: TRT build has been tested independently of CI since TRT builds are deactivated still.


src/backends/tensorrt/tensorrtmodel.h
```diff
@@ -67,7 +67,7 @@ namespace dd
     std::string _def;
     std::string _weights;
     bool _has_mean_file = false;
-    std::string _source_type = "caffe"; // or "onnx"
+    std::string _source_type; //"caffe" or "onnx"
```
Bycob, line 70: Can we still provide the tensorRT engine directly? Some libraries can export tensor engines directly (e.g. torch2trt?) and I know it was possible to import the engine in dede.

beniz: Yes for Caffe, e.g. `net_tensorRT.proto` and any `.onnx` file should work out of the box now.

fantes (approved):

Bycob (approved):

PR #1297, opened 2021-05-28
feat(ml): inference for GAN generators with TensorRT backend

beniz:


fantes (approved):

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -561,8 +571,16 @@ namespace dd
...
-            _inputIndex = _engine->getBindingIndex("data");
-            _outputIndex0 = _engine->getBindingIndex(out_blob.c_str());
+            _inputIndex = 0;
+            if (out_blob == "last")
+              _outputIndex0 = _engine->getNbBindings() - 1;
+            else
+              _outputIndex0 = _engine->getBindingIndex(out_blob.c_str());
+            _dims = _engine->getBindingDimensions(_outputIndex0);
+            this->_logger->info("detected output dimensions: [{}, {} {} {}]",
+                                _dims.d[0], _dims.d[1],
```
Bycob, line 581: what happens if nbDims < 2? segfault?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -603,6 +621,24 @@ namespace dd
                 throw MLLibBadParamException(
                     "timeseries not yet implemented over tensorRT backend");
               }
+            // GAN / raw output
+            else if (!extract_layer.empty())
+              {
+                _buffers.resize(2);
+                _floatOut.resize(_max_batch_size * _dims.d[1] * _dims.d[2]
+                                 * _dims.d[3]);
```
Bycob, line 629: check that _dims.nbDims >= 4?

beniz: Sure, I'll add this. Rational being that an image has channels, height, width and batch size.

src/backends/tensorrt/tensorrtlib.cc
beniz: Yeah the logger will crash. I can catch that error, rational being 2 dims being the absolute minimal since batch size x uni dimensional tensor.

Bycob (approved):

PR #1299, opened 2021-05-28
fix(chain): fixed_size crops now work at the edges of images

Bycob:


src/chain_actions.cc
```diff
@@ -122,9 +122,31 @@ namespace dd
...
+                    cxmin = 0;
+                  }
+                if (cymin < 0)
+                  {
+                    cymax += -cymin;
+                    cymin = 0;
+                  }
+                if (cxmax > img.cols)
+                  {
+                    cxmin -= cxmax - img.cols;
```
beniz, line 138: Here you may run into cxmin < 0 again, and this would mean the crop cannot be set to the proper size. I believe this requires an additional and final 'check' that would fail the crop ?

beniz: Or may this can only happen if the original image has size equal or lower than the requested crop ?

Bycob: Yes, it can only happen if image width < requested fixed_width

beniz (approved):

fantes (approved):

PR #1302, opened 2021-06-03
feat(torch): add dataloader_threads in API

Bycob:


fantes (approved):

fantes (commented):
maybe some test ?

Bycob:
> maybe some test ?

Since dataloader_threads default was changed from 0 to 1, we already use prefetch on every test. I can add a test with dataloader_threads = 2.

beniz (approved):

PR #1304, opened 2021-06-04
feat(torch): support detection models in chains

Bycob:
torch models where only supported at the end of the chain, this adds support for an image model whose outputs go to another model


fantes (approved):

beniz (approved):

fantes:
@Mergifyio refresh

PR #1307, opened 2021-06-07
chore(build): add needed includes for linking with tensorrt dd

Bycob:


src/backends/torch/torchlib.cc
```diff
@@ -1550,6 +1550,22 @@ namespace dd
         unsupo.finalize(output_params, out,
                         static_cast<MLModel *>(&this->_mlmodel));
       }
+
```
beniz, line 1553: I believe this belongs to another PR !

fantes (approved):

CMakeLists.txt
```diff
@@ -813,6 +813,7 @@ if (USE_TORCH)
       "-DCMAKE_CXX_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/ -isystem ${SPDLOG_INCLUDE_DIR} -isystem ${PROTOBUF_INCLUDE_DIR}"
       "-DCMAKE_CUDA_FLAGS=-isystem ${CMAKE_BINARY_DIR}/pytorch/src/pytorch/ -isystem ${SPDLOG_INCLUDE_DIR} -isystem ${PROTOBUF_INCLUDE_DIR}"
       -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc
+    BUILD_COMMAND ""
```
Bycob, line 816: This was accidentally commited, but it prevents pytorch_vision to rebuild twice, so we can let it there? (the library is built at install step anyway)

fantes: ok for me

beniz (approved):

PR #1311, opened 2021-06-16
fix: remove linking errors on oatpp access_log

Bycob:
There was a problem when using access_log.hpp with another oatpp api outside of dd


fantes (approved):

beniz (approved):

src/http/access_log.cpp
```diff
@@ -0,0 +1,29 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2021 Jolibrain SASU
+ * Author: Mehdi Abaakouk <mehdi.abaakouk@jolibrain.com>
```
beniz, line 4: You may want to update the authorship!

PR #1317, opened 2021-07-13
fix(trt): detect architecture and rebuild model if necessary

Bycob:


src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -419,10 +419,19 @@ namespace dd
 
     if (predict_dto->parameters->mllib->gpuid->_ids.size() == 0)
       throw MLLibBadParamException("empty gpuid vector");
+    if (predict_dto->parameters->mllib->gpuid->_ids.size() > 1)
+      throw MLLibBadParamException("multigpu not supported");
```
beniz, line 423: Maybe slightly more explicit, "Tensorrt: multi-gpu inference is not applicable".

tests/ut-tensorrtapi.cc
```diff
@@ -74,12 +74,13 @@ TEST(tensorrtapi, service_predict)
...
   // ASSERT_TRUE(!fileops::remove_file(squeez_repo, "net_tensorRT.proto"));
-  // ASSERT_TRUE(!fileops::remove_file(squeez_repo, "TRTengine_bs48"));
+  // ASSERT_TRUE(!fileops::remove_file(squeez_repo, "TRTengine_archXX_bs48"));
   jstr = "{\"clear\":\"lib\"}";
   joutstr = japi.jrender(japi.service_delete(sname, jstr));
   ASSERT_EQ(ok_str, joutstr);
   ASSERT_TRUE(!fileops::file_exists(squeez_repo + "net_tensorRT.proto"));
-  ASSERT_TRUE(!fileops::file_exists(squeez_repo + "TRTengine_bs48"));
+  // XXX: can't infer name of engine without knowing CUDA architecture
+  // ASSERT_TRUE(!fileops::file_exists(squeez_repo + "TRTengine_archXX_bs48"));
```
beniz, line 83: We need this test though, so we could trigger it at least for our standard CI machine by checking on the arch value, and triggering the test when it fits.

beniz (approved):

fantes (approved):

PR #1320, opened 2021-07-15
Build no server oatpp

Bycob:


CMakeLists.txt
```diff
@@ -965,84 +970,83 @@ if (USE_TENSORRT)
...
-            -DCMAKE_BUILD_TYPE=Release
-            -DBUILD_SHARED_LIBS=OFF
-            -DOATPP_BUILD_TESTS=OFF
-            -DOATPP_INSTALL=OFF
-            -DOATPP_MODULES_LOCATION=CUSTOM
-            -DOATPP_DIR_SRC=${CMAKE_BINARY_DIR}/oatpp/src/oatpp/src
-            -DOATPP_DIR_LIB=${CMAKE_BINARY_DIR}/oatpp/src/oatpp/src
-        DEPENDS oatpp
-    )
+set(OATPP_VERSION "1.2.5")
```
beniz, line 984: Beware, we had the `USE_HTTP_SERVER_OATPP` flag to be able to switch back to the older http server (when available / buildable...). We don't want to remove it just now (if only because of the oatpp bug/leak with zlib).

CMakeLists.txt
```diff
@@ -798,20 +798,25 @@ if (USE_TORCH)
 
   set(PYTORCH_VISION_COMPLETE ${CMAKE_BINARY_DIR}/CMakeFiles/pytorch_vision-complete)
   set(TORCHVISION_LOCATION "${CMAKE_BINARY_DIR}/pytorch_vision/src/pytorch_vision-install")
+
+  set(PYTORCH_VISION_FLAGS
```
beniz, line 802: I guess this does not belong to this PR.

CMakeLists.txt
Bycob: I did not remove the flag, I just always activate oatpp build, since we need it for DTOs

Bycob:
@Mergifyio rebase

fantes (approved):

beniz (approved):

PR #1331, opened 2021-08-24
test(chain): add tests for chains + simsearch

Bycob:


CMakeLists.txt
```diff
@@ -333,8 +333,9 @@ if (USE_CAFFE OR USE_TORCH)
         PREFIX faiss
         SOURCE_DIR ${CMAKE_BINARY_DIR}/faiss/src/faiss
         GIT_REPOSITORY https://github.com/facebookresearch/faiss
+        UPDATE_DISCONNECTED 1
```
beniz, line 336: Sure of this ? this sounds useful for offline building only no ?

Bycob: Same as trt or torchvision, if we don't put it there, faiss is rebuilt every time

PR #1335, opened 2021-08-27
perf: avoid un needed operations in trt process_batch

Bycob:


beniz:
@mergifyio update

beniz:
@mergifyio update

PR #1340, opened 2021-09-10
feat: chain uses dto end to end

Bycob:
Improved performances (0 copy, no expensive structured data conversion)
Can only be used when linking to deepdetect for now

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,9 @@ OPTION(USE_JSON_API "build internal JSON API" ON)
 OPTION(USE_HTTP_SERVER "build cppnet-lib version of http JSON API " OFF)
 OPTION(USE_HTTP_SERVER_OATPP "build oatpp version of http JSON API" ON)
 
+# user set variables
+set(USE_OPENCV_VERSION CACHE STRING "OpenCV Version used by DeepDetect")
+
 # Get the current working branch
 execute_process(
   COMMAND git rev-parse --abbrev-ref HEAD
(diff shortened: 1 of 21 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,9 @@ OPTION(USE_JSON_API "build internal JSON API" ON)
 OPTION(USE_HTTP_SERVER "build cppnet-lib version of http JSON API " OFF)
 OPTION(USE_HTTP_SERVER_OATPP "build oatpp version of http JSON API" ON)
 
+# user set variables
+set(OPENCV_VERSION CACHE STRING "OpenCV Version used by DeepDetect")
```
beniz, line 63: Could we do `USE_OPENCV_VERSION` instead ?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -417,7 +417,32 @@ namespace dd
         _net_mutex); // no concurrent calls since the net is not
                      // re-instantiated
 
-    auto predict_dto = ad.createSharedDTO<DTO::ServicePredict>();
+    oatpp::Object<DTO::ServicePredict> predict_dto;
+    if (ad.has("dto"))
+      {
+        // cast to ServicePredict...
+        auto any = ad.get("dto").get<oatpp::Any>();
```
beniz, line 424: This looks a temporary bypass until the full pipeline is DTO. Maybe we should add a comment that says so ?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -438,17 +463,16 @@ namespace dd
...
-    if (_first_predict)
-      this->_logger->info("GPU {} architecture = compute_{}", _gpuid, arch);
-
     TInputConnectorStrategy inputc(this->_inputc);
 
     if (!_TRTContextReady)
       {
+        // detect architecture
+        cudaDeviceProp prop;
+        cudaGetDeviceProperties(&prop, _gpuid);
```
beniz, line 472: This may need to remain outside the conditional, otherwise are we still have to change the GPU id on an existing trt service ?
What's the performance impact ?

src/inputconnectorstrategy.h
```diff
@@ -216,12 +217,33 @@ namespace dd
...
+      _meta_uris = pred_dto->_meta_uris;
+      _index_uris = pred_dto->_index_uris;
+
+      if (_uris.empty())
+        {
+          throw InputConnectorBadParamException("missing data");
+        }
+    }
+
     void set_timeout(const APIData &ad)
```
beniz, line 236: Do we want to mark these functions as deprecated, for later removal ?

src/backends/tensorrt/tensorrtlib.cc
Bycob: It takes 0.3ms to query the architecture

The architecture is not supposed to change once we compiled the TRT engine, otherwise we would need to recompile the model anyway, and thus this code would be called again.

I don't think it's possible to change gpuid on one service if the gpu architectures are different, for the same reason: you would have to change the TRT engine, and I don't see any code doing that.

src/inputconnectorstrategy.h
Bycob: This function is still used by the APIData parts of the code atm, that's why it's still there

src/backends/tensorrt/tensorrtlib.cc
beniz: OK, thanks, good point. We should be able to change the `gpuid` of a service when architecture is unchanged. Whether we should recompile on the fly whenever the architecture changes is outside the scope of this PR.

fantes (approved):

beniz (approved):

PR #1344, opened 2021-09-20
feat: DETR export and inference with torch backend

beniz:
This adds export and inference for [detr](https://github.com/facebookresearch/detr) models.

Usage:
```
python3 trace_detr.py --model-in-file /path/to/detr-r50-e632da11.pth  --no_aux_loss --path-to-detr /path/to/detr/
```

Output:
image

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -310,7 +310,8 @@ namespace dd
       {
         _seq_training = true;
       }
-    else if (_template == "fasterrcnn" || _template == "retinanet")
+    else if (_template == "fasterrcnn" || _template == "retinanet"
+             || _template == "detr")
       {
         _bbox = true;
         _classification = false;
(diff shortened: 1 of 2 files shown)


fantes (approved):

src/backends/torch/torchlib.cc
```diff
@@ -409,6 +410,9 @@ namespace dd
         _module._loss_id = 0;
         _module._linear_in = 1;
       }
+    else if (_template == "detr")
```
Bycob, line 413: Remove

beniz: No this is required unfortunately, otherwise execution falls within the last `else` statement and fails under unknown / empty template.

Bycob: ok, my bad

Bycob (approved):

PR #1357, opened 2021-10-20
chore(ci): docker tensorrt with nvcuvid

beniz:
NVCuvid support added to the tensorrt docker builds.


beniz:
This should be fixed now, in fact there were two issues:
- OpenCV detects the nvcuvid lib, not the headers (though it uses the headers), so we don't need the patch, but we copy the lib stubs instead
- libnvcuvid does come with the NVidia driver. Since the driver lies on the host, not the docker, it is not loaded up while building the docker, it gets loaded up when running though, see https://github.com/stereolabs/zed-examples/issues/212

So last commit fixes these issues. Thanks for the initial check @louisj !

fantes (approved):

Bycob (approved):

PR #1360, opened 2021-10-22
feat: add yolox onnx export and trt support

Bycob:
Depends on #1351

### DIFF
--- a/src/backends/tensorrt/tensorrtlib.cc
+++ b/src/backends/tensorrt/tensorrtlib.cc
@@ -22,7 +22,6 @@
 #include "tensorrtlib.h"
 #include "utils/apitools.h"
 #include "tensorrtinputconns.h"
-#include "utils/apitools.h"
 #include "NvInferPlugin.h"
 #include "../parsers/onnx/NvOnnxParser.h"
 #include "protoUtils.h"
@@ -32,6 +31,7 @@
 #ifdef USE_CUDA_CV
 #include <opencv2/core/cuda_stream_accessor.hpp>
 #endif
+#include "utils/bbox.hpp"
 
 namespace dd
 {
@@ -99,12 +99,15 @@ namespace dd
     _engineFileName = tl._engineFileName;
     _readEngine = tl._readEngine;
     _writeEngine = tl._writeEngine;
+    _arch = tl._arch;
+    _gpuid = tl._gpuid;
     _TRTContextReady = tl._TRTContextReady;
     _buffers = tl._buffers;
     _bbox = tl._bbox;
     _ctc = tl._ctc;
     _timeserie = tl._timeserie;
     _regression = tl._regression;
+    _need_nms = tl._need_nms;
     _inputIndex = tl._inputIndex;
     _outputIndex0 = tl._outputIndex0;
     _outputIndex1 = tl._outputIndex1;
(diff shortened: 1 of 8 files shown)


src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -196,6 +199,21 @@ namespace dd
...
+    if (ad.has("template"))
+      {
+        std::string tmplate = ad.get("template").get<std::string>();
+        this->_logger->info("Model template is {}", tmplate);
+
+        if (tmplate == "yolox")
+          {
+            this->_mltype = "detection";
+            _need_nms = true;
+            _top_k = 200; // XXX: pass as parameter?
```
beniz, line 211: This should be part of the model yes, or of an external config struct for yolox models, outside of tensorrtlib ?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -408,6 +426,9 @@ namespace dd
     nvinfer1::IHostMemory *n
         = _builder->buildSerializedNetwork(*network, *_builderc);
 
+    if (n == nullptr)
+      throw MLLibInternalException("Could not build model");
```
beniz, line 430: Can we log slight more information about the model, most useful for debugging purposes.

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -884,28 +906,69 @@ namespace dd
...
+                    ad_bbox.add("xmin", static_cast<double>(detection[3]));
+                    ad_bbox.add("ymin", static_cast<double>(detection[4]));
+                    ad_bbox.add("xmax", static_cast<double>(detection[5]));
+                    ad_bbox.add("ymax", static_cast<double>(detection[6]));
                     bboxes.push_back(ad_bbox);
                   }
+
+                if (_need_nms)
+                  {
+                    std::vector<std::vector<double>> sorted_boxes;
```
beniz, line 935: Maybe we could put the full nms processing as a function in `bbox_utils` instead of having it into `tensorrtlib` ?

src/backends/tensorrt/tensorrtlib.cc
Bycob: With caffe model we can read this information directly in the model. With onnx export I think it is way harder. So we would need a model config file somewhere?

beniz: Let's put it into the AP instead (output object with a default value of 200 or 400?)

Bycob: it's in `parameters.mllib.topk` and must be defined during `init` (because it's dependent on the model, and it can't change between predict calls)

Bycob:
@mergify update

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -196,6 +199,22 @@ namespace dd
             + this->_mlmodel._repo);
       }
 
+    _top_k = ad.has("topk") ? ad.get("topk").get<int>() : 200;
```
beniz, line 202: This is the line I'd recommend we have set the default value for top k somewhere else. Though I don't want to block the PR for this if it can't be fixed now.
Can you add a comment that this needs to go elsewhere ?

Bycob: Done

beniz (approved):

PR #1365, opened 2021-11-10
fix: avoid crashes caused by an exception in the training loop

Bycob:


fantes (approved):

beniz:
@Bycob clang formatting appears to be wrong somehow.

beniz (approved):

Issue #1368, opened 2021-11-15
Simsearch query segfault when using IVF indexes, but not default/flat index


dgtlmoon:
#### Configuration

- Version of DeepDetect:
    - [ ] Locally compiled on:
        - [ ] Ubuntu 18.04 LTS
        - [ ] Other:
    - [x] Docker CPU
    - [x] Docker GPU
    - [ ] Amazon AMI
- Commit heads/v0.19.0:1673a99ecc922e01dd7cc8845098291ef46a8902

#### Your question / the problem you're facing:

Performing a prediction/search using simsearch, after building the index, but ONLY when using IVF style indexes causes DD to segfault and crash, in other words ;

- Building the index works fine (6000 images) `IVF20,SQ8`
- Train/Writing the index to disk works fine
- When the search is performed, it crashes

When I try without the extra options, it works fine, but it's far too slow for predict/query search lookups (0.8-0.9s per 150,000 images)
```
'index': True, "ondisk": True,
> remove "index_type": "IVF20,SQ8",
> remove "train_samples": 100,
> remove "nprobe": 64
```

I'm using my finetuned classifier `se_resnet_50`

### Create a search service after indexing and query

What's interesting here, is that it will crash before it even gets a chance to write the log, there is no log here that it received the prediction call.. maybe this is a clue, here you can see the `PUT` but no `POST`

```
[2021-11-14 21:29:30.315] [api] [info] HTTP/1.1 "PUT /services/test" <n/a> 201 551ms
open existing index db
[2021-11-14 21:30:08.347] [torchlib] [info] Opened lmdb /var/www/xxx/web/files/trainer/simsearch/model//names.bin
bash: line 1:     7 Segmentation fault      (core dumped) ./dede -host 0.0.0.0
```

the setup..

```
curl -X PUT "http://localhost:8080/services/test" -d '{
       "mllib":"caffe",
       "description":"similarity search service",
       "type":"unsupervised",
       "parameters":{
         "input":{
           "connector":"image",
           "height": 224,
           "width": 224
         },
         "mllib":{
           "nclasses":20
         }
       },
       "model":{
          "templates":"../templates/caffe/",
          "repository":"/var/www/xxx/web/files/trainer/simsearch/model/",
          "weight": "model_iter_13500.caffemodel"
       }
     }'

sleep 1
curl -X POST "http://localhost:8080/predict" -d '{
       "service":"test",
       "parameters":{
         "input":{ "height": 224, "width": 224  },
[...]

PR #1369, opened 2021-11-15
fix: simsearch ivf index craft after reload, disabling mmap

beniz:
This is a fix to #1368 that removes MMAP when reloading an existing faiss index. This potentially impacts working with very large indexes that do not fit in RAM, but none of the other potential fixes, including disabling faiss prefetching, did work.

Related issue: https://github.com/facebookresearch/faiss/issues/1540

### DIFF
--- a/src/simsearch.cc
+++ b/src/simsearch.cc
@@ -281,8 +281,7 @@ namespace dd
     if (fileops::file_exists(index_filename))
       {
         if (_ondisk)
-          _findex
-              = faiss::read_index(index_filename.c_str(), faiss::IO_FLAG_MMAP);
+          _findex = faiss::read_index(index_filename.c_str());
         else
           _findex = faiss::read_index(index_filename.c_str());
         _index_size = _findex->ntotal;


fantes (approved):

src/simsearch.cc
```diff
@@ -281,8 +281,7 @@ namespace dd
     if (fileops::file_exists(index_filename))
       {
         if (_ondisk)
-          _findex
-              = faiss::read_index(index_filename.c_str(), faiss::IO_FLAG_MMAP);
+          _findex = faiss::read_index(index_filename.c_str());
```
Bycob, line 284: log->warn("ondisk not supported, fallback to memory") ?

dgtlmoon: @Bycob that could be a big time saver for someone figuring out why in the future

dgtlmoon:
Docs need changing and the default needs to be `false`, with a link to this issue from the docs

image

dgtlmoon:
Trained/indexed 350 images, with `ondisk: False` (MMAP disabled)

```
'output': {'index': True, "ondisk": False,
    "index_type": "IVF20,SQ8",
    "train_samples": 10,
    "nprobe": 8
}
```

And managed to run a search/query without DD crashing

src/simsearch.cc
beniz: There's no logger inside simsearch at the moment.

beniz:
`ondisk` is fine without MMAP.

src/simsearch.cc
Bycob: Maybe we should add one?

Bycob (approved):

PR #1372, opened 2021-11-17
fix: torch train shuffles image dataset by default

beniz:
Default is now to shuffle image datasets when training with torch. 

In this configuration it's not possible to undo it via API. I don't believe it's an issue though?

Note that there are two shuffling phases: one a dataset creation, the other at every epoch, but they are governed by the same boolean in `TorchDataset`. That is fine I believe.


Bycob:
I think disabling shuffle can be useful in some corner cases, e.g during debugging. Can't we just set it to true by default?

beniz:
> I think disabling shuffle can be useful in some corner cases, e.g during debugging. Can't we just set it to true by default?

Yeah of course, done.

fantes (approved):

Bycob (approved):

PR #1377, opened 2021-11-28
feat: data augmentation for training segmentation models with torch backend

beniz:
This PR brings:
- basic data augmentation (rotation, cutout, perspectives) to segmentation with torch
- swapped jpg compression (quality 100) to png with compression 0, as jpg was corrupting the segmentation masks (even with quality set to 100...)

seg_aug_geom1_src
seg_aug_geom1

### DIFF
--- a/src/backends/torch/torchdataaug.h
+++ b/src/backends/torch/torchdataaug.h
@@ -161,7 +161,7 @@ namespace dd
         = 0.25;                     /**< persp factor: 0.25 means that new
                                       image corners  be in 1.25 or 0.75. */
     uint8_t _geometry_pad_mode = 1; /**< filling around images, 1: constant, 2:
-                                       mirrored, 3: repeat nearest. */
+                                       repeat nearest (replicate). */
     float _geometry_bbox_intersect
         = 0.75; /**< warped bboxes must at least have a 75% intersect with the
                    original bbox, otherwise they are filtered out.*/
@@ -192,12 +192,13 @@ namespace dd
 
     void augment(cv::Mat &src);
     void augment_with_bbox(cv::Mat &src, std::vector<torch::Tensor> &targets);
+    void augment_with_segmap(cv::Mat &src, cv::Mat &tgt);
 
   protected:
-    bool applyMirror(cv::Mat &src);
+    bool applyMirror(cv::Mat &src, const bool &sample = true);
     void applyMirrorBBox(std::vector<std::vector<float>> &bboxes,
                          const float &img_width);
-    int applyRotate(cv::Mat &src);
+    int applyRotate(cv::Mat &src, const bool &sample = true, int rot = 0);
     void applyRotateBBox(std::vector<std::vector<float>> &bboxes,
                          const float &img_width, const float &img_height,
                          const int &rot);
@@ -206,7 +207,8 @@ namespace dd
     void applyCutout(cv::Mat &src, CutoutParams &cp,
                      const bool &store_rparams = false);
     void applyGeometry(cv::Mat &src, GeometryParams &cp,
-                       const bool &store_rparams = false);
+                       const bool &store_rparams = false,
+                       const bool &sample = true);
     void applyGeometryBBox(std::vector<std::vector<float>> &bboxes,
                            const GeometryParams &cp, const int &img_width,
                            const int &img_height);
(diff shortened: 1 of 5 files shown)


Bycob:
PNG compression is always lossless, the compression level sets the trade-off between image size and compression time
A compression level of 0 means the images will be poorly (not?) compressed but very fast. This could lead to huge databases, which is something we want to avoid. I suggest to keep the default compression level of OpenCV (I think it's 1)

And I think jpg compression may still be a good feature to have for big datasets, maybe we can add a `lossless` boolean to `image_to_stringstream` and set it to true only for segmentation targets?

beniz:
Yes, absolutely, but masks must not be compressed.

Bycob:
PNG lossless compression does not alter the image in any way, so compressing masks in PNG at any level is fine, unless there is something I missed?

beniz:
@Bycob see last commit. The most time spent is decoding the images every epoch, the encoding is done once when building the db.

src/backends/torch/torchdataset.cc
```diff
@@ -119,11 +119,23 @@ namespace dd
   }
 
   void TorchDataset::image_to_stringstream(const cv::Mat &img,
-                                           std::ostringstream &dstream)
+                                           std::ostringstream &dstream,
+                                           const bool &compression)
```
Bycob, line 123: I'd call it "lossless", in both cases the image is compressed, but with jpeg the image can be altered, while with png the image remains exactly the same

beniz: In practice PNG is an order of magnitude bigger than JPG, very impacting on datasets with millions of images. Additionnally it doesn't matter the image is compressed for training, since the neural net much often would need to accomodate that type of distorsion.
Ground truth, aka masks however must not be compressed.

Bycob: I do agree with what you say, that's why I asked for keeping images compressed as jpg.
However there is a vocabulary misunderstanding here:
masks must not be *altered*, however we can *compress* them.
PNG *is* a compressed format, as opposed to bitmap, but it's a *lossless* one, which means it reduces the size of the image without altering it at all. In another hand, jpg produces much smaller images but alters the image, so we can not use it for masks.

In both cases the image *is* compressed, so the parameter name `compression` is misleading, it would be preferable to rename it `lossless`.

beniz: OK!

src/backends/torch/torchdataset.cc
```diff
@@ -119,11 +119,23 @@ namespace dd
...
-                                           std::ostringstream &dstream)
+                                           std::ostringstream &dstream,
+                                           const bool &lossless)
   {
     std::vector<uint8_t> buffer;
-    std::vector<int> param = { cv::IMWRITE_JPEG_QUALITY, 100 };
-    cv::imencode(".jpg", img, buffer, param);
+    std::vector<int> param;
+    std::string ext;
+    if (compression)
```
Bycob, line 128: don't forget to change it here :)
and lossless == png so the branches need to be exchanged

fantes (approved):

src/backends/torch/torchdataaug.h
```diff
@@ -151,7 +151,11 @@ namespace dd
     float _prob = 0.0;
     bool _geometry_persp_horizontal
         = true; /**< horizontal perspective change. */
+    // bool _geometry_persp_horizontal_right = true; /**< right or left, when
```
Bycob, line 154: why these are commented out?

beniz: I will remove them, unneeded.

src/backends/torch/torchdataaug.h
```diff
@@ -161,11 +161,15 @@ namespace dd
...
-                                       mirrored, 3: repeat nearest. */
+                                       repeat nearest (replicate). */
     float _geometry_bbox_intersect
         = 0.75; /**< warped bboxes must at least have a 75% intersect with the
                    original bbox, otherwise they are filtered out.*/
     cv::Mat _lambda; /**< warp perspective matrix. */
+    // float quads_x0 = -1.0; /**< quad on x axis, stored for target. */
+    // float quads_y0 = -1.0; /**< quad on y axis, stored for target. */
+    // cv::Point2f _inputQuad[4];
+    // cv::Point2f _outputQuad[4];
```
Bycob, line 172: Remove this too?

beniz: done

Bycob (approved):

PR #1381, opened 2021-12-06
fix(torch): correct output name for onnx classification model

Bycob:


tools/torch/trace_torchvision.py
```diff
@@ -336,11 +336,15 @@ def get_detection_input(batch_size=1, img_width=224, img_height=224):
         # remove extension
         filename = filename[:-3] + ".onnx"
         example = get_image_input(args.batch_size, args.img_width, args.img_height) 
+
+        # change for detection
+        out_name = "prob"
```
beniz, line 341: Do we want to pass it as argument to the script, just in case ? and have the default value as `prob` ?

Bycob: Good ideea, I'll do that

fantes (approved):

PR #1382, opened 2021-12-08
fix(tensorrt): read onnx model to find topk

Bycob:


Bycob:
I just realized that `_top_k` could actually be read directly from the tensorrt model, thoughts on this?

src/backends/tensorrt/protoUtils.cc
```diff
@@ -209,4 +212,51 @@ namespace dd
...
+        }
+
+      int node_count = net.graph().output_size();
+
+      for (int i = 0; i < node_count; ++i)
+        {
+          const auto &out = net.graph().output(i);
+          std::string name = out.has_name() ? out.name() : "";
+
+          if (name == "detection_out")
```
beniz, line 243: Do we want this as parameter to the function or it does not make sense because it is so specialized ? 
If the function is highly specialized that for instance it's only for yolox or ssd models, do we want to modify it's name, like `findTopKYolox` ?

src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -546,8 +547,13 @@ namespace dd
...
           }
 
-        if (_bbox && !this->_mlmodel._def.empty())
-          _top_k = findTopK(this->_mlmodel._def);
+        if (_bbox)
+          {
+            if (this->_mlmodel.is_onnx_source())
+              _top_k = onnx_proto::findTopK(this->_mlmodel._model);
+            else if (!this->_mlmodel._def.empty())
+              _top_k = findTopK(this->_mlmodel._def);
```
beniz, line 555: Do we want this function to then be within a `caffe_proto::` scope ?

Bycob: I will sort this out

src/backends/tensorrt/protoUtils.cc
Bycob: In torch when we export the model we can set the name of the output to whatever we want. I chose "detection_out" because it was the output name used in deepdetect. So this name should be generic to all detection models, and not specific to Yolox

beniz: ok, but that's external to DD code itself, so my opinion is that it is still best to have it as a parameter, it's a detail though.

beniz (approved):

fantes (approved):

src/backends/tensorrt/protoUtils.cc
Bycob: I made it a parameter

PR #1384, opened 2021-12-09
feat(torch): manage models with multiple losses

Bycob:
If a model output a dict containing multiple losses they are retrieved and embedded in the training statistics.
Added reg_weight parameter for YOLOX model to control balance between classification and regression losses


src/backends/torch/torchlib.cc
```diff
@@ -722,6 +722,11 @@ namespace dd
                       .to(_main_device);
           }
       }
+    if (ad_mllib.has("reg_weight"))
+      {
+        _reg_weight = ad_mllib.get("reg_weight").get<double>();
+        this->_logger->info("reg_weight={}", reg_weight);
```
beniz, line 728: `_reg_weight`

Bycob: ups

fantes (approved):

beniz (approved):

src/backends/torch/torchlib.cc
```diff
@@ -688,7 +688,7 @@ namespace dd
             if (!ad_distort.empty())
               {
                 distort_params._prob = ad_distort.get("prob").get<double>();
-                this->_logger->info("noise: {}", distort_params._prob);
+                this->_logger->info("distort: {}", distort_params._prob);
```
beniz, line 691: This should have been already updated on master.

src/backends/torch/torchlib.cc
```diff
@@ -722,6 +722,11 @@ namespace dd
                       .to(_main_device);
           }
       }
+    if (ad_mllib.has("reg_weight"))
+      {
+        _reg_weight = ad_mllib.get("reg_weight").get<double>();
```
beniz, line 727: `yolox_reg_weight` ? I believe we want it in the `net` object ?

Bycob: It can go in the `net` object
Maybe this parameter can be useful for other detection models that have a balance between regression & classification? e.g fasterrcnn or retinanet

src/backends/torch/torchlib.cc
Bycob: [It's not](https://github.com/jolibrain/deepdetect/blob/a03eb87fcd60267deac403e33850fd38c6a7760e/src/backends/torch/torchlib.cc#L691)

src/backends/torch/torchlib.cc
beniz: It'd go into `net` anyways no ?

src/backends/torch/torchlib.cc
beniz: ok

beniz:
image

PR #1385, opened 2021-12-10
dice loss for torch backend

fantes:
dice loss + mods (see main api doc) for torch backend 
see https://arxiv.org/abs/1707.03237

FTR:
 dice loss directly maximizes IOU for all classes independently
this PR also adds options for inter-class weighting (if not weighted classes with small areas are not well taken into account)
- dice_weighted add weights per class based on image stats (ie number of pixels per class over the image)
- dice_weighted_batch add add weights per class based on batch stats (number of pixels per class over the batch)
- dice_weighted_all uses running stats over all images seen during training


beniz:
Great, can you give the detail here and in the API doc on what difference between `dice`, `dice_weighted`, `dice_weighted_all` and `dice_weighted_batch` ? tks!

Bycob (approved):

beniz (approved):

PR #1390, opened 2022-01-10
fix: segmentation with torch backend + full cropping support

beniz:
This PR fixes bugs in the segmentation pipeline with torch backend.

Additionnally it:
- adds a torch_utils function to simplify the debugging of C++ torch tensor shapes
- adds full support to random cropping by allowing test set to sample crops automatically. This has the drawback to add noise to the test metric though.


fantes (approved):

src/backends/torch/torchdataset.cc
```diff
@@ -217,7 +217,8 @@ namespace dd
       {
         std::vector<uint8_t> img_target_data(targets.begin(), targets.end());
         bw_target = cv::Mat(img_target_data, true);
-        bw_target = cv::imdecode(bw_target, CV_LOAD_IMAGE_GRAYSCALE);
+        bw_target = cv::imdecode(
+            bw_target, CV_LOAD_IMAGE_UNCHANGED); // CV_LOAD_IMAGE_GRAYSCALE);
```
Bycob, line 221: forgot to change back to GRAYSCALE?

beniz: Yes, this is a good catch, as usual, thanks ! At the moment it makes no difference if the target files are properly built, but loading in grayscale is more robust anyways.

Bycob (approved):

PR #1394, opened 2022-01-17
feat(chain): add action to draw bboxes as trailing action

Bycob:
This PR opens the possibility for chains to have trailing actions.

- [x] add a demo script

Example chain call:
```json
{
   "chain":{
      "calls":[
         {
            "parameters":{
               "input":{
                  "keep_orig":true,
                  "connector":"image"
               },
               "output":{
                  "confidence_threshold":0.5,
                  "bbox":true
               },
               "mllib":{
                  "net":{
                     "test_batch_size":2
                  }
               }
            },
            "service":"detect",
            "data":[
               "/opt/platform/data/image.jpg"
            ]
         },
         {
            "id":"face_detection_bbox",
            "action":{
               "parameters":{
                  "output_images":true,
                  "write_prob":true
               },
               "type":"draw_bbox"
            }
         }
      ]
   }
}
```
image

### DIFF
--- a/src/chain.cc
+++ b/src/chain.cc
@@ -25,6 +25,28 @@
 
 namespace dd
 {
+
+  void embed_model_output(
+      oatpp::UnorderedFields<oatpp::Any> &dest,
+      std::unordered_multimap<std::string, oatpp::Object<DTO::Prediction>>
+          &other_models_out,
+      const std::string &uri)
+  {
+    auto rhit_range = other_models_out.equal_range(uri);
+
+    for (auto rhit = rhit_range.first; rhit != rhit_range.second; ++rhit)
+      {
+        oatpp::String model_name = rhit->second->uri;
+        rhit->second->uri = nullptr;
+        if (dest->find(model_name) != dest->end())
+          throw ChainBadParamException(
+              "This key already exists and cannot be used to reference "
+              "model output: "
+              + model_name->std_str());
+        (*dest)[model_name] = rhit->second;
+      }
+  }
+
   oatpp::Object<DTO::ChainBody> ChainData::nested_chain_output()
   {
     // pre-compile models != first model
@@ -38,6 +60,7 @@ namespace dd
       {
         std::string model_id = (*hit).first;
         std::string model_name = get_model_sname(model_id);
+
         if (model_id == _first_id)
           {
             if ((*hit).second.has("dto"))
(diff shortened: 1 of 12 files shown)


src/chain.cc
```diff
@@ -84,29 +85,82 @@ namespace dd
...
-                rhit->second->uri = nullptr;
-                (*class_preds)[model_name] = rhit->second;
+                uri = cls->class_id->std_str();
+                cls->class_id = nullptr;
+                rhit_range = other_models_out.equal_range(uri);
+
+                for (auto rhit = rhit_range.first; rhit != rhit_range.second;
+                     ++rhit)
+                  {
+                    oatpp::String model_name = rhit->second->uri;
```
beniz, line 155: Make this block code reusable as a function ? it's used before the for loop as well.

src/chain_actions.cc
```diff
@@ -336,6 +337,149 @@ namespace dd
...
+        if (_params->output_images)
+          {
+            auto action_pred = DTO::Prediction::createShared();
+
+            // image as float array, same format as GAN
+            // n_chan * width * height
+            std::vector<double> img_vec;
+            img_vec.reserve(rimg.channels() * rimg.rows * rimg.cols);
+
+            for (int c = 0; c < rimg.channels(); ++c)
```
beniz, line 451: Is there any vectorized version of this in OpenCV ?

src/chain_actions.cc
```diff
@@ -336,6 +337,149 @@ namespace dd
...
+        std::string uri = vad.at(i).get("uri").get<std::string>();
+        uris.push_back(uri);
+
+        int im_cols = imgs.at(i).cols;
+        int im_rows = imgs.at(i).rows;
+        int orig_cols = imgs_size.at(i).second;
+        int orig_rows = imgs_size.at(i).first;
+
+        std::vector<APIData> ad_cls = vad.at(i).getv("classes");
+        cv::Mat rimg = imgs.at(i).clone();
```
beniz, line 380: Do we need/want to support GPUMat as well ?

Bycob: Since we cannot yet encode video on the GPU there is no use for a GPUMat at this stage. So the support for GPUMat can be left for another PR.

beniz: OK understood

src/chain_actions.cc
Bycob: I'm swapping channels to match the output format of torch GAN. I think at this point it's simpler to keep it like that.

beniz: Shouldn't we swap only when `rgb` is set to true ?

Bycob: I removed the transformation, now the raw image is returned

beniz (approved):

PR #1397, opened 2022-01-27
chore: yolox export filenames and documentation

beniz:


fantes (approved):

tools/torch/trace_yolox.py
```diff
@@ -104,7 +106,7 @@ def main():
         model.to(device)
         model.eval()
 
-        filename += ".pt"
+        filename += "-" + str(args.img_width) + 'x' + str(args.img_height) + "_cls" + str(args.num_classes) + ".pt"
```
Bycob, line 109: img_width & img_height are not used when exporting the model to torchscript, I think the model would go along with any image dimensions (unless they are not compatible with the convolutions of the backbone? could happen if the image is too small)

beniz: Right, but it does not work when input size is too small. Since the size controls the image before tracing, it seems fair to keep this information.

Bycob: The model is traced only on the onnx branch

When converting the model to torchscript, the resulting model file will be exactly the same whether you specify 300 or 640 or 2 as img_width.

beniz: OK

Bycob (approved):

PR #1398, opened 2022-01-28
fix: torch MaP with bboxes

beniz:
- Using IoU max overlap

### DIFF
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -2144,6 +2144,7 @@ namespace dd
     };
 
     std::vector<eval_info> eval_infos(_nclasses);
+    float overlap_threshold = 0.5; // TODO: parameter
 
     for (int j = 0; j < pred_bbox_count; ++j)
       {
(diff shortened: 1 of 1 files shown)


fantes (approved):

src/backends/torch/torchlib.cc
```diff
@@ -2216,6 +2230,15 @@ namespace dd
...
         vbad.push_back(lbad);
+
+        // debug
+        /*std::cerr << "tp_d=" << eval_infos[label].tp_d
+                  << " / tp_i=" << eval_infos[label].tp_i
+                  << " / fp_d=" << eval_infos[label].fp_d
+                  << " / fp_i=" << eval_infos[label].fp_i << std::endl;
+        std::cerr << "num_pos=" << eval_infos[label].num_pos
+        << " / label=" << label << std::endl;*/
+        // debug
```
Bycob, line 2241: Remove?

beniz: done

Bycob (approved):

PR #1400, opened 2022-02-07
Fix map with torch

fantes:
this PR fixes normalization of MAP when using detection with torch backend

Details: with caffe backend, no empty detection were passed to supervisedoutput connector, but with torch backend this can happen as results are sorted by class, so if a class has no instance in an image, an empty vector is given to supervisedoutput connector. this PR normalizes MAPs with number of non empty detections

### DIFF
--- a/src/supervisedoutputconnector.h
+++ b/src/supervisedoutputconnector.h
@@ -2358,21 +2358,17 @@ namespace dd
     {
       double mmAP = 0.0;
       std::map<int, int> APs_count;
-
+      int APs_count_all = 0;
       // extract tp, fp, labels
       APIData bad = ad.getobj("0");
       int pos_count = ad.get("pos_count").get<int>();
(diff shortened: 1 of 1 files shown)


src/supervisedoutputconnector.h
```diff
@@ -2396,31 +2389,38 @@ namespace dd
...
-              mAP += local_ap;
             }
-          mAP /= static_cast<double>(vbad.size());
-          // do a mean mAP over images in test set
-          mmAP += mAP;
         }
       for (auto ap : APs)
         {
+          if (APs_count[ap.first] == 0)
+            APs[ap.first] = 0.0;
```
Bycob, line 2418: else?

fantes: thx, fixed

Bycob (approved):

PR #1410, opened 2022-03-14
feat(torch): bump to torch 1.11 and torchvision 0.12

fantes:


beniz:
@Mergifyio update

beniz:
@fantes you'd need to update torch version in the `ut-python.py` tests.

fantes:
> @fantes you'd need to update torch version in the `ut-python.py` tests.

mmmm I do not get it

fantes:
> @fantes you'd need to update torch version in the `ut-python.py` tests.

updated dockerfiles , if that was what it was about

CMakeLists.txt
```diff
@@ -787,8 +787,9 @@ if (USE_TORCH)
...
       CONFIGURE_COMMAND ""
       BUILD_COMMAND ""
-      COMMAND test -f ${PYTORCH_COMPLETE} && echo Skipping || ${CMAKE_COMMAND} -E env PATH=${PROTOBUF_LIB_DIR}:$ENV{PATH} BUILD_CUSTOM_PROTOBUF=0 GLIBCXX_USE_CXX11_ABI=1 BUILD_TEST=0 USE_CUDA=${PYTORCH_USE_CUDA} BUILD_CAFFE2_OPS=0 USE_DDLOG=1 CAFFE2_LINK_LOCAL_PROTOBUF=0 "CMAKE_CXX_FLAGS=-isystem${SPDLOG_INCLUDE_DIR} -isystem${PROTOBUF_INCLUDE_DIR}" CMAKE_PREFIX_PATH=${PROTOBUF_LIB_DIR}/cmake MAX_JOBS=8 python3 ../pytorch/tools/build_libtorch.py
+      COMMAND test -f ${PYTORCH_COMPLETE} && echo Skipping || ${CMAKE_COMMAND} -E env PATH=${PROTOBUF_LIB_DIR}:$ENV{PATH} BUILD_CUSTOM_PROTOBUF=0 GLIBCXX_USE_CXX11_ABI=1 BUILD_TEST=0 USE_CUDA=${PYTORCH_USE_CUDA} CMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc BUILD_CAFFE2=1 BUILD_CAFFE2_OPS=1 BUILD_CAFFE2_MOBILE=0 USE_DDLOG=1 USE_TENSORRT=0 CAFFE2_LINK_LOCAL_PROTOBUF=0 "CMAKE_CXX_FLAGS=-isystem ${SPDLOG_INCLUDE_DIR} -isystem ${PROTOBUF_INCLUDE_DIR}" "CMAKE_CUDA_FLAGS=-isystem ${SPDLOG_INCLUDE_DIR} -isystem ${PROTOBUF_INCLUDE_DIR}" CMAKE_PREFIX_PATH=${PROTOBUF_LIB_DIR}/cmake MAX_JOBS=8 python3 ../pytorch/tools/build_libtorch.py
```
beniz, line 790: Do we need the hardcoded path to nvcc ?

tests/ut_python/ut_tools_torch.py
```diff
@@ -34,32 +34,32 @@ def test_resnet50_export(self):
         onnx_file = os.path.join(_temp_dir, "resnet50.onnx")
         self.assertTrue(os.path.exists(onnx_file), onnx_file)
 
-    def test_fasterrcnn_export(self):
-        # Export model (not pretrained because we don't have permission for the cache)
-        subprocess.run(["python3", "trace_torchvision.py", "-vp", "fasterrcnn_resnet50_fpn", "-o", _temp_dir])
-        model_file = os.path.join(_temp_dir, "fasterrcnn_resnet50_fpn-cls91.pt")
-        self.assertTrue(os.path.exists(model_file), model_file)
+    # def test_fasterrcnn_export(self):
```
beniz, line 37: Commented out because it doesn't work anymore ? It seems the issue with fasterrcnn is within torchvision no ?

fantes: this test uses "system" torchvision , which is not patched. To make it work, we would need to patch deepdetect's torchvision and make tracer script use deepdetect torchvision. Do you think we should do this ?

CMakeLists.txt
fantes: we should not. removed and pushed

tests/ut_python/ut_tools_torch.py
fantes: (my hypothesis is that everything will be fixed with torch 1.12 so a few workarounds here are enough, but depending on the use case maybe it should be useful to have a patched torchvision along with a systematic use of deepdetect's version)

beniz (approved):

Bycob (approved):

src/backends/torch/torchlib.cc
```diff
@@ -981,7 +981,7 @@ namespace dd
                       std::string key = e.key().toStringRef();
                       if (!e.value().isTensor())
                         continue;
-                      auto val_t = e.value().toTensor();
+                      const torch::Tensor &val_t = e.value().toTensor();
```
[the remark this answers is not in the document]
fantes: did not think much about performance issues, auto leads to some other type (maybe w/o the const, or not by reference or both, I do not remember), and this prevents compilation below, where val_t is used

PR #1411, opened 2022-03-17
fix: cropped model input size when publishing torch models + tests

beniz:
Fixes a bug introduced by earlier refactoring for publishing torch models with random cropping as data augmentation. 
Adds the relevant unit test as well.


Bycob (approved):

beniz:
@mergifybot update

beniz:
@Mergifyio update

PR #1412, opened 2022-03-17
fix: seeded random crops at test time

beniz:


beniz:
@Mergifyio update

fantes (approved):

src/backends/torch/torchdataset.cc
```diff
@@ -533,52 +546,73 @@ namespace dd
...
-                        bw_target, bw_target.rows, bw_target.cols, true);
-                    t.push_back(targett_seg);
-                  }
-              }
+                    if (_segmentation)
+                      {
+                        at::Tensor targett_seg = image_to_tensor(
+                            bw_target, bw_target.rows, bw_target.cols, true);
+                        t.push_back(targett_seg);
+                        bw_target = bw_target_tmp;
```
Bycob, line 593: Better to do the other way around? Work on bw_target_tmp, so you don't have to replace bw_target in the end.
You can also rename `bw_target_tmp` -> `bw_target_sample`

src/backends/torch/torchdataset.h
```diff
@@ -89,6 +89,10 @@ namespace dd
     bool _test = false;                 /**< whether a test set */
     TorchImgRandAugCV _img_rand_aug_cv; /**< image data augmentation policy. */
 
+    // test control
+    int _test_crop_samples
```
Bycob, line 93: This is supposed to be only related to test, but in the code this actually controls the number of sample per image. The name could be more generic?

src/backends/torch/torchdataset.cc
```diff
@@ -533,52 +546,73 @@ namespace dd
                                    inputc->_bw, inputc->width(),
                                    inputc->height());
 
-                // data augmentation can apply here, with OpenCV
-                if (!_test)
+                int samples = _test_crop_samples;
```
Bycob, line 549: Actually since we know if we are in test set or not, this could just be
`int samples = _img_rand_aug_cv._crop_params._test_crop_samples`

src/backends/torch/torchdataset.h
beniz: OK, changed to `n_crop_samples`, but the API option remains `"test_crop_samples"` because it only affects test samples.

src/backends/torch/torchdataset.cc
beniz: Good point!

src/backends/torch/torchdataset.cc
beniz: I applied the name change only.

Bycob (approved):

PR #1436, opened 2022-05-23
chore: more torch optimizers

beniz:


fantes (approved):

src/backends/torch/torchsolver.cc
```diff
@@ -199,6 +217,9 @@ namespace dd
       }
     if (_sam)
       this->_logger->info("using Sharpness Aware Minimization (SAM)");
+    else
+      this->_logger->info("using optimizer " + _solver_type);
```
Bycob, line 221: I think sam is not a solver, so remove the "else"?

PR #1441, opened 2022-06-02
fix: exported yolox have the correct number of classes

Bycob:
YOLOX no longer have a useless class at index 0 (the "background" class)
- works with torch & trt models
- back compatibility with older models
- old torch models are convertible into trt with current script
- pretrained models need to be reexported


beniz (approved):

fantes (approved):

beniz:
@mergifyio update

Bycob:
@Mergifyio rebase

PR #1448, opened 2022-07-20
Added bool param to service creation for torch to optionally disable …

cchadowitz-pf:
…applying a softmax to the model output (default enabled)

Note: not sure if `_module._softmax_preprocess` (see torchmodule.h) is the right place to store this value, happy to move it around if desired.

### DIFF
--- a/src/dto/mllib.hpp
+++ b/src/dto/mllib.hpp
@@ -176,6 +176,13 @@ namespace dd
       };
       DTO_FIELD(UnorderedFields<Any>, template_params);
 
+      DTO_FIELD_INFO(multi_label)
+      {
+        info->description
+            = "Model outputs an independent score for each class";
+      }
+      DTO_FIELD(Boolean, multi_label) = false;
+
       // Libtorch predict options
       DTO_FIELD_INFO(forward_method)
       {
--- a/docs/api.md
+++ b/docs/api.md
@@ -1124,6 +1124,7 @@ gpu           | bool   | yes      | false   | Whether to use GPU
 gpuid         | int or array | yes | 0      | GPU id, use single int for single GPU, `-1` for using all GPUs, and array e.g. `[1,3]` for selecting among multiple GPUs
 extract_layer | string | yes      | ""      | Returns tensor values from intermediate layers. In bert models "hidden_state" allows to extract raw hidden_states values to return as output. If set to 'last', simply returns the tensor values from last layer.
 forward_method | string | yes | "" | Executes a custom function from within a traced/JIT model, instead of the standard forward()
+multi_label | bool | yes | false | Model outputs an independent score for each class
 
 
 - XGBoost
--- a/src/backends/torch/torchlib.cc
+++ b/src/backends/torch/torchlib.cc
@@ -82,6 +82,7 @@ namespace dd
     _bbox = tl._bbox;
     _segmentation = tl._segmentation;
     _ctc = tl._ctc;
+    _multi_label = tl._multi_label;
     _loss = tl._loss;
     _template_params = tl._template_params;
     _dtype = tl._dtype;
@@ -293,6 +294,11 @@ namespace dd
     else if (mllib_dto->ntargets != 0)
       {
         _regression = true;
+      }
+
+    if (mllib_dto->multi_label)
+      {
+        _multi_label = true;
       }
 
     if (_template == "bert")
@@ -1500,7 +1506,12 @@ namespace dd
                 // XXX: why is (!_timeserie) needed here? Aren't _timeserie and
                 // _classification mutually exclusive?
                 if (extract_layer.empty() && !_timeserie && _classification)
-                  output = torch::softmax(output, 1).to(cpu);
+                  {
+                    if (_multi_label)
+                      output = torch::sigmoid(output).to(cpu);
+                    else
+                      output = torch::softmax(output, 1).to(cpu);
+                  }
                 else if (extract_layer.empty() && !_timeserie && ctc)
                   output = torch::softmax(output, 2).to(cpu);
                 else
--- a/src/backends/torch/torchlib.h
+++ b/src/backends/torch/torchlib.h
@@ -110,6 +110,7 @@ namespace dd
     bool _bbox = false;           /**< select detection type problem */
     bool _segmentation = false;   /**< select segmentation type problem */
     bool _ctc = false;            /**< select OCR type problem */
+    bool _multi_label = false;    /**< whether model outputs multiple labels */
     std::string _loss = "";       /**< selected loss*/
     double _reg_weight
         = 1; /**< for detection models, weight for bbox regression loss. */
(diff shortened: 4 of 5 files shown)


src/dto/mllib.hpp
```diff
@@ -176,6 +176,15 @@ namespace dd
...
       DTO_FIELD(UnorderedFields<Any>, template_params);
 
+      DTO_FIELD_INFO(softmax_postprocess)
+      {
+        info->description
+            = "Post process the model output by applying a softmax. If a "
+              "model "
+              "already applies a sigmoid, and this should be set to false";
+      }
+      DTO_FIELD(Boolean, softmax_postprocess) = true;
```
Bycob, line 186: We might want a higher level name for this parameter. In your case it's multilabel, but maybe we don't need the softmax in other cases either. @beniz any opinion on this?

Bycob: In caffe we had the `multi_label` option, which effectively disabled the softmax: https://github.com/jolibrain/deepdetect/blob/master/src/backends/caffe/caffeinputconns.h#L193. It was in the input connector, but I think you could put it in mllib, since this is also arguably a property of the model/of the output.

Note that the softmax layer is already not used in some cases (eg regression), in general we prefer to let the user chose the type of tasks they want to perform, and do all the dirty work of plugging layers together and post processing the output behind the scenes

cchadowitz-pf:
Prior to my new test, I had `service_predict_native_bw` failing when I ran `ut_torch` - is that expected if I'm not running with GPU support?
```[ RUN      ] torchapi.service_predict_native_bw
[2022-07-21 10:27:29.563] [imgserv] [error] service creation mllib bad param: GPU is not available, service could not be created
[2022-07-21 10:27:29.563] [api] [error] service not found: "imgserv"
joutstr={"status":{"code":404,"msg":"NotFound","dd_code":1002,"dd_msg":"Service Not Found"}}
/deepdetect/tests/ut-torchapi.cc:232: Failure
Expected equality of these values:
  200
  jd["status"]["code"]
    Which is: 16-byte object <94-01 00-00 00-00 00-00 00-00 00-00 00-00 F6-01>
[  FAILED  ] torchapi.service_predict_native_bw (4 ms)
```

cchadowitz-pf:
I've tried running the tests with the changes in [4d010a6](https://github.com/jolibrain/deepdetect/pull/1448/commits/4d010a6b78f3f54073c7d8c483f3f84ca7e9f754) and can't seem to get the param to actually affect the `predict()` call correctly. It's like `_multi_label` isn't actually saved/stored between the mllib initialization and making a predict call. I've tried using `"multi_label": true` in both the service create `mllib` as well as in the predict `mllib`:
```
TEST(torchapi, service_predict_no_softmax_postprocess)
{
  // create service
  JsonAPI japi;
  std::string sname = "imgserv";
  std::string jstr
      = "{\"mllib\":\"torch\",\"description\":\"resnet-50\",\"type\":"
        "\"supervised\",\"model\":{\"repository\":\""
        + incept_repo
        + "\"},\"parameters\":{\"input\":{\"connector\":\"image\",\"height\":"
          "224,\"width\":224,\"rgb\":true,\"scale\":0.0039},\"mllib\":{"
          "\"nclasses\":1000,\"multi_label\":true}}}";
  std::string joutstr = japi.jrender(japi.service_create(sname, jstr));
  ASSERT_EQ(created_str, joutstr);

  // predict
  std::string jpredictstr
      = "{\"service\":\"imgserv\",\"parameters\":{\"input\":{\"height\":224,"
        "\"width\":224},\"output\":{\"best\":10},\"mllib\":{\"multi_label\":true}},\"data\":"
        "[\""
        + incept_repo + "cat.jpg\"]}";

  joutstr = japi.jrender(japi.service_predict(jpredictstr));
  JDoc jd = JDoc();
  std::cout << "joutstr=" << joutstr << std::endl;
  jd.Parse<rapidjson::kParseNanAndInfFlag>(joutstr.c_str());
  ASSERT_TRUE(!jd.HasParseError());
  ASSERT_EQ(200, jd["status"]["code"]);
  ASSERT_TRUE(jd["body"]["predictions"].IsArray());
  ASSERT_EQ(jd["body"]["predictions"][0]["classes"].Size(), 10);

  ASSERT_GT(jd["body"]["predictions"][0]["classes"][0]["prob"].GetDouble(),
            10.0);
  double sum = 0.0;
  for (size_t i = 0; i < jd["body"]["predictions"][0]["classes"].Size(); i++) {
      sum += jd["body"]["predictions"][0]["classes"][i]["prob"].GetDouble();
  }
  ASSERT_GT(sum, 1.0); // Somewhat redundant but still validates the test case
}
```

cchadowitz-pf:
LGTM! Still getting a failure on `torchapi.service_predict_native_bw` due to running the tests without a GPU, I assume that's expected?

cchadowitz-pf:
One additional thing - I have a second torch model that includes a softmax as part of its own `forward()` method:
```
        InceptionResnetV2_Logits_Predictions = F.softmax(InceptionResnetV2_Logits_Logits_MatMul)
        return InceptionResnetV2_Logits_Predictions
```
Obviously if I set the upcoming API param `"multi_label": true` then everything should be fine, but I don't know if this kind of scenario would prompt a different name for the param, as this model isn't a multi-label model, it just already performs the softmax internally and doesn't need post-processing.

Bycob:
> 

Our current policy is to export all the models for a given task with the same input / output format, so that there is one unique path in dede. As torchvision models output logits, this was set as the default. Do you have any way to remove the softmax from your model?

Bycob:
> LGTM! Still getting a failure on `torchapi.service_predict_native_bw` due to running the tests without a GPU, I assume that's expected?

Seems like this test should either not require GPU or not be built on CPU-only dede.

Bycob:
As for the CI: it fails with issues unrelated to this PR's changes. We are fixing the problems, after that it should pass.

cchadowitz-pf:
> Our current policy is to export all the models for a given task with the same input / output format, so that there is one unique path in dede. As torchvision models output logits, this was set as the default. Do you have any way to remove the softmax from your model?
> 
> As such, if we want to be consistent, the `multi_label` model should be post processed with a sigmoid, but that would not go with your models either.

That makes sense.....let me see what I can do about my models on my side, if it's as simple as just removing that lost line in my `forward()` method then we can clean this all up and add the sigmoid call to the `multi_label` code path and the current softmax call to the non-`multi_label` code path and call it a day :)

Bycob:
@cchadowitz-pf multi-label is now post processed by sigmoid, the rest is unchanged. Let me know if it's fine for you

cchadowitz-pf:
Cool thanks! I may not be able to get to it today but I'll test it out and let you know.

Bycob:
Hi @cchadowitz-pf, it would be better if you put the dlib update in another PR, so that we can merge this one.

cchadowitz-pf:
> Hi @cchadowitz-pf, it would be better if you put the dlib update in another PR, so that we can merge this one.

Sounds good - I've reverted the dlib changes here and moved them to a separate PR (https://github.com/jolibrain/deepdetect/pull/1452)  that should be merged **after** this one.

Bycob (approved):

beniz (approved):

PR #1454, opened 2022-08-22
fix: reload best metric correctly after resume

Bycob:


src/utils/utils.hpp
```diff
@@ -68,6 +68,39 @@ namespace dd
       return elems;
     }
 
+    inline std::string trim_spaces(const std::string &item)
```
beniz, line 71: FTR, there seems to be more compact ways of trimming spaces https://www.techiedelight.com/trim-string-cpp-remove-leading-trailing-spaces/ ?

fantes (approved):

beniz (approved):

PR #1459, opened 2022-09-07
test: add a test script to check that yolox export to tensorrt works …

Bycob:
…correctly

### DIFF
--- a/src/backends/tensorrt/tensorrtlib.cc
+++ b/src/backends/tensorrt/tensorrtlib.cc
@@ -789,6 +789,14 @@ namespace dd
         if (num_processed == 0)
           break;
 
+        // some models don't support dynamic batch size (ex: onnx)
+        // this can lead to undetected bad predictions
+        if (num_processed > _dims.d[0])
+          throw MLLibBadParamException(
+              "Trying to process " + std::to_string(num_processed)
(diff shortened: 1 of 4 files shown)


src/backends/tensorrt/tensorrtlib.cc
```diff
@@ -789,6 +789,14 @@ namespace dd
         if (num_processed == 0)
           break;
 
+        // some models don't support dynamic batch size (ex: onnx)
+        // this can lead to undetected bad predictions
+        if (num_processed > _dims.d[0])
```
beniz, line 794: @Bycob don´t we want to limit this to yolox models ? Usually, if batch size > compiled batch size, tensorrt throws a warning saying that performances aren´t optimal, which is enough.

fantes (approved):

src/backends/tensorrt/tensorrtlib.cc
Bycob: Here we check the batch sent by the input connector, so for caffe model this would never trigger, since the model is compiled with the correct batch size. This check is for when max batch size (specified by API) is bigger than the onnx model batch size. Ideally we should check that at service creation, but we only read the onnx model at the first inference

beniz (approved):

PR #1477, opened 2022-12-05
chore: raw image support in cv2 format for dd_client

beniz:
Added `from_img` parameter to `post_predict` to pass raw images, goes along with `use_base64=True`

### DIFF
--- a/clients/python/dd_client/__init__.py
+++ b/clients/python/dd_client/__init__.py
@@ -17,9 +17,9 @@ import base64
 import os
 import re
 import warnings
-
 import requests
-
+import cv2
+import numpy as np
 
 DD_TIMEOUT = 2000  # seconds, for long blocking training calls, as needed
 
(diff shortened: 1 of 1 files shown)


clients/python/dd_client/__init__.py
```diff
@@ -255,6 +258,7 @@ def post_predict(
         parameters_output,
         use_base64=False,
         index_uris=[],
+        from_img=False
```
Bycob, line 261: you could remove "from_img" parameter and just detect whether the image contained in `data` is a `np.array` or a string

beniz: Could be. One drawback is that it requires importing numpy... Already I don´t like having to import cv2 into the client.

beniz: FTR, it's testing for `np.ndarray`.

beniz: cv2 imports numpy anyways...

Bycob: cv2 image is a numpy array

PR #1483, opened 2023-01-04
feat(trt): add int8 inference

Bycob:
int8 inference is the same as standard inference (fp32, fp16) but requires a calibration step first. Calibration should use at least ~100 images that are representative of the space in which the model operates.
```bash
# calibration call example (after creating the service but before any predict call)
curl -X POST localhost:8500/predict -d '{
  "parameters": {
    "input": {
      "rgb": true,
      "scale": 0.00392,
      "mean": [
        0.5,
        0.5,
        0.5
      ],
      "std": [
        0.5,
        0.5,
        0.5
      ]
    },
    "mllib": {
      "extract_layer": "last",
      "calibration": true
    },
    "output": {}
  },
  "service": "int8_gan",
  "data": [
    "/data/calibration_image1.jpg", "/data/calibration_image2.jpg", ...
  ]
}'

```

- Currently, calibration call also returns predictions for every calibration images.
- Calibration must be the first predict call, otherwise the model will fail to build
- Once the calibration has been done, a calibration_table file is generated in the model repository, it can be used to recreate the service with int8 without the need to re-calibrate. This calibration table can be exported to other machines because it does not depends on the gpu.


fantes (approved):

royale (approved):

fantes:
@Mergifyio refresh

fantes:
@Mergifyio refresh

PR #1488, opened 2023-01-10
feat(torch): add radam optimizer

fantes:


Bycob (approved):

src/backends/torch/optim/radam.cc
```diff
@@ -0,0 +1,211 @@
+/**
+ * DeepDetect
+ * Copyright (c) 2019-2020 Jolibrain
```
beniz, line 3: 2023

beniz (approved):

fantes:
@Mergifyio refresh

PR #1494, opened 2023-01-16
feat(torch): allow data augmentation w/o db

fantes:
plus cleanup around image_to_tensor

### DIFF
--- a/src/backends/torch/torchdataset.cc
+++ b/src/backends/torch/torchdataset.cc
@@ -229,20 +229,21 @@ namespace dd
 
     if (bgr.cols != width || bgr.rows != height)
       {
-        float w_ratio = static_cast<float>(width) / bgr.cols;
-        float h_ratio = static_cast<float>(height) / bgr.rows;
         cv::resize(bgr, bgr, cv::Size(width, height), 0, 0, cv::INTER_CUBIC);
 
         if (_bbox)
-          for (int bb = 0; bb < (int)targett[0].size(0); ++bb)
-            {
-              targett[0][bb][0] *= w_ratio;
-              targett[0][bb][1] *= h_ratio;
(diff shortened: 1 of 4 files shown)


royale (approved):

src/backends/torch/torchdataset.cc
```diff
@@ -366,13 +371,86 @@ namespace dd
     return d.at(i).sizes().vec();
   }
 
+  void TorchDataset::dataaug_then_push(const cv::Mat &bgr,
```
Bycob, line 374: dataaug_then_push_back (at first I thought this method was pushing to db)

src/backends/torch/torchdataset.cc
```diff
@@ -366,13 +371,86 @@ namespace dd
...
+              }
+            else if (_segmentation)
+              _img_rand_aug_cv.augment_test_with_segmap(bgr_sample,
+                                                        bw_target_sample);
+            else
+              _img_rand_aug_cv.augment_test(bgr_sample);
+          }
+
+        torch::Tensor imgt
+            = image_to_tensor(bgr_sample, bgr_sample.rows, bgr_sample.cols);
```
Bycob, line 420: It looks like "rows" and "cols" parameters are useless, they can be deduced from the cv::Mat. So we can drop them from image_to_tensor?

fantes: I dont not think it is always the case that we use first param dims when calling image_to_tensor, see for instance torchdataset.cc:260 and torchdataset.cc:279 (ie add_image_batch flavors that are called from inputconn)

fantes: done

Bycob (approved):

PR #1507, opened 2023-02-24
Add the list of labels for a model service to the output of /services/<name>

cchadowitz-pf:
This adds the full list of labels from the corresp.txt file (if present) for the service model when the `/services/<name>` endpoint is hit. E.g.

```
curl -X GET http://localhost:8080/services/placeshybrid | jq
```
produces the following with this change:
```
{
  "status": {
    "code": 200,
    "msg": "OK"
  },
  "body": {
    "mltype": "classification",
    "jobs": [],
    "stats": {
      "data_mem_test": 1152129720,
      "data_mem_train": 0,
      "params": 139839168,
      "flops": 15471759360
    },
    "service_stats": {
      "avg_transform_duration": 0.03255472,
      "avg_predict_duration": 0.837866442,
      "avg_transform_duration_s": 0.03255472,
      "avg_predict_duration_s": 0.837866442,
      "avg_transform_duration_ms": 32.55472,
      "avg_predict_duration_ms": 837.866442,
      "avg_batch_size": 1,
      "total_transform_duration_ms": 32.55472,
      "predict_failure": 0,
      "total_predict_duration_ms": 837.866442,
      "predict_count": 1,
      "predict_success": 1,
      "inference_count": 1
    },
    "labels": [
      "zen garden",
      "youth hostel",
      "yard",
      "windmill",
      "wind farm",
      "wheat field",
      "wet bar",
      "wave",
      "watering hole",
      "waterfall",
      "water tower",
      "water park",
      "waiting room",
      "volleyball court outdoor",
      "volcano",
      "vineyard",
      "village",
      "viaduct",
      "veterinarians office",
      "vegetable garden",
      "valley",
      "utility room",
      "underwater ocean deep",
      "tundra",
      "trench",
      "tree house",
      "tree farm",
      "train station platform",
      "train interior",
      "toyshop",
      "tower",
      "topiary garden",
      "ticket booth",
      "throne room",
      "temple asia",
      "television studio",
      "television room",
      "synagogue outdoor",
      "swimming pool outdoor",
      "swimming pool indoor",
      "swimming hole",
      "swamp",
      "sushi bar",
      "supermarket",
      "subway station platform",
      "street",
      "storage room",
      "staircase",
      "stage outdoor",
      "stage indoor",
      "stadium soccer",
      "stadium football",
      "stadium baseball",
      "stable",
      "soccer field",
      "snowfield",
      "slum",
      "skyscraper",
      "sky",
      "ski slope",
      "ski resort",
      "shower",
[...]

PR #1508, opened 2023-03-01
feat(api): add labels in service info

Bycob:
labels are not returned per default, call service info with query parameter labels=1

Following #1507 

Example:
```bash
curl -X GET http://localhost:8080/services/placeshybrid?labels=1 | jq
```


src/mlservice.h
```diff
@@ -270,6 +271,24 @@ namespace dd
...
+          if (!this->_mlmodel._hcorresp.empty())
+            {
+              labels_vec->reserve(this->_mlmodel._hcorresp.size());
+
+              for (const auto &kv : this->_mlmodel._hcorresp)
+                {
+                  labels_vec->push_back(kv.second);
+                }
+            }
+          // TODO exception if no labels
```
cchadowitz-pf, line 288: Is this exception from oatpp?

Bycob: That's just a note I forgot to remove :) Was thinking of triggering an exception if the model is not supposed to have labels, but hcorresp can also be empty if the model has no corresp file, so triggering an exception would be confusing.

Bycob: I'm removing the TODO, thanks for catching that up

royale (approved):

fantes (approved):

PR #1553, opened 2024-02-08
Tracing CLIP-based vision models for DeepDetect

cchadowitz:
Added basic script to trace HuggingFace Transformers models based on https://huggingface.co/openai/clip-vit-large-patch14-336

Note: this does require patching `transformers/models/modeling_clip.py` as noted in the comments. Without the patch, the values are traced to either `CPUFloatType` or `CUDAFloatType` and so the model cannot be used on either type of device interchangeably.

### DIFF
--- a/tools/torch/trace_clip.py
+++ b/tools/torch/trace_clip.py
@@ -16,30 +16,7 @@ from transformers import CLIPVisionModelWithProjection, TensorType
 import torchvision.transforms as T
 from torchvision.transforms import InterpolationMode
 from typing import Tuple
-import argparse
-import logging
-
-class CLIPWrapper(nn.Module):
-  def __init__(self, visionmodel):
-    super(CLIPWrapper, self).__init__()
-    self.visionmodel = visionmodel
-    self.visionmodel.eval()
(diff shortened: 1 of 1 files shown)


tools/torch/trace_clip.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+# use cache dir if you want to specify a custom huggingface cache dir
+visionmodel = CLIPVisionModelWithProjection.from_pretrained(args.model, torchscript=True, cache_dir=args.cache_dir)
+
+model = CLIPWrapper(visionmodel)
+model.eval()
+
+# Uncomment if you prefer to script the wrapper
+# traced_model = torch.jit.script(model, torch.rand([1, 3, 1366, 1024]))
```
Bycob, line 85: This should be an option? only if torch.jit.script works correctly of course

Bycob: If it's just for debugging, you can tell it in the comment

cchadowitz: It does work, I can make it an option 👍 I wasn't sure if it was worth leaving it in since it seems to behave the same either way.

Bycob:
Closed because it was merged in #1554

